PR #68, opened 2026-04-22
feat(mq): make nats stream name configurable

### LINKED ISSUE #36: Configurable Stream Names
## Problem

We currently hardcode stream names (like "WAVEHOUSE"). This makes it impossible to run a "Staging" and "Production" environment on the same NATS cluster, or for a single user to run multiple independent WaveHouse instances without them clashing and stealing each other's data.

## Proposed Solution

Move stream and subject prefixes into the configuration file. Allow users to define a "Namespace" or "Prefix" (e.g., wh_prod_) that is prepended to all NATS resources.

## Alternatives Considered

None

## Additional Context

None

taitelee:
## Summary

This PR removes the hardcoded "WAVEHOUSE" NATS JetStream name and introduces a configurable StreamName variable to support multi-tenant or multi-environment deployments

- Configurable Namespacing: Added StreamName to the MQ configuration struct, supporting YAML and environment variable overrides (WH_MQ_STREAM_NAME).
- Decoupled Driver Logic: Updated mq.NewEmbedded and mq.NewRemote constructors to accept a dynamic stream name, removing dependency on package-level constants.
- Bento Alignment: Updated ingest.StartIngestWorker to bind its internal consumer to the configured stream, ensuring data ingestion stays within the defined namespace.
- End-to-End Wiring: Refactored all main.go entry points and the integration test suite to propagate the configured stream name throughout the application lifecycle.

## Test plan

- [ ] ...

## Related Issues

Closes #36 

### Configuration Example
Via config.yaml:
`
mq:
  stream_name: "WH_STAGING"
  url: "nats://localhost:4222"
`


internal/api/router.go
```diff
@@ -25,6 +25,7 @@ type Dependencies struct {
 	Pipes           *PipesHandler
 	StructuredQuery *StructuredQueryHandler
 	AuthMW          func(http.Handler) http.Handler
+	AuthEnabled     bool
```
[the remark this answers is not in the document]
taitelee: This was removed for this PR to remove scope creep. The AuthEnabled bool is addressed in another PR.

internal/config/config.go
```diff
@@ -45,6 +45,7 @@ type ClickHouse struct {
 }
 
 type MQ struct {
+	StreamName       string `yaml:"stream_name" env:"WH_MQ_STREAM_NAME" env-default:"WAVEHOUSE"`
```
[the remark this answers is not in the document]
taitelee: This was addressed. I agree we need to add this field in the Validate() method to prevent empty stream names. I have also refactored the DLQ logic to derive its name from this configuration (e.g., WH_PROD_DLQ), ensuring that environment isolation remains consistent across all NATS resources.

internal/api/dlq_test.go
```diff
@@ -60,8 +62,8 @@ func TestDLQStats_ReturnsCorrectCounts(t *testing.T) {
 		require.NoError(t, err)
 	}
 
-	handler := NewDLQHandler(js, slog.Default())
-	req := httptest.NewRequestWithContext(context.Background(), http.MethodGet, "/v1/dlq/stats", nil)
+	handler := NewDLQHandler(js, streamName, slog.Default())
+	req := httptest.NewRequest(http.MethodGet, "/v1/dlq/stats", nil)
```
EricAndrechek: I think we do want `NewRequestWithContext` here, no?

internal/api/dlq_test.go
```diff
@@ -80,20 +82,21 @@ func TestDLQStats_ReturnsCorrectCounts(t *testing.T) {
...
-	require.NoError(t, EnsureDLQStream(ctx, js, 1024*1024))
+	require.NoError(t, EnsureDLQStream(ctx, js, streamName, 1024*1024))
 
 	_, err = js.Publish(ctx, "dlq.orders", []byte(`{"table_name":"orders"}`))
 	require.NoError(t, err)
 
-	handler := NewDLQHandler(js, slog.Default())
-	req := httptest.NewRequestWithContext(context.Background(), http.MethodGet, "/v1/dlq/stats", nil)
+	handler := NewDLQHandler(js, streamName, slog.Default())
+	req := httptest.NewRequest(http.MethodGet, "/v1/dlq/stats", nil)
```
EricAndrechek: Same thing, we do want `NewRequestWithContext` right...? @copilot @claude[agent]

internal/config/config.go
```diff
@@ -124,6 +125,10 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("mq.gap_window_minutes must be non-negative")
 	}
 
+	if c.MQ.StreamName == "" {
+		return fmt.Errorf("mq.stream_name cannot be empty")
```
EricAndrechek: Good to add this, but should make sure that in addition to the `config.yaml` making this field included by default, we:
- also build in a default here if not specified in the config so it doesn't crash and safely starts with `WAVEHOUSE` or some default, OR build a new test case to confirm it fails if not specified
- ensure the test cases are not failing on this constraint (the test in CI fails right now due to this)

EricAndrechek (changes requested):

EricAndrechek (approved):

PR #83, opened 2026-04-28
feat: implement real-time cluster synchronization for Pipes via NATS KV watcher

### LINKED ISSUE #33: KV Watcher Real Time Updates
## Problem

Currently, the Pipes Store (which holds the named SQL queries) only loads data from the NATS Key-Value (KV) store when a WaveHouse instance first boots up. In a clustered environment with multiple API or Worker nodes, this creates a synchronization lag: if an admin updates a Pipe on Node A, Node B will continue running the old, outdated SQL until it is manually restarted. This prevents the "Live Coding" experience where query changes should take effect across the entire cluster instantly.

## Proposed Solution

Implement a KV Watcher pattern within the internal/pipes/store.go. Instead of a one-time "Get" during initialization, the Store should establish a long-lived subscription to the NATS KV bucket.

When a change occurs (Put or Delete), NATS will push a notification to every WaveHouse node. The Store will then reactively update its local in-memory cache. This ensures that every node in the cluster is perfectly synchronized within milliseconds of a Pipe being updated, without requiring any manual restarts or coordination.

## Alternatives Considered

None

## Additional Context

None

taitelee:
## Summary

This PR implements a reactive synchronization pattern for the Pipes store to ensure consistency across WaveHouse clusters. Previously, the Pipes store only loaded named queries from NATS KV during the initial boot sequence, leading to synchronization lag where configuration changes on one node were not reflected on others without a restart.

- KV Watcher Integration: Added a Watch method to the Pipes store using the WatchAll NATS JetStream pattern.
- Reactive Cache Updates: The store now listens for Put and Delete operations on the WAVEHOUSE_PIPES bucket, dynamically updating the local in-memory cache in milliseconds.
- Lifecycle Management: Integrated the watcher into the main application lifecycle via background goroutines in the API and Standalone binaries, ensuring graceful shutdown support through context cancellation.

## Test plan

- [ ] ...

## Related Issues

Closes #33


EricAndrechek:
Triage note as part of the pre-alpha priority pass.

This PR predates the standalone-only refactor in #89. The "real-time cluster synchronization" framing assumes clustered mode, which was removed — the embedded NATS instance is per-binary, so there is no "cluster" to synchronize across.

The underlying UX problem is still real though: today, pipes load from the `WAVEHOUSE_PIPES` KV bucket only at boot, so an admin who updates a pipe via the API (or by editing the bootstrap `.sql` directory) has to restart the binary to see the change reflected. KV Watcher would fix that in a single-instance world too.

**Two paths:**

1. **Rebase + reframe** as `feat(pipes): hot-reload pipe definitions via NATS KV Watcher`. Same code shape, different framing — drop the "cluster" wording, refocus on the live-update use case. Closes #33 (KV Watcher Real Time Updates) and is a real win.
2. **Close** if no near-term need for live pipe updates.

Currently failing E2E + REVIEW_REQUIRED + DRAFT, so deferred from alpha either way. Recommend (1) post-alpha unless there's user demand sooner. Setting Priority to P3 on the board to reflect.

@EricAndrechek / @taitelee — your call which path.

PR #88, opened 2026-04-30
feat(mq)!: thread context through Ack/Nak with synchronous DoubleAck

### LINKED ISSUE #35: Context Aware MQ
## Problem

The Ack() and Nak() methods in our message queue implementation currently lack context.Context. This means they cannot respect request timeouts or graceful shutdown signals, potentially leading to "hanging" operations if the connection to NATS is interrupted.

## Proposed Solution

Refactor the mq interface to include ctx context.Context in all acknowledgment methods. This ensures that every interaction with the message broker is bound by the application's lifecycle and timeout policies.

_Taite can take this_

## Alternatives Considered

None

## Additional Context

None

taitelee:
# Pull Request

## Description

1. Interface Refactor: Updated mq.Message.Ack and mq.Message.Nak to accept context.Context and return an error.
2. Reliability Upgrade: Switched from asynchronous Ack() to synchronous DoubleAck(ctx) in the NATS implementations (embedded.go and remote.go). This makes sure WaveHouse receives a confirmation "receipt" from the NATS server before proceeding.
3. Graceful Shutdown Integration: Wired the application's global shutdown context from main.go into the MQ subscription handlers.
4. Bento Alignment: Updated the jsInput in internal/ingest/bento.go to use DoubleAck(ctx), ensuring background ingestion workers are connected to the 30-second shutdown timer when triggered.

Without this change, if the network connection to NATS was interrupted during a shutdown, the application would "hang" indefinitely. The Ack(ctx) call now acts as a Kill Switch: it will attempt to acknowledge the message, but if the shutdown timer expires first, it will instantly abort the network call and allow the application to exit cleanly.

## Type of Change

- [X] Bug fix (non-breaking change that fixes an issue)
- [X] New feature (non-breaking change that adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] My code follows the project's code style (`make lint` passes)
- [ ] I have added tests that prove my fix/feature works
- [ ] All existing tests pass (`make test`)
- [ ] All binaries compile (`make build`)
- [ ] I have updated the documentation accordingly
- [ ] I have updated the CHANGELOG.md (if applicable)

## Related Issues

Closes #35 

## Screenshots / Logs

If applicable, add screenshots or log output to help explain the change.


internal/mq/embedded.go
```diff
@@ -129,7 +129,14 @@ func (e *EmbeddedNATS) Subscribe(ctx context.Context, subject, consumerName stri
 	cctx, err := cons.Consume(func(m jetstream.Msg) {
 		msgCtx := observability.ExtractNATS(context.Background(), m)
```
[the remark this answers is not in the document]
taitelee: Swapped context.Background() with the ctx passed into the Subscribe method to ensure the trace propagation and cancellation signals properly cascade through the NATS headers.

internal/mq/mq_test.go
```diff
@@ -21,16 +22,16 @@ func TestMessage_AckNak(t *testing.T) {
 		"ingest.x",
 		[]byte("hi"),
 		time.Unix(1000, 0),
-		func() { acked++ },
-		func() { naked++ },
+		func(ctx context.Context) error { acked++; return nil },
+		func(ctx context.Context) error { acked++; return nil },
```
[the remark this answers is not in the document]
taitelee: Corrected the copy-paste error in the test callback to correctly increment the naked variable.

internal/ingest/bento.go
```diff
@@ -224,6 +246,93 @@ func (d *dlqOutput) WriteBatch(ctx context.Context, batch service.MessageBatch)
...
+	}
+
+	// Building the HTTP request
+	url := fmt.Sprintf("http://%s:%s/?database=%s&query=INSERT+INTO+%s+FORMAT+JSONEachRow&input_format_skip_unknown_fields=1&date_time_input_format=best_effort",
+		c.host, c.port, c.db, tableName)
+
+	req, err := http.NewRequestWithContext(reqCtx, "POST", url, &buf)
+	if err != nil {
+		chSpan.RecordError(err)
+		return err
```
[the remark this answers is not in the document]
taitelee: Wrapped the NewRequestWithContext error with %w to preserve the underlying error trace per the style guide.

internal/ingest/bento.go
```diff
@@ -224,6 +246,93 @@ func (d *dlqOutput) WriteBatch(ctx context.Context, batch service.MessageBatch)
...
+			chSpan.RecordError(err)
+			continue
+		}
+		buf.Write(data)
+		buf.WriteString("\n") // ClickHouse JSONEachRow requires newline separation
+	}
+
+	// Building the HTTP request
+	url := fmt.Sprintf("http://%s:%s/?database=%s&query=INSERT+INTO+%s+FORMAT+JSONEachRow&input_format_skip_unknown_fields=1&date_time_input_format=best_effort",
+		c.host, c.port, c.db, tableName)
```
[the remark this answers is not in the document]
taitelee: Added the mandatory safeIdentifierRe validation before constructing the HTTP request, and refactored the URL construction to use url.Values for safe encoding of the query string and database parameters, while strictly formatting the table name using backticks (ClickHouse format).

taitelee:
@gemini-code-assist

internal/ingest/bento.go
```diff
@@ -65,37 +75,39 @@ func (j *jsInput) Read(ctx context.Context) (*service.Message, service.AckFunc,
 			return nil, nil, service.ErrNotConnected
 		}
```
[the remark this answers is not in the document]
taitelee: Bento handles plugin shutdown by explicitly calling `Close()` on the input component, which in turn calls` j.iter.Stop()`. Stopping the iterator cleanly unblocks `Next()` and causes it to return an error, which breaks the loop naturally. Adding a `ctx.Err()` poll here is redundant and works against the NATS iterator's built-in signaling.

internal/ingest/bento.go
```diff
@@ -65,37 +75,39 @@ func (j *jsInput) Read(ctx context.Context) (*service.Message, service.AckFunc,
...
 		if err := json.Unmarshal(m.Data(), &raw); err != nil {
-			slog.Error("rejecting message: invalid JSON", "error", err)
-			if ackErr := m.Ack(); ackErr != nil { // Drop — not retryable.
-				slog.Warn("ack failed for invalid JSON message", "error", ackErr)
+			slog.ErrorContext(msgCtx, "rejecting message: invalid JSON", "error", err)
+			if doubleAckErr := m.DoubleAck(ctx); doubleAckErr != nil {
+				slog.WarnContext(msgCtx, "double ack failed for dropped message", "error", doubleAckErr)
 			}
 			continue
```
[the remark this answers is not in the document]
taitelee: If the context cancels during `DoubleAck`, the shutdown signal is already being processed by the system. By logging and calling continue, we allow the loop to immediately hit `j.iter.Next()`, which will cleanly unblock with an iterator-closed error (triggered by Bento's `Close()` call). Returning an arbitrary error mid-loop here could disrupt Bento's expected shutdown sequence.

internal/ingest/bento.go
```diff
@@ -217,18 +285,154 @@ func (d *dlqOutput) WriteBatch(ctx context.Context, batch service.MessageBatch)
...
 // running stream for lifecycle management. Callers should call stream.Stop(ctx)
 // during graceful shutdown to drain in-flight batches. The provided ctx controls
 // the stream's lifetime — cancelling it initiates shutdown of the Bento pipeline.
-func StartIngestWorker(ctx context.Context, nc *nats.Conn, streamName string, chConn driver.Conn, chHost, chHTTPPort, chUser, chPassword, chDB string) (*service.Stream, error) {
+//
+// NOTE: Bento components are registered globally. The ClickHouse and NATS
+// configurations passed to this function are frozen during the first call per
+// process. Subsequent calls in the same process will silently use the original
+// configuration for the output plugins.
+func StartIngestWorker(ctx context.Context, nc *nats.Conn, streamName string, chConn driver.Conn, chHost, chHTTPPort, chHTTPScheme, chUser, chPassword, chDB string) (*service.Stream, error) {
```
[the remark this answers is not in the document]
taitelee: This was on purpose. I added this godoc note to acknowledge the process wide singleton behavior of Bento's plugin registry. Because this is an internal application component called exactly once in production, adding global state tracking just to fail fast in a multi-call testing scenario is unnecessary and unnecessarily complex.

internal/ingest/bento.go
```diff
@@ -321,9 +533,11 @@ output:
...
 
 	go func() {
-		logger.Info("ingest worker started")
+		logger.InfoContext(ctx, "ingest worker started")
 		if err := stream.Run(ctx); err != nil && !errors.Is(err, context.Canceled) {
-			logger.Error("ingest worker stopped", "error", err)
+			logger.ErrorContext(ctx, "ingest worker stopped unexpectedly", "error", err)
+			// Force a container restart so the API doesn't become a zombie
+			os.Exit(1)
 		}
```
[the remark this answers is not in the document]
taitelee: Nice. Removed os.Exit(1) in favor of an onFatal callback so main.go can initiate a proper graceful teardown without bypassing deferred cleanups.

internal/mq/mq_test.go
```diff
@@ -15,31 +16,35 @@ func TestStreamName(t *testing.T) {
...
 	assert.Equal(t, []byte("hi"), msg.Data)
 	assert.Equal(t, int64(1000), msg.Timestamp.Unix())
 
-	msg.Ack()
-	msg.Nak()
+	_ = msg.Ack(t.Context())
+	_ = msg.AsyncAck()
+	_ = msg.Nak(t.Context())
 	assert.Equal(t, 1, acked)
+	assert.Equal(t, 1, asyncAcked)
```
EricAndrechek: If async ack is truly async and not blocking, how do we know this test won't flake and will deterministically pass the assertion here and not some time later?

internal/mq/embedded.go
```diff
@@ -127,12 +127,27 @@ func (e *EmbeddedNATS) Subscribe(ctx context.Context, subject, consumerName stri
...
+		msgCtx := observability.ExtractNATS(ctx, m)
+
+		msg := NewMessage(msgCtx, m.Subject(), m.Data(), time.Now(),
+			func(ctx context.Context) error {
+				return m.DoubleAck(ctx)
+			},
+			func() error {
+				return m.Ack()
+			},
+			func(ctx context.Context) error {
```
EricAndrechek: If our Ack implementation for mq involves passing in context to perform a JetStream message DoubleAck(ctx) behind the scenes, whereas the async ack is now just a normal ack that doesn't need context, why does the Nak need context? It doesn't look like the JetStream message Nak function needs a context passed, just like the case with Ack(), so why do we require it?

internal/ingest/bento.go
```diff
@@ -65,37 +74,39 @@ func (j *jsInput) Read(ctx context.Context) (*service.Message, service.AckFunc,
 			return nil, nil, service.ErrNotConnected
 		}
 
-		msgCtx := observability.ExtractNATS(context.Background(), m)
+		msgCtx := observability.ExtractNATS(ctx, m)
```
EricAndrechek, line 78: Could you help me understand why we are passing in plain `ctx` now instead of `context.Background()`?

internal/ingest/bento.go
```diff
@@ -135,54 +178,78 @@ func (j *jsInput) Read(ctx context.Context) (*service.Message, service.AckFunc,
...
 
-				bentoEventsProcessed.Add(ctx, 1, metric.WithAttributes(
+				bentoEventsProcessed.Add(ackCtx, 1, metric.WithAttributes(
 					attribute.String("table", raw.TableName),
 				))
 			}
 
-			bufferSpan.End()
-
 			if err != nil {
```
EricAndrechek: Why a separate `if err != nil` block? Why not just have the part above say:
```suggestion
if err != nil {
  slog.ErrorContext(msgCtx, "batch processing failed", "error", err)

  // NATS jetstream.Msg.Nak() is a non-blocking, fire-and-forget network call
  // that does not accept a context. The lack of context here is intentional.
  return m.Nak()
} else {
  slog.InfoContext(msgCtx, "message batch successfully acknowledged by ClickHouse")

  bentoEventsProcessed.Add(ackCtx, 1, metric.WithAttributes(
    attribute.String("table", raw.TableName),
  ))

  return m.DoubleAck(ackCtx)
}
```

internal/ingest/bento.go
```diff
@@ -65,37 +74,39 @@ func (j *jsInput) Read(ctx context.Context) (*service.Message, service.AckFunc,
...
 		slog.InfoContext(msgCtx, "received message from JetStream", "subject", m.Subject())
 
 		var raw struct {
-			Action    string `json:"action"`
-			TableName string `json:"table_name"`
-			ID        string `json:"id"`
+			Action    string          `json:"action"`
+			TableName string          `json:"table_name"`
+			ID        string          `json:"id"`
+			Data      json.RawMessage `json:"data"` // Handles lowercase
```
EricAndrechek: Why `Data` and `DataCap`?? It also is hard to read quickly, when there is both `raw.Data` as well as `m.Data()`... different names would be good. My understanding (could be wrong) with the Go JSON parser though is that it is largely best-effort, and will handle the capitalization concern for us. But, is this `data` parameter in the JSON coming from the user (ie API data fed in raw), or is it being normalized and validated first somewhere? My thinking is that this interface should be strict and require a decision on `Data` vs `data` from its caller, assuming that is our own code. If it is passed in from user-code, there should be parsing and standardization before it reaches us here via the API layer.

EricAndrechek (changes requested):
Some thoughts:

- [x] Don't like the inconsistency between Ack, Nak, and DoubleAck in some places, and AsyncAck, Nak, and Ack in others. As I understand it, AsyncAck == Ack, and Ack == DoubleAck, which is confusing for the developer.
- [x] Similar to above, unclear on which require context in which cases and which do not, and why. Seems that some Nak and Ack calls don't use context, but others require it...?

internal/mq/mq_test.go
taitelee: In this test case, we are solely testing whether or not our `Ack`, `Nak`, and `AsyncAck` wrappers properly adopt the functions that is passed in when constructing a new message. The `AsyncAck()` call here should properly increment the `asyncAcked` variable and when the assertion is checked `asyncAcked` would have been incremented. Nothing in this test case would have delayed the fire of the `AsyncAck()` call. This does make me wonder if we are sophisticated in our tests for the wrapper, and I will look into writing more test cases if I determine we fall short in thorough testing.

internal/ingest/bento.go
taitelee: We pass in plain ctx for the scope of graceful shutdown and tracing. By passing the worker's `ctx` instead of `context.Background()`, the newly created `msgCtx` inherits the worker's lifecycle signals. If the server initiates a graceful shutdown, `ctx` is cancelled, and `msgCtx` will respect that cancellation. If we anchored it to context.Background(), the message processing would be orphaned from our shutdown signals and could potentially hang indefinitely.

internal/mq/embedded.go
taitelee: You're right on this. Manually checking `ctx.Done()` before calling `m.Nak()` is an anti pattern since `m.Nak()` is non blocking and doesn't need cancellation protection. And even if the `Nak` is sent we'd want the redelivery of the message in the case a context is cancelled (like server shutdown). Will fix this.

internal/ingest/bento.go
taitelee: Yep. This is just poor design and redundant coding on my end. Will fix.

internal/ingest/bento.go
```diff
@@ -221,6 +221,10 @@ func (j *jsInput) Read(ctx context.Context) (*service.Message, service.AckFunc,
 
 		msg.MetaSet("table_name", raw.TableName)
 
+		if raw.ReceivedTimestamp != "" {
+			msg.MetaSet("received_timestamp", raw.ReceivedTimestamp)
+		}
```
EricAndrechek: Should we allow no received timestamp here, or would that not be asserted that the API/ingress handler MUST always attach it before passing to us here?

internal/ingest/bento.go
```diff
@@ -142,21 +144,17 @@ func (j *jsInput) Read(ctx context.Context) (*service.Message, service.AckFunc,
...
+							span.End()
+							flushCancel()
+							return nil, nil, ctxErr
 						}
 
+						slog.ErrorContext(spanCtx, "timeout waiting for in-flight inserts to drain",
+							"table", raw.TableName,
+							"id", raw.ID,
+						)
+						_ = m.Nak()
```
EricAndrechek: Why do we catch other `Nak()` errors (like below) but not this one? Just curious, doesn't necessarily need changing, not sure if this path was too loud, etc

internal/ingest/bento.go
```diff
@@ -375,6 +376,29 @@ func (c *clickhouseOutput) WriteBatch(ctx context.Context, batch service.Message
...
+
+				var extra string
+				if needsComma {
+					extra = fmt.Sprintf(`,"received_timestamp":%s}`, rtJSON)
+				} else {
+					extra = fmt.Sprintf(`"received_timestamp":%s}`, rtJSON)
+				}
+				data = append(data, []byte(extra)...)
+			}
+		}
```
EricAndrechek, line 406: What is this block for, what's it being added for?

EricAndrechek (commented):
Just a couple clarifications/questions, but nothing strictly blocking approval

internal/ingest/bento.go
taitelee: We can catch the error here. I will change it.

internal/ingest/bento.go
taitelee: The API/ingress handler does attach the timestamp in `ingest.go`. If it's there, we preserve the exact millisecond the API received it. If it's gone, we still save the data and let ClickHouse record the time it arrived at the database. ClickHouse will use whatever fallback we have like a default timestamp (`DEFAULT now64()`) or, if allowed, a null value, which means the record is saved rather than dropped. We avoid a strict assertion here just in case there are manually injected NATS messages or legacy data that might lack a timestamp.

internal/ingest/bento.go
taitelee: So here I was trying to include the `received_timestamp` metadata field from the NATS envelope within the actual data object being sent to ClickHouse. I wanted to avoid full JSON unmarshalling and marshalling to do that. This block just includes byte level slicing while maintaining efficiency and speed for inserts.

internal/ingest/bento.go
EricAndrechek: To confirm I'm understanding correctly, our API/ingress either uses the timestamp that was already included there, or adds one itself? So when the `raw` object gets to ingest here, it **always** has a `ReceivedTimestamp` field, no? I am just curious what the case is where this could even possibly be an empty string like this?

internal/ingest/bento.go
EricAndrechek: My opinion would be let's do this the "right" way with proper JSON parsing (unmarshalling and marshalling) to make it more legible and understandable and easy to test, plus we gain the safety benefits of Go's built-in JSON stuff. Then, down the line if we are seeing it really hurting performance, we can make a helper abstraction to do it in-memory without unmarshalling like this. Thoughts?

internal/ingest/bento.go
taitelee: There really is no case that a user would go through where the timestamp would not be included. Should I get rid of this safety check? If we as developers wanted to inject some message for debugging or while testing the timestamp could potentially not be included, and if they somehow needed it for testing, they would not see that they did not include it. If you're ok with that I can juts get rid of it.

internal/ingest/bento.go
taitelee: Sure. I have no opposition to either method.

taitelee: Regarding the move to json.Unmarshal/Marshal: We’ve decided to stick with the byte-splicing approach for the received_timestamp injection.

While JSON parsing is cleaner, Go's standard library coerces all numbers into float64 when unmarshaling into a map. This results in silent precision loss for 64-bit integers (IDs/Timestamps) exceeding 53 bits of mantissa. Since this is a high-volume ingest path for arbitrary user data, we cannot risk data corruption. I have updated the byte-splicing logic to be memory-safe by using a fresh allocation (make([]byte, ...)), ensuring we don't mutate the underlying message buffer.

internal/api/ingest.go
EricAndrechek, line 1: Should have a test case for this

EricAndrechek (changes requested):

internal/api/ingest.go
taitelee: Will add

taitelee:
@claude[agent] can you review this

taitelee:
@claude[agent] can you please review this

EricAndrechek (approved):

PR #89, opened 2026-04-30
chore: remove clustered mode and unscoped features

### LINKED ISSUE #38: E2E Integration Tests
## Problem

We have some unit tests, but we lack automated verification of the pipeline from an API POST request all the way to the data appearing in ClickHouse. We might miss bugs where the individual parts work but the "glue" between them fails.

## Proposed Solution

Build a suite of integration tests to verify data persistence and that data is properly formatted.

## Alternatives Considered

None

## Additional Context

None

### LINKED ISSUE #40: GoReleaser Setup
## Problem

Releasing new versions of WaveHouse is currently a manual process. This leads to inconsistent versioning, missing changelogs, and a lack of standardized binaries for different architectures (Linux, Mac, Windows).

## Proposed Solution

Integrate GoReleaser into our CI/CD pipeline. This will automate the creation of GitHub releases, generate standardized version strings (via Git tags), and build cross-platform binaries and Docker images automatically.

## Alternatives Considered

None

## Additional Context

None

### LINKED ISSUE #41: Multi-Arch Docker Support (amd64/arm64)
## Problem

Currently, our Docker images are likely tied to the architecture they were built on. This limits deployment flexibility, especially for users running on Apple Silicon (M1/M2) or ARM-based cloud instances (AWS Graviton).

## Proposed Solution

Describe the feature or change you'd like.

## Alternatives Considered

Any alternative approaches you've thought about.

## Additional Context

Any mockups, examples, or references that help explain the request.

### LINKED ISSUE #45: Makefile & CI/CD Pipeline Update
## Problem

The build and release processes are fragmented. We need a "Single Source of Truth" for how the project is tested, built, and deployed.

## Proposed Solution

Standardize the Makefile with targets like make release (triggering GoReleaser) and make test-e2e. Update GitHub Actions to automate the entire flow: testing, building multi-arch images, and drafting GitHub releases upon new Git tags.

## Alternatives Considered

None

## Additional Context

None

### LINKED ISSUE #87: chore(mvp): cut half-baked features for mvp focus
As part of the discussion in #73, we need to remove some features that are currently half-baked at best in order to improve our focus on a core subset of features to let us get a shippable v1 released. This includes:
- [ ] Cutting support for on-the-fly SQL queries/streaming from the server from NATS/buffered data – only stream raw events (row insertions, etc) as they are received, no attempting to aggregate with CH data to give updated SUMs, etc. TBD on if basic WHERE filters will be allowed/applied or not, as we don't want to build a custom CH SQL parser but also understand we need some level of filtering, for both security purposes (ie restricted table or column access) and for tables with high write volumes that we cannot afford to stream raw.
- [x] Removing the distributed/clustered approach for the time being. The plan will be to focus on the standalone ONLY, helping us speed up build times (fewer test cases and release versions) and remove complexity. If teams need more reliability than standalone (one server) can provide, even on k8s or similar, they can use our managed service at [wavehouse.cloud](https://wavehouse.cloud). We will, however, still build all component integrations with highly abstracted + modularized frameworks, like we do for our dedupe layer, so that adding other DB options or extending to cluster modes in the future doesn't require a large scale refactor.
- [x] Remove broken Github actions workflows and jobs and simplify existing ones to be more efficient + better mirror the makefile
- [x] Remove now-deprecated config options for cluster mode and other config cleanup where applicable
- [ ] Simplify hub fanout process (wildcards) given reduced complexity for current MVP plans
- [x] Remove stream name config options and implementation in MQ – not needed when only MQ impl. is embedded, will re-add for remote NATS when needed

### LINKED ISSUE #99: feat(mq): warn at startup when embedded JetStream store dir is non-persistent
In `cmd/wavehouse` standalone mode, with no volume mounted, embedded JetStream
silently writes to `./data/nats/jetstream` inside the container's writable overlay
layer — visible only as `Store Directory: "data/nats/jetstream"` in startup logs,
with no warning that the path isn't persistent. Every container recreation
(Portainer redeploy, image rebuild, healthcheck restart) wipes JetStream state:
in-flight events between an ingest POST and the Bento→ClickHouse flush are lost,
SSE/WS gap-fill stops bridging restarts, and NATS retention (`WH_MQ_MAX_BYTES_GB`,
default 50) accumulates inside `/var/lib/docker` instead of the operator's chosen
SSD tier. The reference compose mounts a volume so the docs *imply* persistence is
required — but a fresh user wiring their own compose can easily miss it.

## Proposed Solution

When the resolved `mq.embedded_dir` is a relative path **and** nothing is mounted at
it (or it resolves onto the container's overlay rather than a volume), log a startup
WARN. One line that catches the mistake before any data is lost.

## Alternatives Considered

- Make a persistent dir mandatory (hard-fail if not mounted) — too aggressive;
  ephemeral JetStream is legitimate for throwaway/dev stacks.
- Document-only — easy to miss; a runtime warning catches the case the docs already
  imply.

## Additional Context

From nas-observability `WHissues.md`, 2026-05-06. Mirrors the volume pattern in
`deployments/compose/standalone.yaml` (established by #42, Docker Compose:
Standalone vs. Clustered). Adjacent to #84 (JetStream `SyncAlways` durability) —
same `internal/mq/embedded.go` footgun territory, different axis: #84 is about the
sync-mode/throughput tradeoff, this is about the store dir silently being
ephemeral.

EricAndrechek:
## Summary

Removes features we are cutting out for MVP to move focus to the minimum surface possible to get a first version functional.

## Related Issues

Closes #87, #45, #40, #41, #38, #99


EricAndrechek:
/gemini review
@claude[agent] review this

jfwoods:
@claude[agent] What's the word on this?

internal/api/dlq.go
```diff
@@ -6,24 +6,24 @@ import (
...
-	return &DLQHandler{JS: js, DLQStreamName: baseStreamName + "_DLQ", Logger: logger}
+func NewDLQHandler(js jetstream.JetStream, logger *slog.Logger) *DLQHandler {
+	return &DLQHandler{JS: js, Logger: logger}
 }
 
 // Stats returns per-table message counts in the DLQ stream.
 // Supports optional ?table= query parameter to filter by table name.
 func (h *DLQHandler) Stats(w http.ResponseWriter, r *http.Request) {
-	stream, err := h.JS.Stream(r.Context(), h.DLQStreamName)
+	stream, err := h.JS.Stream(r.Context(), mq.DLQStreamName())
```
taitelee, line 26: Just confirming we are going to move away from this approach for stream naming and make it configurable?

tests/e2e/sdk/streaming.test.ts
taitelee, line 1: Does an issue exist for this? Commenting for reminder.

internal/api/dlq.go
EricAndrechek: Eventually for clustered mode/running on external NATS, yes, but for now have it resolve a value from the MQ provider so it's easier to replace with that in the future but doesn't need all the same custom string pass-throughs.

taitelee (approved):
Stripped away clustered version of WaveHouse and moved to a unified binary architecture. Added CI/CD workflow optimizations that include CI building a single Docker image so there's parity between testing and prod environments. Introduced ephemeral Docker containers for integration testing via Testcontainers, ensuring that the testing is physically isolated from the production data paths. A ton of internal cleanup was also included, especially around the Bento ingestion engine. Since Bento uses global registers that only allow a single startup per process, the internal wiring was refactored to use a 'Shared Environment' for the whole test suite. This kills off those 'already registered' crashes from the old monolithic file and lets the API and Worker run as siblings without fighting over resources. By moving that setup logic into TestMain, this built one big, reliable 'Conveyor Belt' for NATS and ClickHouse.

Some notes taken along the way:
- Standardized Volume Mounting: Cleaned up how the binary interacts with host storage. By moving to a unified `WH_DATA_DIR`, all internal state (NATS JetStream, Pebble dedupe, and ClickHouse data) is now organized into a predictable folder hierarchy. This makes managing persistent Docker volumes much simpler and prevents data from being scattered across the container filesystem.
- Orchestrator Logic: The logic that manages local containers was tightened up. The system now waits for specific health signals like the ClickHouse /ping HTTP endpoint before allowing the API to start. This prevents the "connection refused" errors that used to happen when the app tried to talk to a container that was still booting up.
- Automated Container Scoping: In the test suite, the move to Testcontainers means the life cycle of every container is now tied strictly to the Go process. No orphaned containers or dangling volumes are left behind to clog up the system's memory or disk space.

PR #92, opened 2026-05-06
feat(mq): make NATS stream name configurable (Re-Iteration)

### LINKED ISSUE #36: Configurable Stream Names

taitelee:
## Summary

This PR makes the NATS JetStream name configurable, enabling multi-environment isolation (e.g., separating Staging and Prod traffic). It replaces the previous hardcoded `mq.StreamName()` global helper with proper dependency injection to ensure the stream name propagates correctly through the entire system.

- Configurable Namespacing: Added `stream_name` to MQ configuration; supports YAML and env overrides (`WH_MQ_STREAM_NAME`).
- Refactored Injection: Updated all handlers (`SSE`, `WS`, `Sweeper`, `DLQ`) and constructors (`NewEmbedded`, `NewRemote`) to accept dynamic stream names.
- Removed Global Dependency: Removed `mq.StreamName()` global helper to enforce clean dependency injection and prevent hardcoding.
- Documentation & Deployment Sync: Updated `docs/configuration.md` examples and descriptions to match the dynamic config. Added missing env vars to `standalone.yaml` and `clustered.yaml`.
- Testing: Added regression test `TestEnsureDLQStream_CustomName` to explicitly verify dynamic stream derivation logic.

## Related Issues

Closes #36 

### Configuration Example
Via config.yaml:
`
mq:
  stream_name: "WH_STAGING"
  url: "nats://localhost:4222"
`


EricAndrechek:
Triage note as part of the pre-alpha priority pass.

#89 deliberately removed the configurable stream name because the previous implementation (#36) shipped a silent drift bug: `cfg.MQ.StreamName` was threaded into `mq.NewEmbedded`, `api.EnsureDLQStream`, `ingest.StartIngestWorker`, and `api.NewDLQHandler`, but the **Sweeper, SSE handler, and WS handler** all called the hardcoded `mq.StreamName()` function (returning `"WAVEHOUSE"`) instead. Setting `WH_MQ_STREAM_NAME=PROD` would have created the stream as `PROD` but left the Sweeper looking up `WAVEHOUSE` — silent state-sync drift, as documented in the #89 changelog entry.

Since WaveHouse is now fully embedded (each instance has its own NATS process), there is **no other tenant to namespace against** — the configurable stream name was an artifact of the external-NATS era. Reintroducing it would need either (a) a concrete use case where multiple embedded WaveHouse instances share a NATS process (which contradicts the "embedded NATS per binary" architecture), or (b) a multi-tenant story where one binary serves several logical streams (not currently planned).

Without that, this PR re-introduces a config knob with no use case and a known footgun history.

**Recommend: close this PR as superseded by the #89 architectural decision.** Reopen with a fresh issue + concrete use case if one emerges. Setting Priority to P3 on the board until then.

@taitelee — you opened this re-iteration; is there a use case I'm missing? If not, I'll close.

PR #114, opened 2026-05-12
fix(ci): dependabot major-bump comment uses marker-based upsert

EricAndrechek:
## What

The "Comment on major bumps" step in \`dependabot-automerge.yml\` calls \`gh pr comment\` unconditionally on every workflow run. Dependabot re-synchronizes its PRs (lockfile resolution, version metadata refetches), and each synchronize re-fires the workflow → a fresh duplicate of the same hold-for-review comment each time.

PR #106 currently has five identical copies of this comment from the same Dependabot bump cycle.

## Fix

Apply the same marker-comment upsert that \`housekeeping.yml\` already uses for its title-lint comment: stash a hidden marker (\`<!-- dependabot-major-bump-comment -->\`) in the body, look it up on the issue's comments list, and PATCH the existing comment in place. New comment only on the first run for the PR.

## Test plan

- [ ] Next Dependabot major-version bump: one comment, edited in place as the PR re-syncs (not duplicated).

### DIFF
--- a/.github/workflows/dependabot-automerge.yml
+++ b/.github/workflows/dependabot-automerge.yml
@@ -116,15 +116,38 @@ jobs:
         if: steps.meta.outputs.update-type == 'version-update:semver-major'
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          PR_URL: ${{ github.event.pull_request.html_url }}
+          REPO: ${{ github.repository }}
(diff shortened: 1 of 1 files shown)


EricAndrechek:
Superseded by #115, which contains this exact marker-comment upsert change verbatim inside a larger orchestrator-removal rewrite of `dependabot-automerge.yml`. Closing to avoid two merges of the same fix.

— Co-authored with Claude Opus 4.7 (1M context)

PR #172, opened 2026-05-20
feat(rbac)!: fail-closed authorization + default_role public access

### LINKED ISSUE #145: feat(auth): RequireRoles middleware — fail closed, no permissive fallback
## Problem

The README and AGENTS.md (design decision #11) advertise "Hasura-style access control" with per-table, per-role column- and row-level permissions. The data plane for this is in place — `internal/policy/` evaluates JWT-claim-templated policies against `WAVEHOUSE_POLICY` KV — but **there is no `RequireRoles` middleware** that gates routes on a role list.

What exists today in `internal/api/middleware.go`:

- `JWTMiddleware` parses the token and validates the signature.
- `extractClaim()` reads the role string out of the configured claim path (default `role`).
- The role is placed in `r.Context()` under `ContextKeyRole`.

What does **not** exist: any per-route check that says "this endpoint requires role X / Y / Z." Handlers can pull `RoleFromContext(r)` and gate themselves, but there's no enforcement primitive, no central policy, and no fail-closed default for routes that *forget* to gate. Shipping alpha with the public docs claiming RBAC enforcement, while the only enforcement is whatever the policy evaluator (or individual handlers) happen to do, is a credibility gap.

## Proposed Solution

Build the middleware fail-closed from day one:

- New `RequireRoles(roles ...string) func(http.Handler) http.Handler` in `internal/api/middleware.go`.
- On entry: extract role from context (must already have been populated by `JWTMiddleware`).
- If the role is missing/empty, or doesn't match any of the allowed roles, return `403` with `writeJSONError`.
- If `auth.dev_mode: true`, bypass entirely (explicit configured intent, not a fallback).
- If `auth.enabled: false`, also bypass (consistent with the rest of the auth stack).
- Log a structured `WARN` on every denial (`reason`, `role_observed`, `roles_allowed`, `route`) so operators see misconfiguration immediately.

Then apply it to the routes that need it. Suggested initial scope:

- `/v1/query` (raw SQL) — admin role only.
- `/v1/dlq/*` — admin role only.
- `/v1/policy/*` — admin role only.
- `/v1/pipes/*` (mutations) — admin role only; reads can stay open or use the named-pipe's own role gate.
- Ingest, structured query, SSE/WS — continue to rely on policy evaluator for per-table/per-row checks (this middleware is a coarser outer ring, not a replacement).

## Tests

For each route: token-with-allowed-role → 200, token-with-wrong-role → 403, no-token (auth on) → 401, no-token (auth off) → 200, no-token (dev mode) → 200. Plus a fuzz-y "every role list permutation" table-driven test for the middleware itself.

## Breaking change implications

For deployments that have `auth.enabled: true` today but no `auth.role_claim` configured (or a claim path that doesn't resolve), routes that newly require a role will start returning `403`. Document under "Breaking changes" in CHANGELOG with the exact config remediation. Confirm against current deployments before tagging alpha.

## Sequencing

Blocks v0.1.0-alpha.1 (#149). The README RBAC pitch and the actual enforcement need to match before public launch.

## Related

- AGENTS.md design decision #7 (optional auth with JWKS) and #11 (Hasura-style access control).
- `internal/policy/` — the per-row/per-column engine that already exists.
- `internal/api/middleware.go:122-126` — where the role currently lands in context.

### LINKED ISSUE #159: fix(pipes): per-pipe AllowedRoles fails open when context role is empty
## Problem

`PipesHandler.Execute` in `internal/api/pipes.go:86-102` enforces the per-pipe `AllowedRoles` allowlist only when the request context already carries a non-empty role. If the role is empty, the entire `AllowedRoles` check is skipped and the request is served.

```go
// internal/api/pipes.go:86-102
if len(q.AllowedRoles) > 0 {
    role := RoleFromContext(r.Context())
    if role != "" {                       // <-- guard
        allowed := false
        for _, ar := range q.AllowedRoles {
            if ar == role || ar == "*" {
                allowed = true
                break
            }
        }
        if !allowed {
            writeJSONError(w, http.StatusForbidden, "forbidden")
            return
        }
    }
    // empty role with AllowedRoles set → falls through and serves the pipe
}
```

This is a fail-open per-pipe authorization bug. A pipe configured with `AllowedRoles: ["admin"]` is reachable by any request whose context lacks a role under any of these conditions:

- `auth.enabled: false` (no `JWTAuthMiddleware` populates the role at all).
- `auth.enabled: true` but the JWT lacks the configured `auth.role_claim` (or the claim path doesn't resolve). `extractClaim` returns `""`, `ContextKeyRole` is set to `""`, the gate falls through.
- `auth.dev_mode: true` — `JWTAuthMiddleware` sets role to `"admin"` for all requests, so this configuration happens to match `AllowedRoles: ["admin"]` and is not a vector. But any non-`admin` allowlist (e.g. `["service"]`, `["analyst"]`) is bypassed.

Pipe execute routes (`GET/POST /v1/pipes/{name}`) sit **outside** the `/v1/admin/*` `RequireRole(authEnabled, "admin", "service")` block (`internal/api/router.go:115-119` vs `:126-158`). Per-pipe `AllowedRoles` is the only authorization gate on the execute path — there is no outer middleware that catches the empty-role case for these routes.

## Why this wasn't caught by the existing tests

The existing pipe handler tests cover three role scenarios (`internal/api/pipes_test.go`):

| Test | `AllowedRoles` | Context role | Expected | Actually passes |
| ---- | -------------- | ------------ | -------- | --------------- |
| `TestPipesHandler_Execute_RoleForbidden` (L97) | `["admin"]` | `"viewer"` | 403 | yes |
| `TestPipesHandler_Execute_RoleAllowed` (L121) | `["admin"]` | `"admin"` | not 403 | yes |
| `TestPipesHandler_Execute_WildcardRole` (L145) | `["*"]` | `"viewer"` | not 403 | yes |
| `TestPipesHandler_Execute_NoRolesAllowsAll` (L309) | `nil` | `"random_role"` | not 403 | yes |

The missing row is the one that exposes the bug:

| (missing) | `["admin"]` | `""` (or absent) | **should be 403** | **currently serves the pipe** |

Every existing test sets `ContextKeyRole` to a non-empty string, so the `if role != ""` branch is always taken; the implicit `else` branch (the actual production code path when the JWT lacks a role) has no test coverage at all. The `NoRolesAllowsAll` test name is misleading — it tests the case where the *pipe's* allowed roles is `nil`, not where the *request's* role is empty.

This is the actionable framing of the test-coverage failure: **every branch of an authorization check needs an explicit negative test, and the `(restricted pipe, empty role)` tuple was the one nobody wrote**. The bug is one line of code; the test gap is the systemic problem.

## Proposed implementation order (tests-first)

The intent is for the failing test to land first so the bug is demonstrably present in the test record, then the fix flips it green.

### Step 1 — add the failing test (red)

Add to `internal/api/pipes_test.go`:

```go
func TestPipesHandler_Execute_RestrictedPipe_EmptyRoleDenied(t *testing.T) {
    t.Parallel()
    store := pipes.NewMemoryStore(
        &pipes.NamedQuery{
            Name:         "admin_report",
            SQL:          "SELECT * FROM clicks",
            AllowedRoles: []string{"admin"},
        },
    )
    h := NewPipesHandler(store, nil, nil, 0)

    w := httptest.NewRecorder()
    // No ContextKeyRole set — simulates auth-disabled or JWT-without-role-claim.
    r := pipesRequest(t, http.MethodPost, "/v1/pipes/admin_report/execute", "admin_report", nil)

    h.Execute(w, r)

    assert.Equal(t, http.StatusForbidden, w.Code,
        "pipe restricted to %v must reject requests with no role in context", []string{"admin"})
    assertJSONErrorResponse(t, w)
}
```

This should fail against `main`: the response will not be 403 (it'll reach `executeQuery` and 500 on the nil ClickHouse connection, or some other non-403 status).

While there, convert the four existing role tests into a single table-driven matrix that makes the gap structurally hard to recreate:

```go
func TestPipesHandler_Execute_RoleAuthorization(t *testing.T) {
    t.Parallel()
    cases := []struct {
        name           string
        allowedRoles   []string
        contextRole    string // "" means no role set in context
        setRole        bool   // distinguishes "no role" from "empty role"
        wantForbidden  bool
    }{
        {"open pipe, no role", nil, "", false, false},
        {"open pipe, empty role", nil, "", true, false},
        {"open pipe, any role", nil, "anyone", true, false},
        {"restricted pipe, matching role", []string{"admin"}, "admin", true, false},
        {"restricted pipe, non-matching role", []string{"admin"}, "viewer", true, true},
        {"restricted pipe, NO role in context", []string{"admin"}, "", false, true},  // <-- the bug
        {"restricted pipe, EMPTY role in context", []string{"admin"}, "", true, true}, // <-- the bug
        {"wildcard, any role", []string{"*"}, "viewer", true, false},
        {"wildcard, no role", []string{"*"}, "", false, false}, // wildcard intentionally permissive
        {"multi-role allowlist, matching", []string{"admin", "editor"}, "editor", true, false},
        {"multi-role allowlist, non-matching", []string{"admin", "editor"}, "viewer", true, true},
    }
    for _, tc := range cases {
        t.Run(tc.name, func(t *testing.T) {
            // ...
        })
    }
}
```

The two `// <-- the bug` rows are the new coverage; the rest preserve the existing assertions. Decide explicitly whether `wildcard, no role` is intended to permit (current behavior) or deny — this issue does not pre-judge it, but the test should pin whichever is chosen.

### Step 2 — fix the handler (green)

Close the implicit-else branch in `internal/api/pipes.go:86-102`. The minimal fix is to drop the `if role != ""` guard so an empty role is treated as "no role matches, deny":

```go
if len(q.AllowedRoles) > 0 {
    role := RoleFromContext(r.Context())
    allowed := false
    for _, ar := range q.AllowedRoles {
        if ar == role || ar == "*" {
            allowed = true
            break
        }
    }
    if !allowed {
        writeJSONError(w, http.StatusForbidden, "forbidden")
        return
    }
}
```

Note that `ar == role` with `role = ""` will not match any non-empty role string, so an empty allowlist entry (which shouldn't exist but `Put` doesn't validate against it today) would still match an empty role. Either validate `AllowedRoles` entries on `Put` to reject empty strings, or skip empty entries during comparison. Implementation choice — call it out in the PR.

Status code choice: `RequireRole` returns 401 for empty-role-with-auth-enabled and 403 for role-mismatch. The per-pipe gate has no access to `authEnabled`, so 403 (forbidden) is the consistent answer here — the auth layer's job is to put a role in context; if there isn't one, the gate just sees "not in allowlist." If a stronger 401 distinction is wanted, that requires plumbing `authEnabled` into `PipesHandler`.

### Step 3 — audit for the same pattern elsewhere

This bug is shaped like a general anti-pattern: "if a sensitive value is missing, skip the check." Grep the codebase for the same shape and add tests where they're missing:

- `grep -rn 'if role != ""' internal/` — known instance: pipes.go:88. Verify others.
- `grep -rn 'if .* != "" {' internal/api/` for any handler doing security-relevant string comparison gated on non-empty.
- Cross-check `internal/policy/` evaluation against empty `role`/empty JWT claims.

Document any others found as separate issues or in this PR.

### Step 4 — test-framework prevention

The systemic gap is that role-gating code has no enforced minimum coverage shape. Two complementary changes:

1. **Shared auth-matrix helper in `internal/testutil/`.** Add `testutil.RunRoleMatrix(t, handler, cases)` (or similar) that runs a standard (allowed_roles, observed_role) → expected matrix. Adopt it in `pipes_test.go` and any other per-resource role-gated handler. Goal: writing a handler that takes `AllowedRoles` without invoking the matrix should look obviously incomplete in review.
2. **Coverage gate on `internal/api/pipes.go` (and other auth-touching files).** The current `.testcoverage.yml` enforces 70% on unit. Pipes-specific minimum should be higher (90%+) — the file is small enough that 70% can miss a whole branch like this one. Alternative or addition: a per-package gate for `internal/api/` that requires every branch in handlers tagged with role logic to be covered. `go-test-coverage` supports per-file thresholds.

Optional: explore mutation testing (e.g. `go-mutesting`) on `internal/api/pipes.go` — the `if role != ""` guard is the textbook case a mutation tester catches (negate the condition → tests should fail; here, they wouldn't).

### Step 5 — docs

- `CHANGELOG.md` under `[Unreleased]` — security fix entry.
- `docs/api.md` — if pipe `AllowedRoles` is documented, note the corrected fail-closed semantics.
- `AGENTS.md` — if design decision #11 references pipe-level role enforcement, ensure the description matches the fixed behavior.

## Security impact

- **Severity**: medium-to-high depending on deployment. Most exposed configuration is `auth.enabled: true` with a JWT issuer that doesn't populate `auth.role_claim` — pipes intended for `admin`-only access are reachable by any authenticated principal.
- **Not exploitable**: dev_mode (role is forced to `"admin"`, only matches `["admin"]` allowlists).
- **Trivially exploitable**: `auth.enabled: false` with `AllowedRoles` configured — every pipe is open. This is documented as "no auth, no enforcement" so it's a design-consistency call rather than a strict CVE, but the per-pipe allowlist's presence suggests intent that isn't honored.

No public disclosure recommended ahead of fix since the project is pre-1.0 alpha; flag in CHANGELOG once merged.

## Out of scope

- Building a route-level `RequireRoles` for pipe **execute** path. The model is per-pipe `AllowedRoles`, and #145 explicitly leaves reads/executes on that model. Don't expand scope to add an outer middleware unless that design decision is revisited separately.
- Renaming/consolidating `RequireRole` (singular, exists in `router.go:209`) with the proposed `RequireRoles` in #145. Separate concern.

## Related

- #32 — Native SQL Templating. This issue was discovered while reviewing `pipes.go` for #32 and noted as a side observation in the issue body.
- #145 — `RequireRoles` middleware. Adjacent but does not cover this case: #145's "reads can stay open or use the named-pipe's own role gate" rests on the assumption that the per-pipe gate works; this issue is that it doesn't.
- `internal/api/router.go:209` — existing `RequireRole` middleware, which fails closed correctly for the routes it protects.
- `internal/api/middleware.go:122-126` — where `ContextKeyRole` is populated (or left empty) by `JWTAuthMiddleware`.
- `internal/api/middleware.go:46-61` — auth-disabled and dev-mode short-circuits (relevant for understanding when role is empty vs `"admin"`).
- `internal/testutil/jwt.go` — `MakeJWT` helper, useful for end-to-end auth tests if a router-level integration test is added.

taitelee:
## Summary
Started as the #159 fail-open fix in pipe `AllowedRoles` and grew into a fail-closed pass over the whole RBAC surface, plus opt-in public access. Authorization is now uniformly **allowlist membership**: a caller's role (a tokenless/roleless request is first resolved to the policy `default_role`) must appear in the relevant allowlist, the built-in **`admin` role is the only bypass**, and the `*` any-role wildcard is gone.

### Pipe authorization — `internal/api/pipes.go` (Closes #159)
The original bug: `Execute` enforced `AllowedRoles` only when the request carried a non-empty role, so an empty/absent role (a JWT missing `auth.role_claim`, or a tokenless request) skipped the gate and a restricted pipe was served — a fail-open authorization bug, and `AllowedRoles` is the only gate on the execute path.

`Execute` now runs one path for every request:
1. Read the role; if empty, resolve it to the policy `default_role` (via `PolicyStore`/`ResolveRole`).
2. The admin role (`policy.AdminRole`) always passes.
3. Otherwise the role must be an **exact** match in `allowed_roles`; empty-string entries are skipped (a stray `""` can't authorize a roleless caller).

- **BREAKING:** a pipe with **no `allowed_roles`** now authorizes nobody but the admin role (previously it was open / "still required a token" under public access). Public reach is opt-in only by listing a role the `default_role` resolves to.
- **BREAKING:** the admin role bypasses **every** pipe, including restricted ones, consistent with `policy.Evaluate` and the `/v1/admin/*` `RequireAdmin` gate.
- Removed the now-unused `PipesHandler.AllowAnonymous` field/branch.

### Policy engine — `internal/policy/`
- `Evaluate`: an empty/absent role never matches a role key; the `"*"` any-role role wildcard is **removed** (column allow-lists, e.g. `allow_columns: ["*"]`, are unaffected); the admin-role bypass is retained. Roleless access is granted only via `default_role` substitution. A **nil policy** (none configured yet, or deleted from KV) fails closed in `Evaluate` itself — only the admin role passes — so deleting the policy denies all non-admin traffic on its own.
- `Validate`: rejects empty role keys, and a `default_role` equal to the admin role (it would hand every roleless request full admin).

### Public / anonymous access via `default_role` — `internal/auth/auth.go`, `router.go`, `main.go`, `errors.go`
- No config flag, and no auth on/off switch: the auth middleware always runs, authenticates the Bearer token if present, and **never rejects** — authentication is decoupled from authorization (`internal/auth`). A no-token request, or one whose token lacks the role claim, carries an **empty role** that `ResolveRole` maps to the policy `default_role` before evaluation.
- Public access is therefore driven entirely by the policy: define a usable (non-admin) `default_role` and anonymous callers reach whatever that role is authorized for; with no usable `default_role`, a roleless request has no role and is denied **403** (`"forbidden: request has no role and no public default_role is configured"`) by whatever gate it hits. A policy PUT/delete flips this live (no restart).
- A present-but-invalid/expired token still resolves to the empty role, but the auth middleware stashes the reason so a denying gate **fails loud with 401** (`token expired` / `invalid token`) rather than a bare 403.
- `default_role` is honored for authenticated-but-roleless requests too (valid token, no role claim).
- The admin surface is gated by a single **`RequireAdmin`** middleware (`router.go`), driven by the live policy admin role (`policy.AdminRole`, configurable via `admin_role`, default `admin`): `/v1/admin/*` (raw-SQL `/v1/admin/query`, policy, pipe mutations, log-level), plus `/v1/schema`, `/v1/schema/refresh`, and `/v1/dlq/stats` — the infra endpoints that have no policy gate of their own.
- `forbiddenForRole` returns a clearer 403 when the effective role is empty.

### Divergence from #159's plan
#159 (and the original PR body) treated the `"*"` wildcard as **kept** ("wildcard intentionally permissive") and left "wildcard, no role" as a decision. We went further and **removed the role wildcard entirely** so authorization is pure allowlist membership with the admin role as the only bypass — no implicit "any role" grant. The role-matrix test cases were updated accordingly.

### Tests & docs
- `internal/testutil/rolematrix.go`: shared `RunRoleMatrix`/`StandardRoleMatrix` — the canonical `(AllowedRoles, observed-role)` matrix every allowlist gate must pass; a handler taking `AllowedRoles` without running it looks under-tested in review.
- Expanded `internal/api/pipes_test.go` (no-allowlist deny/allow, default_role resolution, param paths), `internal/policy/policy_test.go` (empty-role guards, `*`-no-longer-grants regression, default_role substitution/escalation refusal), `internal/api/router_test.go`, `internal/auth/auth_test.go`, `internal/policy/store_test.go`, `internal/config/config_test.go`.
- `CHANGELOG.md`, `docs/api.md`, `docs/configuration.md`, `AGENTS.md` updated to the fail-closed / no-wildcard public-access semantics.

## Breaking changes
- A pipe with no `allowed_roles` denies all callers except the admin role (was open).
- The `"*"` any-role wildcard is removed from pipe `allowed_roles` and from `policy.Evaluate` role keys. Existing `"*"` role entries now match nothing (column `allow_columns: ["*"]` is unaffected).
- The admin role bypasses every pipe.

## Related issues
- Closes #159
- Out of scope (per #145): a route-level `RequireRoles(roles ...string)` primitive — the route layer stays admin-vs-not (`RequireAdmin`) and finer access stays per-pipe `allowed_roles` + policy.


internal/api/pipes_test.go
```diff
@@ -94,31 +94,69 @@ func TestPipesHandler_Execute_NotFound(t *testing.T) {
...
+		{"restricted pipe, matching role", []string{"admin"}, "admin", true, false},
+		{"restricted pipe, non-matching role", []string{"admin"}, "viewer", true, true},
+		{"restricted pipe, no role in context", []string{"admin"}, "", false, true},
+		{"restricted pipe, empty role in context", []string{"admin"}, "", true, true},
+		{"wildcard, any role", []string{"*"}, "viewer", true, false},
+		{"wildcard, no role", []string{"*"}, "", false, false}, // wildcard is intentionally public
+		{"multi-role allowlist, matching", []string{"admin", "editor"}, "editor", true, false},
+		{"multi-role allowlist, non-matching", []string{"admin", "editor"}, "viewer", true, true},
+		{"non-admin allowlist, no role in context", []string{"service"}, "", false, true},
+	}
```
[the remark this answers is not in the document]
taitelee: Yes. The description commits to ignoring empty allowlist entries, so the matrix should pin it. Adding: `{"empty allowlist entry, empty role", []string{""}, "", true, true}`

Like the other empty-role rows this is going to fail until the fix lands: dropping the `if role != ""` guard alone isn't sufficient because `"" == ""` would match, so the scan also skips empty entries (`ar != "" && ar == role`) to keep a malformed `[""]` allowlist fail-closed.

internal/api/pipes_test.go
```diff
@@ -94,31 +94,69 @@ func TestPipesHandler_Execute_NotFound(t *testing.T) {
...
+			safeHandle(h.Execute, w, r)
+
+			if tc.wantForbidden {
+				assert.Equal(t, http.StatusForbidden, w.Code,
+					"pipe restricted to %v must reject role=%q (set=%v)", tc.allowedRoles, tc.role, tc.setRole)
+				assertJSONErrorResponse(t, w)
+			} else {
+				assert.NotEqual(t, http.StatusForbidden, w.Code,
+					"pipe with AllowedRoles=%v must allow role=%q (set=%v)", tc.allowedRoles, tc.role, tc.setRole)
+				assert.NotEqual(t, http.StatusNotFound, w.Code)
```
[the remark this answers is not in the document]
taitelee: Keeping `assert.Equal` + `assertJSONErrorResponse` here, for a few reasons:
- `assertJSONErrorResponse` is the package's shared error-response helper. These tests follow the dominant pattern, so switching would make them the outlier.
- It's strictly stronger for error bodies: it also asserts `Content-Type: application/json` and `X-Content-Type-Options: nosniff`, which neither `testutil` helper checks. Switching would drop that coverage.
- The suggested call wouldn't compile — `AssertJSONContains`'s 4th arg is `map[string]any` (testutil.go:43), not string substring. (Heads-up: the AGENTS.md:137 description still says "substring" and is stale vs. the implementation — worth a separate doc fix.)
- The not-forbidden rows can't use these helpers anyway: they recover the nil-Conn panic via `safeHandle`, so the body is empty and `json.Unmarshal` inside the helper would fail. The per-row `%v/%q` failure messages also matter in a 12-row matrix.

So the tests already meet the intent (shared helper for HTTP assertions) with the better-fit one. Leaving as-is.

Claude did add some remarks to this comment.

docs/src/content/docs/api.md
```diff
@@ -322,7 +322,7 @@ JSON array of result rows, with `X-Cache: HIT` or `X-Cache: MISS` indicating whe
 | Status | Body | Cause |
 | ------ | ---- | ----- |
 | 404 | `{"error":"pipe not found"}` | Pipe name not registered |
-| 403 | `{"error":"forbidden"}` | Role not in pipe's `allowed_roles` |
+| 403 | `{"error":"forbidden"}` | Role not in pipe's `allowed_roles`. Fails closed: when `allowed_roles` is set, a request with no role (auth disabled, or a JWT missing `auth.role_claim`) is denied unless the list includes `"*"`. |
 | 400 | `{"error":"missing required parameter: x"}` | Required parameter not supplied |
```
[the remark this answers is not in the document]
taitelee: Added.

internal/policy/policy_test.go
```diff
@@ -324,6 +324,27 @@ func TestValidate_NegativeMaxExecutionTime(t *testing.T) {
...
+				Select: map[string]RolePermissions{
+					"": {AllowColumns: []string{"page"}},
+				},
+			},
+		},
+	}
+	err := Validate(p)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "empty role")
+}
```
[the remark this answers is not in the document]
taitelee: Fixed.

taitelee:
@EricAndrechek `pipes.Store.Put` doesn't reject empty `AllowedRoles` entries at write time, meaning that an empty role is accepted in a `pipes.Store.Put` call if the role is set to be empty in the context (I didn't look to deep into where this can actually occur). However, this PR fixes checking if the role is empty at read time in `pipes.Store.Execute`. So we still do block empty roles correctly for this PR, but I was just wondering if this fix is sufficient and that we don't need to block at write time.

AGENTS.md
```diff
@@ -34,13 +34,13 @@ Eleven internal packages under `internal/` (plus `internal/testutil/` for shared
...
+7. **Optional auth with JWKS**: JWT authentication is opt-in via `auth.enabled`. Supports HMAC shared secret and/or JWKS endpoint (`auth.jwks_url`). Roles are extracted from a configurable claim path (`auth.role_claim`). Dev mode (`auth.dev_mode`) bypasses validation for development. **Public (no-token) access is policy-driven, not a flag:** when auth is enabled, a request with **no** token is `401`'d unless the policy has a usable (non-`admin`/`service`) `default_role`, in which case it's admitted with an empty role that resolves to that `default_role`. The middleware decides admit-vs-`401` live via an injected `func() bool` wired to `policy.ResolveRole(store.Get(), "") != ""`, so a policy PUT/delete flips public access without a restart. A present-but-**invalid** token still `401`s; `/v1/admin/*` plus the token-only schema/DLQ routes stay gated by `RequireAuthenticated`; an unrestricted pipe (no `allowed_roles`) still requires a token while public access is on; and the policy cache is held fail-closed when auth is enabled (`Store.SetFailClosed` — a runtime KV delete is retained, not nilled) so `Evaluate`'s nil-policy allow-all can never expose tables to tokenless requests.
```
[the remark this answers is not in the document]
taitelee: Fixed.

cmd/wavehouse/main.go
```diff
@@ -259,6 +259,11 @@ func run() int {
 		return 1
 	}
 
+	// When auth is on, a no-token request is admitted only if a usable
+	// default_role is configured (see allowAnon below); hold the policy cache
+	// fail-closed so a runtime KV delete can't nil it into Evaluate's allow-all.
+	policyStore.SetFailClosed(cfg.Auth.Enabled)
```
EricAndrechek: Why not just move this into the `policy.NewStore()` function's init? Why set a separate function to set it?

internal/policy/policy.go
```diff
@@ -258,6 +283,10 @@ func Validate(p *Policy) error {
 	if p == nil {
 		return fmt.Errorf("policy is nil")
 	}
+	switch strings.ToLower(strings.TrimSpace(p.DefaultRole)) {
```
EricAndrechek: This is interesting – do we want `Admin` to be differentiable from `admin`? My leaning would be yes – we should allow roles to be whatever someone configuring them would want them to be, which could/may include differences in tenants or something by JWT role capitalization, so we should NOT treat them the same here either.

internal/policy/policy.go
```diff
@@ -56,12 +56,32 @@ type ResolvedPermissions struct {
...
+// ResolveRole maps an empty/absent role to the policy's default_role, so a
+// request with no role — a token without a role claim, or no token at all
+// when public access is enabled — is evaluated as the configured default. A default_role of "admin"/"service" is refused here
+// (returns "") so the privileged built-in bypass in Evaluate can never be
+// reached via the default; Validate also rejects saving such a policy.
+func ResolveRole(p *Policy, role string) string {
+	if role != "" || p == nil {
+		return role
+	}
+	switch strings.ToLower(strings.TrimSpace(p.DefaultRole)) {
```
EricAndrechek: See comment on `Validate`, but 3 notes:
1. Do we want `strings.ToLower` like this? In `Validate`'s comment I outlined why I think the answer is no. Similar question actually for `TrimSpace`... I could see a clearer argument for yes there, especially given roles may come from a YAML policy file which couldn't start with a space, but I still think when touching security like this we ***shouldn't*** be forgiving with "helpful" grace handling like this.
2. Do we actually want to disallow default role to be admin? I see a future where we could eliminate the dev and auth enabled flags and force auth to be on, but then allow admin to be a default for easy testing locally still. Thoughts?
3. There is some logic here (and elsewhere in this file) that is mirrored in validate for instance – is there a way we can cleanly merge all this so that there is only one source-of-truth, and there's no possibility for the two to diverge? (Unrelated tangent: we should expose a subcmd of our binary to validate a policy file, like `wavehouse validate config.yaml` or something like lots of other programs do – may be better in separate PR though)

cmd/wavehouse/main.go
taitelee: Yea it's probably better to do that I can make the fix for it.

cmd/wavehouse/main.go
```diff
@@ -356,6 +361,12 @@ func run() int {
 	wsHandler := api.NewWSHandler(hub, js, cfg.Server.CORSAllowedOrigins)
 	wsHandler.PolicyStore = policyStore
 
+	// allowAnon reports whether a no-token request should be admitted: true iff
+	// the org has configured a usable (non-privileged) default_role — that is
```
EricAndrechek: ok, but this function (as documented in the comment here) doesn't actually check that the configured "usable" role **is** non-privileged... if this is the gate, I'd feel better moving that privilege check to here then too. not to mention tbd on if that's what we want or not.

internal/policy/policy.go
taitelee: 1. The bypass in `Evaluate` is actually already exact match `(role == "admin" || role == "service")`, so a role of `Admin` or even something like `" admin "` never triggers it. In `Validate`, I'll get rid of  both `ToLower` and `TrimSpace` so we deny exactly the strings that would trigger the bypass. I guess there's no wrong to being strict
2. Maybe in future? I have no thoughts or opinions on this. I'm fine with whatever you think is the behavior that would be best for the users.
3. I can create a helper so that we have a singular point where we check if a role is privileged or not.

internal/policy/policy.go
taitelee: Sure I can change it.

internal/api/pipes.go
```diff
@@ -83,22 +85,34 @@ func (h *PipesHandler) Execute(w http.ResponseWriter, r *http.Request) {
...
+			if ar != "" && ar == role {
+				allowed = true
+				break
 			}
 		}
+		if !allowed {
+			writeJSONError(w, http.StatusForbidden, forbiddenForRole(role))
+			return
+		}
 	}
```
EricAndrechek: I don't like how many times + places we have duplicate code like this – I think a refactor may be in order to pull more of the auth checking, role/claim assertion/matching, etc, used for everything from pipes to queries etc to a separate package or something that they would all import and use. Thoughts?

taitelee: Yea again I can just isolate this kind of stuff to helpers in a separate package or something.

internal/api/router.go
```diff
@@ -255,6 +260,26 @@ func RequireRole(authEnabled bool, roles ...string) func(http.Handler) http.Hand
...
+			if authEnabled {
+				if _, ok := ClaimsFromContext(r.Context()); !ok {
+					writeJSONError(w, http.StatusUnauthorized, "unauthorized")
+					return
+				}
+			}
+			next.ServeHTTP(w, r)
+		})
+	}
+}
```
EricAndrechek: This is a good point, but my thinking would be that we should expose this sort of stuff under the same sort of config as other policy stuff, or leave it admin only for now if too big of a scope change.

internal/api/errors.go
```diff
@@ -19,3 +19,16 @@ func writeJSONError(w http.ResponseWriter, status int, message string) {
...
+// forbiddenForRole returns the 403 message for a policy/allowed_roles denial.
+// When the effective role is empty — a request with no token, or a token
+// without a role claim, and no default_role that permits the resource — it
+// says so, since that's the common public-access denial and "forbidden" alone
+// is opaque. For an ordinary mismatch (the caller has a role, just not a
+// permitted one) it stays terse.
+func forbiddenForRole(role string) string {
+	if role == "" {
+		return "forbidden: request has no role and no default_role permits this resource"
+	}
```
EricAndrechek, line 51: Does this case ever evaluate if there is a default role?

internal/policy/policy.go
EricAndrechek: Ok yeah, let's make sure its consistent then (which a helper would help with). And yeah, I do think that behavior is ideal – if I'm understanding correctly we could just fully remove the auth and dev flags in this PR then too, right?

internal/policy/policy.go
```diff
@@ -56,12 +56,32 @@ type ResolvedPermissions struct {
...
+	default:
+		return p.DefaultRole
+	}
+}
+
 // Evaluate resolves a policy for a given role, table, and operation against JWT claims.
 func Evaluate(p *Policy, role, table, operation string, claims map[string]any) *ResolvedPermissions {
 	if p == nil {
 		return &ResolvedPermissions{Allowed: true}
 	}
```
EricAndrechek, line 111: This seems incredibly unsafe, am I understanding it correctly?? Should this not be removed too as part of the security PR?

internal/policy/store.go
```diff
@@ -134,6 +132,29 @@ func (s *Store) Watch(ctx context.Context) {
 	}
 }
 
+// SetFailClosed controls how Watch reacts to a policy deletion from KV. When
+// true (set when auth is enabled), a delete is ignored: the last-known policy
+// is retained and an error logged, so Evaluate's nil-policy allow-all can
+// never be reached at runtime. Set once at startup, before Watch starts.
+func (s *Store) SetFailClosed(v bool) {
```
EricAndrechek: Want to talk about this more before deciding on it

internal/api/errors.go
taitelee: What do you mean? This is called if there is absolutely no role in the list of `AllowedRoles` allow the user of a `role` to make such a request regardless of whether or not they have the default role or not. I'm not sure what you're getting at here if you could clarify.

EricAndrechek (changes requested):
Changes requested, needs code cleanup for simpler policy evaluation codebase given we are turning this into one mega policy/RBAC PR

internal/api/errors.go
EricAndrechek: I guess I think the error message there is misleading, it suggests that a default role didn't permit that resource but if I am understanding it correctly, that only prints if there IS NOT default role, and the plain "forbidden" msg prints if the default role didn't have permissions

EricAndrechek:
Updates from discussion:
- moving policy/auth to own package to be applied in middleware automatically + any functions that need to check (code refactor/cleanup) to better maintain consistency across codebase for auth checks
- remove fail closed switch, deleting policy will do that anyway
- add an "AdminRole" policy config option, like we have for default role, then use that to check if admin perms and not string comparisons of the role to "admin" or "service"
- an invalid, malformed, expired, no role set/empty, etc JWT should fallback to the default role (but still fail loadly with the relevant error when requiring something with higher permissions)
- roles will NOT inherit the default role's policy – this will later be done via the policy config read in/engine and likely shown/validated in an admin panel UI
- remove auth enabled and dev mode flags, relying on default role and admin role instead
- make DLQ, Schema, etc permissions all exclusively admin-allowed for now

Future TODOs to be tracked in separate issues (need to make them still):
- enforce role inheritence from default in policy and pipe configs
- work out how to re-check JWT perms during a stream
- ensure we validate JWT expiration
- admin UI
- an admin deleting the policy via API is unrecoverable – must SSH into server and re-add policy and reboot manually (eventually would like to add a service or API admin directly for wavehouse control/config, not through JWT)
- add a validate subcommand to the binary for offline/local policy file validation + fmt checks
- add a method to allow access to schemas and refresh, dlqs, and other system paths for non-admins and specific roles/claims

docs/src/content/docs/api.md
```diff
@@ -25,13 +25,15 @@ GET /v1/stream/ws?token=<jwt>
...
 
-When `auth.dev_mode` is enabled, all requests are treated as admin with no JWT validation — useful for development.
+**Authentication is decoupled from authorization.** A request with **no token**, or an **invalid/expired/malformed** one, is *not* rejected outright — it falls back to an empty role that resolves to the policy `default_role`, and authorization is decided downstream. Because the bad-token reason is remembered, a request that is then denied for lacking permission fails loud (`401` "invalid/expired token") instead of a bare `403`. Elevated access requires a valid token whose role is granted (or equals the `admin_role`).
+
+**Public (unauthenticated) access is driven by the policy.** Define a usable `default_role` and no-token requests are evaluated as that role (see [Roles & Access Control](#roles--access-control)); remove it and roleless requests are denied. `default_role` may **not** equal the `admin_role`. `/v1/admin/*` **and** the schema/DLQ endpoints are admin-only, and a pipe with **no `allowed_roles` authorizes nobody but the admin role** (pipe access is allowlist membership — it never opens to the public).
```
EricAndrechek: I thought we DID want to allow default role to equal admin role, particularly for testing locally and in dev and whatnot?

EricAndrechek: Also for pipe access, can't it be open to the public ie if it allows the default role? How is that any different from structured queries and whatnot? I'd actually argue pipes are safer to allow to the public as they can be pre-written and defined SQL queries, whereas structured ASTs will have resource exhaustion risks, etc.

docs/src/content/docs/api.md
```diff
@@ -537,6 +539,8 @@ Replaces the entire access control policy. Validated before saving.
 }
 ```
 
+The `default_role` field (optional) is the role assigned to any request that reaches the policy engine **without** a role — a valid token carrying no role claim, a request with no token at all, or one whose token was invalid/expired. **Setting it enables unauthenticated access:** roleless requests are evaluated as that role and receive exactly its permissions (or are denied if it grants none on the table/operation). If `default_role` is unset, a roleless request is denied. It may **not** equal the `admin_role` — that is rejected at validation, so a roleless request can never inherit admin.
```
EricAndrechek: Same claim here, didn't we want to allow this, just to be loud about it when set etc.?

docs/src/content/docs/architecture.md
```diff
@@ -68,8 +68,8 @@ internal/
 
 The API layer uses [Chi](https://github.com/go-chi/chi) for routing with standard middleware (RequestID, RealIP, Recoverer).
 
-- **router.go** — Route definitions. Public: `/health`, `/ready`. Protected: `/v1/ingest/{table}`, `/v1/tables/{table}/query` (structured), `/v1/pipes/{name}` (named pipes), `/v1/stream/sse`, `/v1/stream/ws`, `/v1/schema/*`, `/v1/dlq/stats`. Admin (admin/service): `/v1/admin/policy`, `/v1/admin/pipes/*`, `/v1/admin/log-level`, `/v1/admin/query` (raw SQL — same gate as the rest of `/v1/admin/*`).
-- **middleware.go** — JWT auth middleware supporting HMAC and JWKS validation, role extraction from configurable claim path, and dev mode bypass. Controlled by `auth.enabled`.
+- **router.go** — Route definitions. Public: `/health`, `/ready`. Policy-gated: `/v1/ingest/{table}`, `/v1/tables/{table}/query` (structured), `/v1/pipes/{name}` (named pipes), `/v1/stream/sse`, `/v1/stream/ws`. Admin-only (`RequireAdmin`, role == `policy.admin_role`): `/v1/schema/*`, `/v1/dlq/stats`, `/v1/admin/policy`, `/v1/admin/pipes/*`, `/v1/admin/log-level`, `/v1/admin/query` (raw SQL — same gate as the rest of `/v1/admin/*`).
```
EricAndrechek: Shouldn't these be `/v1/ingest?table={table}` and `/v1/query?table={table}`? Were they missed in my PR? Can you fix them in this one?

docs/src/content/docs/architecture.md
```diff
@@ -105,7 +105,7 @@ The API layer uses [Chi](https://github.com/go-chi/chi) for routing with standar
...
... a Bento-based ingest pipeline: a JetStream input (`jsInput`) reads from the `WAVEHOUSE` stream via a durable `buffer-consumer` pull consumer, batches events per table, and performs bulk INSERTs to ClickHouse. The pipeline is **insert-only**. The wire format `EventMessage` carries `{table_name, scope, received_timestamp, data}` and nothing else; the worker validates the table name (via `safeIdentifierRe`) and the payload's presence, then bulk-INSERTs. The embedded NATS server runs with `DontListen: true` (`internal/mq/embedded.go`), so the only Publishers reachable on the `ingest.>` subjects are in-process Go code — today, only the HTTP `/v1/ingest/{table}` handler. Non-insert mutations (`DELETE`/`UPDATE`/`TRUNCATE`/…) must go through `POST /v1/admin/query` under the admin role (`policy.admin_role`) — see the Query Path section below; the `/v1/admin/*` `RequireAdmin` middleware enforces the check at the API layer, so a no/invalid-token request (resolved to `default_role`, never admin) never reaches the proxy. Failed batches are routed to a DLQ output (`dlqOutput`) which publishes the inner data payload (`{"id":"abc","field":...}`) to `dlq.{table}` NATS subjects when DLQ is enabled.
```
EricAndrechek: Here too, should be `/v1/ingest?table={table}`

docs/src/content/docs/configuration.md
```diff
@@ -83,11 +83,16 @@ export WH_CONFIG=/etc/wavehouse/config.yaml
...
 | `auth.role_claim` | `WH_AUTH_ROLE_CLAIM` | `role` | Dot-separated JWT claim path for role extraction (e.g., `app_metadata.role`). |
-| `auth.dev_mode` | `WH_AUTH_DEV_MODE` | `false` | When `true`, skips JWT validation and treats all requests as admin. **For development only.** |
+
+**There is no auth on/off switch.** The JWT middleware always runs. A request with no token, or an invalid/expired one, falls back to the policy `default_role`; elevated access needs a valid token whose role is granted (or equals the policy `admin_role`). The privileged role and public access are **policy** settings, not config flags:
+
+- **`admin_role`** (policy field, `"admin"` by default, exact case-sensitive match): the role granted full access and the `/v1/admin/*` gate. There is no separate `service` role.
+- **`default_role`** (policy field): set it to open public (no-token) access — roleless requests are evaluated as that role; remove it to close public access. It may **not** equal `admin_role`. `/v1/admin/*` and the schema/DLQ endpoints are admin-only, and a pipe with no `allowed_roles` authorizes nobody but the admin role.
```
EricAndrechek: See previous – is this what we wanted?

internal/policy/roles.go
```diff
@@ -0,0 +1,46 @@
...
+
+// AdminRole returns the role string that grants full administrative access.
+// Configurable per-policy via admin_role (case-sensitive, exact match);
+// defaults to "admin" when unset or when the policy is nil.
+func AdminRole(p *Policy) string {
+	if p != nil && p.AdminRole != "" {
+		return p.AdminRole
+	}
+	return defaultAdminRole
+}
```
EricAndrechek: I don't know if we should do this. I understand the idea, but in the scenario we talked about, where a policy is deleted EVERYONE is locked out, and the only intervention path is SSH access to the server to correct policies and reboot and whatnot, this would instead fallback to meaning the "admin" role is allowed again, which may be dangerous on their deployment. My vote is to break the deployment (for now, until we work out API type keys as a devops type fallback/method for configuring and policy etc if/when JWT or policies are broken) when policies don't exist like this. I do agree a bootstrapped "admin" role is useful for setup in the meantime though, but I think that may be better set as a boot-time config type thing when policies are being loaded in for the first time – but the implementation of this is still tricky/risky, like if Policy is deleted and server reboots, will it re-setup with "admin" fallback role? That seems dangerous... What do you think, what should we do here...?

internal/auth/auth.go
```diff
@@ -0,0 +1,138 @@
...
+	if roleClaim == "" {
+		roleClaim = "role"
+	}
+
+	var jwks keyfunc.Keyfunc
+	if cfg.JWKSURL != "" {
+		var err error
+		jwks, err = keyfunc.NewDefault([]string{cfg.JWKSURL})
+		if err != nil {
+			slog.Error("failed to initialize JWKS keyfunc", "url", cfg.JWKSURL, "error", err)
```
EricAndrechek: Should we likely return this up as an init failure or something? If they've clearly set a JWKSURL (that was not "") then they likely intend to use it, so a successful boot without it setup is probably bad and should be a loud startup failure? Alternatively, if the JWKS server is briefly offline, they may still want this to start up to serve unauthed requests at least, etc. Maybe we should be setting up `keyfunc`'s settings for storage, refreshing, etc, then if booting and the JWKSURL can't resolve and doesn't exist as cached we hard fail, but if we have a cached version of it from a prior boot/run, we just warn/log it and use that in the interim? Thoughts? Regardless, I'd think this should either fail boot loudly OR have a fallback method as described above. Also, probably shouldn't use `slog` like this, we need to give it our logger settings etc and context stuff, no? Or it won't go to JSON/stdout with proper tracing stuff, etc?

internal/auth/auth.go
```diff
@@ -0,0 +1,138 @@
...
+		return strings.TrimPrefix(h, "Bearer ")
+	}
+	if q := r.URL.Query().Get("token"); q != "" {
+		params := r.URL.Query()
+		params.Del("token")
+		r.URL.RawQuery = params.Encode()
+		return q
+	}
+	return ""
+}
```
EricAndrechek, line 152: Really like this, much cleaner than old version – nice work

internal/auth/auth.go
```diff
@@ -0,0 +1,138 @@
...
+	keyFunc := func(t *jwt.Token) (any, error) {
+		// Try JWKS first if configured, then fall back to the HMAC shared secret.
+		if jwks != nil {
+			return jwks.Keyfunc(t)
+		}
+		if _, ok := t.Method.(*jwt.SigningMethodHMAC); !ok {
+			return nil, jwt.ErrSignatureInvalid
+		}
+		return []byte(cfg.JWTSecret), nil
+	}
```
EricAndrechek, line 80: In prior spots in docs, etc, it was said we accept JWKS and/or JWTSecret – but in how this is implemented with the one keyFunc then used int the http handler next, it looks like it is strictly an OR? IMO OR instead of AND is fine, but it needs to be clear this is the case, and clear which is used in the event both are set or make it explicitly defined or something, idk.

internal/auth/auth.go
```diff
@@ -0,0 +1,138 @@
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			tokenStr := bearerToken(r)
+			if tokenStr == "" {
+				// No token: roleless request, resolved to default_role downstream.
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			token, err := jwt.Parse(tokenStr, keyFunc)
```
EricAndrechek: the `jwt.Parse` function has docstring: "The caller is strongly encouraged to set the WithValidMethods option to validate the 'alg' claim in the token matches the expected algorithm. For more details about the importance of validating the 'alg' claim, see https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/" – want to look at what that looks like and how big of a scope it is? I can't imagine it's too in-depth and is an easy fix for vuln coverage

internal/auth/auth.go
```diff
@@ -0,0 +1,138 @@
...
+			claims, ok := token.Claims.(jwt.MapClaims)
+			if !ok {
+				r = r.WithContext(WithAuthError(r.Context(), errInvalidToken))
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			ctx := WithClaims(r.Context(), claims)
+			ctx = WithRole(ctx, extractClaim(claims, roleClaim))
+			next.ServeHTTP(w, r.WithContext(ctx))
```
EricAndrechek: Really nit-picky detail/request, but I'd feel a lot better gating **these** explicitly in an if statement and having the empty/error case fallback at the end. Right now, if we edit the above code in this function or add a new gate and forget to add `return`, or are using `!ok` or `!token.Valid` incorrectly for some edge case, etc, then the request accidentally gets the claims and role it requested which is a massive security issue. Basically I'd love the order to just be reversed so the default we fall back to is the `r = r.WithContext(WithAuthError(r.Context(), whateverErrFromAbove)); next.ServeHTTP(w, r); return` instead of validating the auth check. Does that make sense?

internal/auth/auth.go
```diff
@@ -0,0 +1,138 @@
...
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			claims, ok := token.Claims.(jwt.MapClaims)
+			if !ok {
+				r = r.WithContext(WithAuthError(r.Context(), errInvalidToken))
+				next.ServeHTTP(w, r)
+				return
+			}
```
EricAndrechek: Is there a case where a role succeeds but the claims do not? Should that be allowed through? This may be more of a authn/authz question, idk, just genuinely curious if we should consider or think about that and need to cover that case

internal/auth/auth.go
```diff
@@ -0,0 +1,138 @@
...
+		if !ok {
+			return ""
+		}
+		current = m[part]
+	}
+	if s, ok := current.(string); ok {
+		return s
+	}
+	return ""
+}
```
EricAndrechek: This function is really confusing to read through, which is probably something that genuinely should be addressed for a security-critical spot like this. At some level optimization is important to get it fast since its in the hot ingest path, but it also needs to be easily auditable – I think even just comments would improve it enough

internal/auth/auth_test.go
```diff
@@ -0,0 +1,197 @@
+package auth
```
EricAndrechek, line 1: Would love to see some more tests on this to bump up coverage, specifically/especially on JWKS and assertion of the claim failure responses

internal/auth/auth_test.go
```diff
@@ -0,0 +1,197 @@
...
+
+func TestMiddleware_QueryParamToken_StrippedFromURL(t *testing.T) {
+	t.Parallel()
+	tok := testutil.MakeJWT(t, map[string]any{"role": "viewer"})
+	c := run(t, cfg(), func(r *http.Request) { r.URL.RawQuery = "token=" + tok })
+	assert.Equal(t, "viewer", c.role)
+	assert.Empty(t, c.tokenInURL, "the ?token param must be stripped so it can't leak into logs")
+}
+
+func TestMiddleware_HeaderTakesPrecedenceOverQuery(t *testing.T) {
```
EricAndrechek, line 144: Genuinely didn't think about this case, so good call, but on that note it should probably be documented somewhere that header > query

internal/policy/policy.go
```diff
@@ -8,8 +8,16 @@ import (
 
 // Policy is the top-level access control configuration.
 type Policy struct {
-	DefaultRole string                 `json:"default_role" yaml:"default_role"`
-	Tables      map[string]TablePolicy `json:"tables" yaml:"tables"`
+	// DefaultRole is the role an empty/absent role resolves to (a tokenless or
+	// roleless request). Empty means no public access. It may not equal
+	// AdminRole — refused in ResolveRole and Validate — so a roleless request
+	// can never inherit admin.
```
EricAndrechek: Same here, I think we wanted this – I know you know that by now, just trying to be helpful and document where I see it so you know where to fix

internal/policy/policy.go
```diff
@@ -8,8 +8,16 @@ import (
...
+	DefaultRole string `json:"default_role" yaml:"default_role"`
+	// AdminRole is the role granted full access and the allowlist bypass.
+	// Configurable per org (case-sensitive, exact match); defaults to "admin"
+	// when unset (see AdminRole()).
```
EricAndrechek, line 19: Unsure on defaulting, commented elsewhere but we should probably discuss this – maybe next standup?

internal/policy/policy.go
```diff
@@ -56,19 +64,44 @@ type ResolvedPermissions struct {
...
+// public access is configured — is evaluated as the configured default.
+// Matching is exact and case-sensitive (no normalization): roles are opaque
+// strings, mirroring the admin check in IsAdmin. A default_role that is empty
+// or equal to the admin role is refused here (returns "") so the admin bypass
+// can never be reached via the default; Validate also rejects saving such a
+// policy. A non-empty role is returned unchanged — roles do not inherit the
+// default's permissions.
+func ResolveRole(p *Policy, role string) string {
+	if role != "" || p == nil {
+		return role
```
EricAndrechek, line 81: Trying to think through this, there's not a case where having no policy but still allowing the role to go through could be bad/dangerous, right? On the flip side, is there a point of having the role passthrough here if there are no policies set to make use of it?

internal/policy/policy.go
```diff
@@ -56,19 +64,44 @@ type ResolvedPermissions struct {
...
+	}
+	if p.DefaultRole == "" || p.DefaultRole == AdminRole(p) {
+		return "" // fail closed: no default, or a default that resolves to admin
```
EricAndrechek: This is the case I think we wanted to change (and in the function comment too)

internal/policy/policy.go
```diff
@@ -56,19 +64,44 @@ type ResolvedPermissions struct {
...
+	}
+	return p.DefaultRole
```
EricAndrechek, line 83: Comment to note this is will need changed based on discussion in standup

internal/policy/policy.go
```diff
@@ -82,22 +115,21 @@ func Evaluate(p *Policy, role, table, operation string, claims map[string]any) *
 	}
```
EricAndrechek, line 127: What is the `default` case here, and should it, with these new changes, allow the admin role through?

internal/policy/policy.go
```diff
@@ -56,19 +64,44 @@ type ResolvedPermissions struct {
...
+	return p.DefaultRole
+}
+
 // Evaluate resolves a policy for a given role, table, and operation against JWT claims.
 func Evaluate(p *Policy, role, table, operation string, claims map[string]any) *ResolvedPermissions {
+	// Map an empty/absent role to the configured default_role (no-op if none,
+	// or if the policy is nil). A non-empty role is unchanged — roles never
+	// inherit the default's permissions.
+	role = ResolveRole(p, role)
+
```
EricAndrechek, line 92: There are a LOT of `return &ResolvedPermissions{Allowed: IsAdmin(p, role)}` type lines below – wondering if we can refactor this to short circuit admin case and catch it from the jump? Basically first thing is to detect if admin and return right away if so, and only continue if not admin? Then the returns are simpler right?

internal/policy/policy.go
```diff
@@ -258,6 +290,12 @@ func Validate(p *Policy) error {
 	if p == nil {
 		return fmt.Errorf("policy is nil")
 	}
+	// A default that resolves to the admin role would grant every roleless
+	// (tokenless, or invalid/expired-token) request full admin access. Exact,
+	// case-sensitive match — roles are opaque strings (see IsAdmin/ResolveRole).
+	if p.DefaultRole != "" && p.DefaultRole == AdminRole(p) {
+		return fmt.Errorf("default_role %q must not equal the admin role: it would grant every roleless request full admin access", p.DefaultRole)
```
EricAndrechek: fix

AGENTS.md
```diff
@@ -19,7 +19,7 @@ Eleven internal packages under `internal/` (plus `internal/testutil/` for shared
...
+- **`ingest/`** — Bento-based ingest pipeline (`bento.go`: JetStream input → per-table batch INSERT with DLQ output). The pipeline is **insert-only**. The wire format `EventMessage` (`types.go`) carries `{table_name, scope, received_timestamp, data}` and nothing else; the worker validates the table name and the payload's presence, then bulk-INSERTs. In the embedded-NATS deployment (the default), the server runs with `DontListen: true` (`internal/mq/embedded.go`), so the only Publishers reachable on the `ingest.>` subjects are in-process Go code — today, only the HTTP `/v1/ingest/{table}` handler. Non-insert mutations (`DELETE`/`UPDATE`/`TRUNCATE`/…) must go through `POST /v1/admin/query` under the admin role (the same `RequireAdmin` gate as the rest of `/v1/admin/*`), so non-admin callers never reach the proxy. A request with no token (or an invalid one) resolves to the `default_role`, which is never the admin role, so it can't reach this endpoint. Plus `Sweeper` (Active Sweeper for NATS message lifecycle) + `EventMessage`/`BufferConsumerName` types (`types.go`)
```
EricAndrechek: `/v1/ingest?table={table}`

EricAndrechek: And default role can be admin role fix

AGENTS.md
```diff
@@ -34,13 +34,13 @@ Eleven internal packages under `internal/` (plus `internal/testutil/` for shared
...
...ow-level permissions with JWT claim templating (`{{ jwt.path }}`). Policies stored in NATS KV with file-based bootstrap and cluster-wide sync via KV Watch. The **single** admin check is `policy.IsAdmin` (role == `admin_role`, configurable per policy, `"admin"` by default, **exact case-sensitive** match — no `ToLower`/`TrimSpace` normalization); it is the one source of truth shared by `Evaluate`, `ResolveRole`, `Validate`, the `/v1/admin` gate, and pipe authorization (`policy.RoleAllowed`). An empty/absent role is fail-closed: it never matches a role key — matching is exact (no `"*"` any-role wildcard) and only the admin role bypasses the allowlist — `Validate` rejects empty role keys at write time, and a `nil` policy denies every non-admin. Preserve this when touching `internal/policy` (the policy twin of the pipe invariant in #13, see #159). The optional `default_role` is the one sanctioned exception: `ResolveRole` maps an empty role to it *before* evaluation (so a roleless request gets that role's perms), but never to the `admin_role` — refused in both `ResolveRole` (returns `""`, fail-closed) and `Validate` (rejects the write). Roles do **not** inherit the default's permissions.
```
EricAndrechek: Given our strict string comparison policy, I wonder, if we should add the ability for various roles to map to admin...? Doesn't need to be in this PR if changes scope too much, just something I think we should document/think about

SECURITY.md
```diff
@@ -24,10 +24,10 @@ We will acknowledge receipt within 48 hours and aim to provide an initial assess
...
+- **Query passthrough**: Raw SQL via `POST /v1/admin/query` is restricted to the admin role — the same `RequireAdmin` gate as the rest of `/v1/admin/*`. A request with no/invalid token resolves to the `default_role` (never the admin role), so it cannot reach this endpoint. Raw SQL has no per-statement scope check (a full SQL parser would be needed to authorize predicates), so the role gate is the entire authorization story. Non-admin callers use structured queries (`POST /v1/tables/{table}/query`, validated against schema with permission injection) or named pipes (`GET/POST /v1/pipes/{name}`); raw-SQL grants to non-admin roles via the policy engine are no longer supported (the `raw_sql` field on policies has been removed).
```
EricAndrechek: default can be admin

EricAndrechek: And need to fix `/v1/query?table={table}` endpoint (from old `/v1/tables/{table}/query`)

EricAndrechek (changes requested):
Big PR but largely good, much better than last check almost there for sure. Some highlights and some things not mentioned in the comments:
- I think we agreed we wanted to allow the default role to be admin, especially for dev/local testing, etc – I commented all the places I saw this either documented/commented or code incorrect to tag for fixes
- I want to talk about how the process of having no policy set will work, especially after recovering from a previous policy or something. I think a lot of the issues/concerns around this would/could be fixed with an admin api type thing separate from JWT and whatnot, but still worth discussing
- didn't check and wasn't sure, but I want to confirm that the way we do ctx injection for auth stuff still works with spans and traces for otel and logging?
- also, I think some of my comments may introduce PR scope creep in asking for changes so please push back on them as needed, ESPECIALLY for the things we explicitly document as being future TODOs in my [comment](https://github.com/Wave-RF/WaveHouse/pull/172#issuecomment-4523275591)

docs/src/content/docs/api.md
taitelee: Yes

docs/src/content/docs/api.md
taitelee: Yes to first. The pipes are already allowed by the default_role though so technically they are open to the public I thought that was what we decided with?

docs/src/content/docs/architecture.md
taitelee: Yes. Fixing.

docs/src/content/docs/architecture.md
taitelee: Yes. Fixing.

docs/src/content/docs/configuration.md
taitelee: Roleless requests are evaluated to default role, yes. I am fixing that default role can be the same as admin role.

internal/policy/roles.go
taitelee: - Ok, so `AdminRole(nil)` used to hand back `"admin"`. I've changed it to return `""`, so even someone who forgets to gate through `IsAdmin` can't pull admin out of a missing policy.
- On reboot there's no hardcoded fallback either. Bootstrap only fires when the KV is empty, and it seeds the policy file as-is (whatever `admin_role` that file declares). So admin only comes back when a real policy is reloaded, so a missing policy never silently re-grants admin at runtime.
- One thing to note. Because bootstrap reseeds from the file when KV is empty, deleting the policy via the API and then rebooting (with the file still on disk) restores it without a manual edit. It still needs server access and only ever restores your own config, so I left it as is. The file is the source of truth.

On breaking the deployment what do you mean? Like do we forcefully do something besides the lockout we have now? Also, I realized that even without policies pipes can still work for authenticated, role matching requests even with no policy. Just to confirm we are okay with no policy but still can service authenticated requests right?

taitelee: Sorry if my response is confusing when I reference certain changes because I haven't actually pushed anything to the PR yet.

taitelee: Oh I see this is up for discussion. Leaving this for standup.

internal/auth/auth.go
taitelee: Yea. If the JWKS URL is just unreachable (not malformed), `keyfunc.NewDefault` doesn't even return an error. It tolerates the first failed fetch by default, so we'd boot with a non-nil-but-empty keyset, no log at all, and every token silently fails verification Agreed that shouldn't be the default.

So I went with fail loud at boot. Changed Middleware to force a synchronous initial fetch (`NewDefaultOverrideCtx` with `NoErrorReturnFirstHTTPReq=false`) so an unreachable/misconfigured JWKS surfaces an error, and `main.go` treats it as a fatal startup error (return 1). Skipped the persistent/stale-cache fallback. `keyfunc` already auto-refreshes once running, so a JWKS server dropping after boot is handled in memory. The only gap is "down at boot," and there I'd rather hard fail than serve on stale keys.

On the slog bit, you're right that it does reach JSON/stdout (and OTLP when enabled). `main.go` calls `slog.SetDefault`, so the package level `slog.Error` wasn't bypassing our config, and it's a boot-time log so there's no request span anyway. But the fail-loud change drops that `slog.Error` entirely. The error just propagates up and gets logged by the normal startup path.

Need to look at this with a clearer head tomorrow.

docs/src/content/docs/api.md
EricAndrechek: Yeah it is this just made it sound like pipes specifically are **never** open to the public

taitelee: Ah right yes changing.

AGENTS.md
taitelee: So having a configurable list of roles that are all admin roles?

internal/policy/policy.go
taitelee: The default case here I just treated like an unknown operation. Callers would only pass select or inset (structured query and ingest paths) so the default I think is unreachable, therefore I would just deny everyone including admins on some operation the engine doesn't understand. I don't know if it's worth having admin even be allowed in this case, and not sure when that'd be useful honestly. What do you think? I think keeping as is is harmless.

AGENTS.md
EricAndrechek: Yeah? Like if the default is `admin_role: "admin"` that's fine, and I think a single role/string should still work, but I'm imagining if we are (currently) _only_ allowing **one** role to be the "admin" role, but wanted to apply it to multiple people without having to change JWT stuff at the auth provider level, etc, it'd be nice for them to be able to set `admin_role: ["admin", "devops", "Administrator"]` or something like that – but I think we could also make this into a new issue to track if u want to keep this ones scope tighter...

SECURITY.md
taitelee: Yep.

AGENTS.md
taitelee: Yep.

internal/policy/policy.go
taitelee: Fixed.

AGENTS.md
taitelee: I see. I think I'm going to leave this out for a later addition, but until I push my changes and have them review I will leave unresolved and make note.

internal/policy/policy.go
taitelee: I agree with this. Making fix.

internal/auth/auth.go
taitelee: Ok so for this I made a fix for `WithValidMethods` where I hand it an explicit allowList and it checks `alg` before touching any key. We key the list off the active verifier (using `validMethods` list). JWKS deployment only (RS/ES/PS/EdDSA), or HMAC deployment (HS only). So `none` would be rejected, and we have logic so that an HS256 token against an RSA deployment or vice versa would be rejected (as this is an issue the article in the link brings up). I believe this is a sufficient fix but you can reply if you think otherwise.

internal/auth/auth.go
taitelee: Yea so before if the JWKS URL was misconfigured or unreachable, `jwks` stayed empty and every token then failed to verify, and we'd quietly run as a pure public deployment. So now we have a synchronous first fetch and returns an error. After a successful boot, `keyfunc` keeps the keys refreshed in process, so a JWKS outage later tolerated. The `slog.Error` I removed. The error now just propagates to `main.go` and is logged by our logger.

internal/auth/auth.go
taitelee: Check new comment in this `keyFunc` section. It is strictly OR yes.

internal/auth/auth.go
taitelee: There should not be a case where a role succeeds but the claims do not because we extract the role from the claims using `extractClaim(claims, roleClaim)`. In my new changes I have both `token.Claims.(jwt.MapClaims)` and `extractClaim(claims, roleClaim)` in the same block after we parse the token. There wouldn't therefore be a case where we have roles separated from the claims.

taitelee: I don't know if roles but no claims in something we want though I'd think not?

internal/auth/auth.go
taitelee: Adding comments, looking into more and will follow up. Basically we're looping to find the role within `jwt.MapClaims` that could have nested maps based on I guess how some identity provider could have nested in the role within the payload. Given the path, we just make sure that the role exists within the amps and that it is a string.

internal/auth/auth.go
taitelee: TODO: add to docs

internal/auth/auth.go
taitelee: I added comments. I am confident that there should be no behavior changes but I am open to feedback.

internal/auth/auth_test.go
taitelee: Once you review everything I can add more test cases ;-; but I'll wait till then due to time

internal/policy/policy.go
taitelee: Outdated

internal/auth/auth.go
taitelee: Reversed it so the only path that sets claims + role is the explicit `if err == nil && token.Valid { if claims, ok := token.Claims.(jwt.MapClaims); ok { … return } }` block. Everything else like parse error, invalid/expired, non-MapClaims, and anything we tack on later now falls through to the fail-safe default at the bottom (roleless + recorded auth error). So the fall-through is the deny now, not the grant: a dropped return or a mis-set !ok/!Valid can only ever over-restrict, never promote an unverified token to a real role.

internal/auth/auth.go
```diff
@@ -61,41 +79,56 @@ func Middleware(cfg Config) func(http.Handler) http.Handler {
...
+			// means a future missing return, or a mis-set !ok/!Valid condition,
+			// can never accidentally promote an unverified token to a real role.
+			token, err := jwt.Parse(tokenStr, keyFunc, jwt.WithValidMethods(validMethods))
+			if err == nil && token.Valid {
+				if claims, ok := token.Claims.(jwt.MapClaims); ok {
+					ctx := WithClaims(r.Context(), claims)
+					ctx = WithRole(ctx, extractClaim(claims, roleClaim))
+					next.ServeHTTP(w, r.WithContext(ctx))
+					return
+				}
```
EricAndrechek, line 116: Is there a case or anything for us to return an error or whatever if Parse fails/is not ok?

internal/auth/auth_test.go
EricAndrechek: Didn't see this documented anywhere – bump @taitelee

internal/auth/auth_test.go
```diff
@@ -154,6 +156,46 @@ func TestMiddleware_InvalidQueryParamToken_FallsBackWithError(t *testing.T) {
...
+	t.Parallel()
+	// A reachable JWKS endpoint (even with an empty key set) constructs cleanly.
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		_, _ = w.Write([]byte(`{"keys":[]}`))
+	}))
+	defer srv.Close()
+
+	mw, err := Middleware(Config{JWKSURL: srv.URL})
+	require.NoError(t, err, "a reachable JWKS endpoint must construct successfully")
+	require.NotNil(t, mw)
```
EricAndrechek, line 199: What happens if the set is empty though – like this does indeed "construct successfully"?

internal/policy/policy.go
```diff
@@ -68,17 +69,19 @@ var claimTemplateRe = regexp.MustCompile(`\{\{\s*jwt\.([a-zA-Z0-9_.]+)\s*\}\}`)
...
 func ResolveRole(p *Policy, role string) string {
 	if role != "" || p == nil {
 		return role
 	}
-	if p.DefaultRole == "" || p.DefaultRole == AdminRole(p) {
-		return "" // fail closed: no default, or a default that resolves to admin
+	if p.DefaultRole == "" {
+		return "" // fail closed: no public default_role configured
 	}
 	return p.DefaultRole
```
EricAndrechek: If the `p.DefaultRole == ""` and we return `""`, isn't that the exact same as returning `p.DefaultRoles`...?

internal/policy/policy.go
EricAndrechek: Good fix nice work

internal/auth/auth.go
EricAndrechek: Still not super clear to me, would love inline comments on it, but not going to block approval over this

EricAndrechek (commented):
Largely a few doc updates sort of things left but basically there – do want to talk about the ones left open though

internal/auth/auth_test.go
taitelee: TODO: Api and SDK doc additions

internal/policy/policy.go
taitelee: Great catch -- we can get rid of this lol.

internal/auth/auth_test.go
taitelee: Added.

internal/auth/auth.go
taitelee: Ok so this middleware is authentication only, it never writes a response or rejects. It just establishes a role if it can and hands off. A failed/invalid/expired parse (and the `!ok` claims case) all fall through to the same fail-safe at the bottom, which records the reason with `WithAuthError` instead of returning it.

The reason we capture instead of throw is that the right status code isn't knowable here. It depends on whether the route actually needs a role, which only the downstream gate knows. `writeAuthzDenied` in `internal/api/errors.go` reads the recorded error. A denial with a present but bad token comes back as 401 (invalid/expired token), otherwise it's a 403. So a bad token on a public route still succeeds, and a bad token on `/v1/admin/` fails loud as a 401 instead of a confusing 403. Returning early here would break that and re-couple authn with authz.

On the `ok` bool specifically, `jwt.Parse` always returns `jwt.MapClaims` since we don't pass a custom claims type, so `!ok` is I think unreachable. If it ever did hit, falling through to the fail-safe is still correct (roleless + invalid-token, never promoted to a role). Because of this I left the behavior as-is and added a AI comment in `auth.go` walking through this.

internal/auth/auth_test.go
taitelee: Two different things here. At boot, which is what this test checks, `{"keys":[]}` is a valid JWK Set, the fetch succeeded, there just aren't any keys, so construction succeeds. Only an unreachable/erroring endpoint fails boot (`TestMiddleware_JWKSUnreachableAtBoot_FailsLoud`). At request time, with zero keys there's nothing to verify a signature against, so every token fails and falls back to the roleless `default_role` which is fail-closed, an empty set means trust nobody, never trust everybody.

I added `TestMiddleware_JWKSEmptyKeySet_TokenDoesNotAuthenticate` (it should be a valid test case I kind of rushed it). It signs a real EdDSA token on purpose. EdDSA passes the JWKS alg allowlist so it actually reaches the empty key set and fails there; an HMAC token would've been rejected by the allowlist first and tested the wrong thing. Asserts empty role + `errInvalidToken`.

One thing I want your input on. An empty but reachable set is basically the same "no token can validate" state we already fail fast on for an unreachable endpoint, so we could fail boot on an empty set too. I left it fail-closed at request time since a set can be legitimately empty mid rotation, but would be open to make boot fail loud on an empty set if you think that's better.

internal/auth/auth_test.go
```diff
@@ -0,0 +1,265 @@
...
+		"exp":  jwt.NewNumericDate(time.Now().Add(time.Hour)),
+	})
+	signed, err := tok.SignedString(priv)
+	require.NoError(t, err)
+
+	c := run(t, Config{JWKSURL: srv.URL, RoleClaim: "role"}, bearer(signed))
+	assert.Empty(t, c.role, "empty JWKS → no key validates the token → roleless default")
+	assert.False(t, c.hasClaims)
+	assert.True(t, errors.Is(c.authErr, errInvalidToken), "present-but-unverifiable token records invalid-token")
+}
```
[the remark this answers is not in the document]
taitelee: I looked at this but I'm going to keep it standalone. The guideline (`AGENTS.md:355`) frames table driven tests as the pattern for multiple same shaped scenarios, and this is a single one (empty key set → token doesn't authenticate). It also doesn't fold cleanly into the adjacent JWKS cases: `JWKSUnreachableAtBoot` and `JWKSReachableAtBoot` assert on `Middleware()`'s return value with no request, whereas this test constructs the middleware, signs a fresh EdDSA token, and drives a full request through `run()`, so a shared table would need per-row setup closures and branching assertions, which reads worse than three named functions. It's also consistent with the rest of `auth_test.go`, which keeps standalone functions for middleware scenarios and reserves the table-driven form for `TestExtractClaim`'s pure input/output enumeration.

internal/auth/auth_test.go
EricAndrechek: I think failed closed is fine if the address was reachable and resolving a proper JWKS, even if empty, as (as I understand the keyFunc library we are using) our keyFunc stuff will automatically be doing its own goroutine to periodically refetch the key from there and whatnot, so them having it pointing somewhere where they don't yet have auth details is valid knowing that they can change what the endpoint is serving and our jwks will refetch with new valid credentials without requiring a server restart – right?

taitelee: Yes. `keyfunc` runs a background goroutine that refetches the JWKS on a timer and on demand when a token shows up with a key ID we don't have (rate limited to once every 5min), so an endpoint that's empty now and gets real keys later is picked up without a restart.

EricAndrechek (approved):

PR #181, opened 2026-05-24
fix(flake): local cache tests wait for admission, not sleep

EricAndrechek:
## Summary

Fixed potentional or real flake possibility in local cache's tests, that used sleep instead of waits for admission. Bug found by @taitelee


taitelee (approved):
Nice. So the reason these changes needed to be made was because ristretto could have possibly not processed an item on `setBuf` and wrote it into our map, so `Get` wouldn't have succeeded by the time we got to checking if our cached item was actually processed.

PR #189, opened 2026-05-27
feat(api): structured WARN logging on authorization denials

### LINKED ISSUE #145: feat(auth): RequireRoles middleware — fail closed, no permissive fallback

taitelee:
## Summary

Every authorization denial now flows through one shared helper that emits a structured `slog` WARN, so a misconfigured role or policy is visible in logs without reproducing the request.

- **`writeAuthzDenied`** (`internal/api/errors.go`) is the single denial path for all gates. It classifies the failure — a present-but-invalid/expired token fails loud as **401** with the sanitized token reason, everything else is a
  **403** — and derives a greppable `reason` (`role not in allowed roles`, `no role and no default_role configured`, or the token error).
- **`logAuthzDenied`** emits the `"authorization denied"` WARN with `reason`, `role_observed` (pre-default-resolution) vs `role_resolved`, `roles_allowed`, the chi `route` **pattern** (low-cardinality, no concrete path params), `method`, and `status`. slog escapes control chars, so the request-derived fields carry no log-injection risk.
- **`forbiddenForRole`** returns the 403 body — verbose for the anonymous / no-`default_role` case, terse otherwise (doesn't enumerate roles).
- All existing gates route denials through it: the `/v1/admin/*` gate (`router.go`), the pipe gate (`pipes.go`), and the policy-evaluator paths (`ingest.go`, `structured_query.go`). `roles_allowed` is populated only where a flat allowlist exists (a pipe's `allowed_roles`); the admin gate and evaluator paths pass `nil`.

## Test plan

- [x] `make ci` (and `go test ./internal/api/`) passes locally
- [x] Admin-gate denial logs a WARN: reason `role not in allowed roles`, the role, route `/v1/admin/query`, status 403
- [x] Tokenless request under a `default_role` logs `role_observed=""` and `role_resolved=<default>`
- [x] Present-but-invalid token fails loud: status 401, `reason` carries the token error
- [x] Pipe denial logs `roles_allowed` equal to the pipe's `allowed_roles`
- [x] Denial routed through the real mux logs the chi route *pattern*, not the raw path

## Related Issues

Part of #145: implements the "structured WARN on every denial" requirement from its Proposed Solution. Does not fully close #145; the broader role-gating enforcement is tracked there.


internal/api/errors.go
```diff
@@ -22,23 +24,77 @@ func writeJSONError(w http.ResponseWriter, status int, message string) {
...
+	}
+	slog.LogAttrs(r.Context(), slog.LevelWarn, "authorization denied",
+		slog.String("reason", reason),
+		slog.String("role_observed", auth.RoleFromContext(r.Context())),
+		slog.String("role_resolved", resolvedRole),
+		slog.Any("roles_allowed", allowedRoles),
+		slog.String("route", route),
+		slog.String("method", r.Method),
+		slog.Int("status", status),
+	)
```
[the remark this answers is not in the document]
taitelee: This is just wrong. `slog.LogAttrs` is a real packet level function in `log/slog`. It calls `LogAttrs` on the default logger.

internal/api/errors.go
```diff
@@ -22,23 +24,77 @@ func writeJSONError(w http.ResponseWriter, status int, message string) {
...
+		}
+	}
+	slog.LogAttrs(r.Context(), slog.LevelWarn, "authorization denied",
+		slog.String("reason", reason),
+		slog.String("role_observed", auth.RoleFromContext(r.Context())),
+		slog.String("role_resolved", resolvedRole),
+		slog.Any("roles_allowed", allowedRoles),
+		slog.String("route", route),
+		slog.String("method", r.Method),
+		slog.Int("status", status),
```
EricAndrechek: Just want to confirm – since we aren't passing this a logger from main or using the logger in router or whatever, is that ok – does it (slog) still use the defaults and stuff we set in main (like json vs text, where to log to, etc)?

Also, does this still include the trace and span from the request in the log?

internal/api/errors_test.go
```diff
@@ -33,3 +42,161 @@ func TestWriteJSONError_EscapesSpecialCharacters(t *testing.T) {
...
+// and Go pauses parallel tests until the non-parallel ones finish, so a
+// non-parallel test owns the default logger for the duration of fn.
+func captureWarnLogs(t *testing.T, fn func()) []map[string]any {
+	t.Helper()
+	var buf bytes.Buffer
+	prev := slog.Default()
+	slog.SetDefault(slog.New(slog.NewJSONHandler(&buf, &slog.HandlerOptions{Level: slog.LevelWarn})))
+	defer slog.SetDefault(prev)
+
+	fn()
```
EricAndrechek: This feels **incredibly** janky... Look at how [`internal/policy/store_test.go#L40-L43`](https://github.com/Wave-RF/WaveHouse/blob/d50cd9a1ad7cf4818eb083b4966b8488e9d3d521/internal/policy/store_test.go#L40-L43) does it... combined with my comment on `errors.go`, would this not work instead and be a bit cleaner?

internal/api/errors_test.go
```diff
@@ -33,3 +42,161 @@ func TestWriteJSONError_EscapesSpecialCharacters(t *testing.T) {
...
+	t.Helper()
+	for _, m := range records {
+		if m["msg"] == "authorization denied" {
+			assert.Equal(t, "WARN", m["level"], "denial must be logged at WARN")
+			return m
+		}
+	}
+	t.Fatal("expected an 'authorization denied' WARN log, got none")
+	return nil
+}
```
EricAndrechek: Again same thing here, look at [store_test.go](https://github.com/Wave-RF/WaveHouse/blob/d50cd9a1ad7cf4818eb083b4966b8488e9d3d521/internal/policy/store_test.go#L46-L47) – it accomplishes the same thing more or less, but is a lot simpler/easier

internal/api/ingest.go
```diff
@@ -77,8 +77,7 @@ func (h *IngestHandler) Handle(w http.ResponseWriter, r *http.Request) {
 		claims, _ := auth.ClaimsFromContext(ctx)
 		perms = policy.Evaluate(p, role, table, "insert", claims)
 		if !perms.Allowed {
-			slog.WarnContext(ctx, "policy enforcement rejected request", "role", role, "table", table)
-			writeAuthzDenied(w, r, role)
+			writeAuthzDenied(w, r, role, nil)
```
EricAndrechek: I'm noticing here we don't pass context but replace the `WarnContext`, which I assume means that is not longer tagged with the trace or span or any of the data in the request/context which (as it currently was) was incredibly useful.

Also, unlike the other ones that you just add the required role to the `writeAuthzDenied` call, this one feels like it gives us less info/context now, like we don't have a way of seeing that it was perms that failed and what perms they had vs were required, etc – no?

internal/api/router.go
```diff
@@ -249,7 +249,7 @@ func RequireAdmin(store *policy.Store) func(http.Handler) http.Handler {
 				next.ServeHTTP(w, r)
 				return
 			}
-			writeAuthzDenied(w, r, role)
+			writeAuthzDenied(w, r, role, nil)
```
EricAndrechek: Same concern as in `ingest.go` here, just a bit less severe (apart from the context and trace/span concern from all of them) but it feels as though we are still missing the context of that this was an explicit admin check then or something – sometimes a /admin prefixed route would make that clear in the logs, but when its a policy decision or something it isn't really.

internal/api/structured_query.go
```diff
@@ -70,7 +70,7 @@ func (h *StructuredQueryHandler) Handle(w http.ResponseWriter, r *http.Request)
 	claims, _ := auth.ClaimsFromContext(r.Context())
 	perms := policy.Evaluate(p, role, table, "select", claims)
 	if !perms.Allowed {
-		writeAuthzDenied(w, r, role)
+		writeAuthzDenied(w, r, role, nil)
```
EricAndrechek: Same concern here

EricAndrechek (changes requested):
I like the idea behind it, and you did a good job trying to keep it DRY, but I don't think that this implementation is complete, and honestly is doing more harm than good to logging details in its current form.

internal/api/errors_test.go
taitelee: Yea. I dropped the `[]map[string]any` round trip and `findDenialLog`. There's now one `captureLogs(t, fn)` string helper that returns the buffer, and the tests assert with `assert.Contains(out, "reason":"…")` including the same buffer+Contains style as store_test.go.

I did keep the `slog.SetDefault` swap, and the reasoning ties back to your `errors.go` question. `store_test.go` can inject a logger because `policy.NewStore` is built once at boot and takes a logger as a dependency, whereas the request handlers in `internal/api` deliberately log through `slog.Default()+context` (global logger). So capturing that output in a test means swapping `slog.Default()` for the duration of the test. That's all `captureLogs` does. Giving each handler a logger to write to (the way the store has one) would work too, but it's a broader change. What do you think? Should I go with that approach?

internal/api/errors.go
taitelee: So `slog.LogAttrs` resolves to `slog.Default()` which is defaulted to stdout at boot, but then we swap in the OTLP logger from `observability.NewLogger` when `otel.logs.enabled`. So yes it still uses the defaults and what we set in main. To answer the second part, yes it also includes the trace and span from the request in the log because we pass in `r.Context()`, and the OTel logger wraps everything in `observability.TraceHandler`, which lifts `trace_id`/`span_id` off the span in the context and adds them to the record (shown in `TestTraceHandler_AddsTraceIDs`). The request span is already on `r.Context()` (otelhttp middleware + ingest's explicit span), so the denial is correlated to its trace like any other log.

internal/api/errors_test.go
taitelee: I got rid of findDenialLog in my current not staged commit so this might be outdated.

internal/api/errors_test.go
EricAndrechek: I'm not sure – I think consistency is best, so my thinking would be we should either have logger passed to all of our packages and whatnot, OR have them all just use the default logger. I didn't know the default logger worked like that and could be like a global singleton and use the same configs and pull in the context automatically without explicitly passing it and stuff, but if that's the case then I wonder if that's the cleaner option, and what the point of us passing a logger into each package/module initialization is anyway... Thoughts?

taitelee: Ok. It's going to take me some time to change so I'll keep you updated.

EricAndrechek: Which way are you changing it?

taitelee: The cleaner option you mentioned just being in line with how `store_test.go` does it. I'll just first try finding a way to pass in logger to all packages. `errors_test.go` already did use the default logger so I'm not sure what you meant by that.

internal/api/router.go
```diff
@@ -36,6 +36,9 @@ type Dependencies struct {
 	JS          jetstream.JetStream // for SSE/WS gap-fill
 	CORSOrigins []string            // allowed CORS origins; ["*"] = allow all
 	LogLevel    *slog.LevelVar
```
EricAndrechek: I thought we were/we could get rid of this now, no?

internal/api/router.go
```diff
@@ -237,7 +240,8 @@ func jsonRecoverer(next http.Handler) http.Handler {
 // resolves to an empty (non-admin) role and is denied here. Denials go through
 // writeAuthzDenied, so a present-but-invalid token fails loud (401 + token
 // reason) rather than as a bare 403.
-func RequireAdmin(store *policy.Store) func(http.Handler) http.Handler {
+func RequireAdmin(store *policy.Store, logger *slog.Logger) func(http.Handler) http.Handler {
+	logger = loggerOrDefault(logger)
```
EricAndrechek: What is this? Can we not just use from deps or whatever here too? I don't feel like this is how other packages we have do it

internal/api/errors.go
```diff
@@ -22,23 +24,88 @@ func writeJSONError(w http.ResponseWriter, status int, message string) {
...
+// pipe's allowed_roles); the gates with no flat role list — the /v1/admin gate
+// and the policy-evaluator paths (ingest, structured query) — pass nil. attrs
+// are gate-specific structured fields appended to the WARN: each gate tags a
+// "gate" (admin / policy / pipe) so a denial is attributable to the check that
+// raised it (the route pattern alone can't — /v1/schema runs the admin gate,
+// not a policy one), and the policy paths add the table + action they evaluated.
+// logger is the calling gate's injected logger (each handler holds one; main
+// wires it, tests pass their own) — the denial WARN goes there, not to a
+// package global.
+func writeAuthzDenied(w http.ResponseWriter, r *http.Request, logger *slog.Logger, role string, allowedRoles []string, attrs ...slog.Attr) {
```
EricAndrechek, line 47: Need to check and confirm trace context is being passed in/inherited here, so logs show in the proper trace/span in otel, and are not just orphaned logs with no trace.

EricAndrechek (changes requested):

internal/api/ingest.go
taitelee: Looking into it, and will continue in a separate thread below.

internal/api/router.go
taitelee: Whoops. This was brought up in an in person discussion, and removed in person, but I didn't actually push the change yet. We did get rid of this though. For documentation, basically, we noticed that LogLevel was being used by an untracked, undocumented API endpoint where it appeared that the API endpoint assumed to have the capability of changing the log level. This was added sometime in early development, by me I'm pretty sure, but forgotten about, became outdated very quickly, and managed to survive unnoticed in numerous iterations. Removed everything related to it in this PR.

EricAndrechek: It looks like it is still here as a struct dependency though, do we need it/use it still?

tests/e2e/sdk/dlq.test.ts
```diff
@@ -14,6 +14,8 @@ describe("Dead Letter Queue (DLQ) & Failures", () => {
   it("routes only the failed row to DLQ while valid rows are inserted", async () => {
     const runId = testId();
 
+    await new Promise(resolve => setTimeout(resolve, 5000));
```
EricAndrechek, line 18: Can we add a `// TODO` comment on this to remove it when #192 lands?

EricAndrechek (changes requested):

tests/e2e/sdk/dlq.test.ts
taitelee: Added.

EricAndrechek (approved):

PR #213, opened 2026-06-03
feat(api): add /healthz + /readyz probes, deprecate /health + /ready

### LINKED ISSUE #144: feat(api): align health endpoints with K8s convention (/healthz, /readyz + per-dependency sub-endpoints)
## Problem

WaveHouse exposes `/health` and `/ready` (registered in `internal/api/router.go:87-88`), which work but don't follow the K8s convention (`/healthz`, `/readyz`). Operators wiring kube `livenessProbe` / `readinessProbe` paths against a WaveHouse deployment have to remember the non-standard names.

Additionally, the current endpoints return a single overall status. When `/ready` flips to NOT_READY, there's no fast way to see which dependency is the cause (NATS? ClickHouse? Schema discovery? DLQ?). Per-dependency sub-endpoints would let on-call see the failing leg without grepping logs.

## Decide before alpha

**Compat window for `/health` and `/ready`** is the gating decision. Options:

- **Hard-cut**: rename only. `/health` and `/ready` return 404 in v0.1.0-alpha.1. Cleanest, but anyone who scripted against the old names in the pre-alpha period breaks.
- **Aliases**: register both. Half the disk for the cost of one comment in CHANGELOG noting `/health`/`/ready` are deprecated and will be removed in v0.2.0. **(Recommended.)** The alpha is the cheapest moment to introduce the K8s names; deferring locks `/health`/`/ready` in as v1.0 stable surface forever.

**Aggregate `/readyz` behavior** is the other open question. Recommend evaluating *all* subsystems and returning a unified failure picture (`{"ready": false, "checks": {"nats": "ok", "clickhouse": "fail: timeout", ...}}`) rather than short-circuiting — same latency in the happy path and far more useful in the failure path.

## Proposed Solution

1. Register `/healthz` + `/readyz` as the canonical paths going forward. Keep `/health` + `/ready` as aliases for v0.1.x; drop in v0.2.0. Document in CHANGELOG.
2. Add per-dependency status endpoints, all under `/healthz/`:
   - `/healthz/nats` — embedded JetStream + connection status
   - `/healthz/clickhouse` — DB ping + driver state
   - `/healthz/schema` — last successful refresh timestamp
   - `/healthz/dlq` — stream depth / consumer health
3. Keep all of these lightweight (no fan-out to internal stats that would themselves block).

## Sequencing

Recommend landing before v0.1.0-alpha.1 (#149) — the URL contract is harder to change post-launch than pre-launch.

## Additional Context

Distinct from #95, which addresses the schema-discovery retry/503 semantics — that issue focuses on boot-time non-fatal handling. This issue is about endpoint naming + per-dependency drill-down.

taitelee:
## Summary

Adopts the Kubernetes-convention health-probe names from #144: `GET /healthz` (liveness) and `GET /readyz` (readiness) are now the canonical paths. The existing `/health` and `/ready` are kept as **deprecated aliases** of the same handlers for the v0.1.x line and are slated for removal in v0.2.0 — so nothing breaks for anyone already probing the old paths. Response bodies are unchanged (`/healthz` → `{"status":"ok"}`, `/readyz` → `{"status":"ready"}`, same boot-gated 503s as before).

WaveHouse's own tooling is repointed at the canonical names so it doesn't ride an alias we've announced for removal: the distroless `health` subcommand, the e2e orchestrator's readiness wait, the SDK e2e setup probe, and both Dockerfile `HEALTHCHECK`s. The `prometheus.path` reserved-collision check now also covers `/healthz` + `/readyz`, and the tracing span-skip middleware excl probe paths. Docs, `AGENTS.md`, and `CHANGELOG.md` updated to match.

**Part 2 of #144 is deferred**: the per-dependency drill-down probes under `/healthz/{clickhouse,nats,schema,dlq}` and the richer aggregate `{"ready", "checks"}` readiness body are left for a follow-up. This PR is a pure rename + alias.

## Related Issues

Part of #144 (part 1: `/healthz` + `/readyz` rename + aliases). Intentionally **not** `Closes` — part 2 (per-dependency probes, aggregate readiness body) is deferred.


clients/ts/src/http.test.ts
EricAndrechek, line 1: Would prefer to keep these at `/health` but not a big concern

cmd/wavehouse/health.go
EricAndrechek, line 1: More detail in the full review, but we should determine if docker compose health checks (which I'd imagine will be the primary consumer of this route/cmd) want `/readyz`, `/livez`, or `/health`

docs/src/content/docs/development.md
```diff
@@ -111,10 +111,14 @@ curl -N "http://localhost:8080/v1/stream?table=clicks"
...
 
-# Health check (no auth required)
-curl http://localhost:8080/health
+# Liveness check (no auth required)
+curl http://localhost:8080/healthz
 # → {"status":"ok"}
 
+# Readiness check (no auth required)
+curl http://localhost:8080/readyz
+# → {"status":"ready"}
```
EricAndrechek, line 120: [Earlier in `api.md`](https://github.com/Wave-RF/WaveHouse/pull/213/changes#diff-fc2f88698611158d025a9cbc054b59d7a0cd0ddca2eff431f572bebe1d6ff4dfR58) we say that `/readyz` always attaches an `error` field that is parseable and contains things different from the `/healthz` endpoint, but this seems to suggest otherwise?

tests/integration/boot_resilience_test.go
```diff
@@ -24,14 +24,14 @@ import (
 // post-#95 contract matrix against a real ClickHouse testcontainer that we
 // stop/start mid-test:
 //
-//	| State                    | /health | /ready |
+//	| State                    | /healthz | /readyz |
 //	|--------------------------|---------|--------|
 //	| Boot, CH down            | 503     | 503    |
 //	| Boot, CH up after retry  | 200     | 200    |
 //	| Post-boot, CH dies       | 200 ★   | 503    |
 //	| Post-boot, CH back       | 200     | 200    |
```
EricAndrechek: I think this is useful to show in the docs instead of hidden away in a unit test

EricAndrechek (changes requested):
A quick google make it appear as though k8s has deprecated or started phasing out `/healthz` and recommending `/readyz` and `/livez` instead – considering we are starting from scratch here we should probably conform with the current recommendations. I also don't know what the norms are for what each endpoint must return, or what cases they return success/failures in, so additional implementation work may be needed to get it all conforming properly.

Then we also need to work out whether we need `/healthz` for backwards compatability with older k8s versions or not, and how (if at all) it will differ.

We also need to think about which if any of these endpoints is what we'd want to use for the docker compose health checks, and which our own e2e orchestrator uses, and what if anything should be returned along with those routes.

I also am thinking/wondering, but am unsure on the hosting and deployment conventions, if `/readyz`, `/livez`, `/healthz` etc. are usually on the same port/exposed to traffic like this or not. As we have it now as I understand it, if WaveHouse's port is forwarded to the internet, these k8s endpoints become public for anyone to query, whereas (I _think_ I can't remember for sure) we had it setup so that the prometheus metrics, for instance, could optionally be hosted on a different port. Ideally it'd be the same port, or a competent devops person deploying this would filter those paths out at the reverse proxy layer so they were only addressable internally, but unsure if we should build that in as a feature too or not, worth looking into what the k8s norm is.

I DO think, however, we should have a public facing `/health` and/or `/live` or `/ready` or similar, that simply returns 200 OK or an error and likely no content (so we don't need to do JSON encoding on each request and don't need to cache anything), specifically for use in the SDK to check if a server is online before sending data, or to check if a particular server is healthy and accepting data when choosing between multiple in a distributed setup.

Happy to talk about this all more in person if needed today.

clients/ts/src/http.test.ts
taitelee: Reverted those test paths back to /health.

cmd/wavehouse/health.go
taitelee: Went with /livez (liveness) for the Docker HEALTHCHECK and compose service_healthy.

docs/src/content/docs/development.md
taitelee: Yea that api.md line was overstated. Scoped the error field guarantee to failure responses (a healthy /readyz is just {"status":"ready"}, no error field), and added a test asserting success bodies carry no error.

tests/integration/boot_resilience_test.go
taitelee: Yep moved to docs.

EricAndrechek (changes requested):
Only want `/v1/health` accessible and used by API, not readiness.

clients/ts/src/sys.ts
EricAndrechek, line 1: Can we have it return a bool or something instead, since we are getting rid of that Health type? Just looking to make the DX as clean as possible here, and I think them running `const result = await wh.sys.health();`, and then having to check that `result.error` is null feels weird to me – obviously we should still pass any `result.error` through to them for debugging, like whether its a network error or server error etc, but allowing a clear `if (result.ok) {}` path feels cleaner – thoughts?

taitelee: Ok. I think we can add `ok` to `Result<T>` Iadding `ok` flag to the one shared `Result` type that all methods already reutrn) so `if (result.ok)` works everywhere, `health()` stays `Result<void>`, and `error` still passes though.

EricAndrechek (approved):

Issue #196, opened 2026-06-02
feat(sdk): bounded parallel uploading with per-message error handling


EricAndrechek:
Replace the one-at-a-time upload path with **bounded** parallelism (configurable concurrency — not unbounded), keeping per-message error handling so a single bad record doesn't fail or obscure the rest of the batch.

**Acceptance**
- Configurable concurrency limit (sane default)
- Per-message success/error results surfaced to the caller
- Tests for partial failure + backpressure

Part of #194

Issue #194, opened 2026-06-02
epic(sdk): SDK feature rollout after #190 cleanup


EricAndrechek:
Tracking issue for the SDK/API feature work originally scoped into #190.

The **foundation/cleanup** (WebSocket removal, pnpm-workspace consolidation, biome, root Makefile, coverage enforcement) ships in **#190**. The remaining **feature work** is split into focused, individually-reviewable PRs — one per sub-issue below — so nothing balloons into another mega-PR. Each feature PR targets `main` after #190 merges.

## Foundation (shipping in #190)
- [x] Kill WebSocket support; SSE-only transport
- [x] Dedicated SDK Makefile target / root Makefile consolidation
- [x] pnpm workspace consolidation (single root lockfile)
- [x] biome lint + format for TS
- [x] Test coverage enforcement (closes #176)
- [x] Re-enable the SSE public-events streaming test (closes #153)

## Planned features (one PR each)
- [x] #195 — NDJSON upload support (API + SDK)
- [ ] #196 — Bounded parallel uploading with per-message error handling
- [ ] #197 — Live DLQ streaming
- [ ] #198 — More helpful TypeScript errors
- [ ] #199 — Restructure query builder + command naming to match new API routes *(breaking)*
- [ ] #200 — Expose raw response/debug internals (cache HIT/MISS)
- [ ] #201 — More exported, specific error types
- [ ] #202 — Auto-injected dedupe ID for retransmission detection
- [x] #203 — Migrate SSE auth from `?token=JWT` to Fetch `EventSource` (+ deprecate token path) *(closed by #470; retiring `?token=` server-side split out as #468)*
- [ ] #98 — Distinguish NATS replay from genuinely-live events *(breaking)*
- [ ] #204 — SSE replay semantics across multiple subscriptions
- [x] #215 — CORS to SSE Stream API

## Later / not yet scoped (backlog — triage before promoting to Ready)
- [ ] #252 — Receiving data in other formats (e.g. CSV)
- [ ] #253 — On-device SQL validation for the admin route (ClickHouse SQL JS lib or similar)
- [ ] #254 — Groundwork for streaming (non-live) large up/down queries
- [ ] #255 — SSE multiplexing — in-page per-topic + cross-tab (browser leader election)
- [ ] #256 — Local caching shared across workers/subscriptions/tabs (IndexedDB?)
- [ ] #257 — Embedded WASM engine (DuckDB?) for local queries/aggregations on streamed data
[...]

PR #217, opened 2026-06-03
feat(api): multi-format ingest (object/array/NDJSON) + SDK helpers

### LINKED ISSUE #195: feat(sdk): NDJSON upload support (API + SDK)
Add support for uploading newline-delimited JSON (NDJSON) for batch ingest, on both the API and the SDK — a streaming-friendly alternative to sending one large JSON array.

**Acceptance**
- API accepts an NDJSON request body (e.g. `Content-Type: application/x-ndjson`), one record per line
- SDK helper to send NDJSON (string / stream / file)
- Tests for well-formed and malformed/partial input

Part of #194

EricAndrechek:
## Summary

Makes `POST /v1/ingest` **forgiving about input shape** — it now accepts a single flat JSON object, a **JSON array** of objects (any length, even one), or an **NDJSON** batch, and routes the SDK's array `insert` through the batch path with per-record partial-failure reporting. Closes #195 (part of #194).

Bounded parallel uploading + per-message concurrency control (#196) stays deferred to a focused follow-up — see [Scope](#scope-why-not-196).

## API

The body format is **auto-detected — `Content-Type` is only a hint.** The first non-whitespace byte decides (`[` → array, else → single object); an explicit `application/x-ndjson` selects line-framing *unless* the body starts with `[` (the array wins). So a batch works whether or not the header matches.

| Body | Typical `Content-Type` | Response |
| --- | --- | --- |
| one flat JSON object | `application/json` *(default)* | `{"ok":true}` / `{"duplicate":true}` — **byte-for-byte unchanged** |
| a JSON array of objects (any length, even 1) | `application/json` | per-record summary |
| one JSON object per line (NDJSON) | `application/x-ndjson` | per-record summary |

Each record is validated, authorized, deduplicated, and published **independently**, so one bad record never blocks the rest. A batch returns full per-record fidelity — each `results` entry mirrors the single-object response plus its 1-based `index`:

```json
{ "total": 3, "succeeded": 1, "failed": 1, "duplicates": 1,
  "results": [ {"index":1,"ok":true}, {"index":2,"duplicate":true},
               {"index":3,"error":"validation failed: ..."} ] }
```

- **200** whenever the body was readable (branch on `failed`/`results`, not the status). An explicit empty array (`[]`) is a valid record-less batch (200, `total:0`).
- Per-record problems (a bad NDJSON line, a non-object array element, schema validation, column/check denials) → reported in `results`, batch continues.
- Whole-request conditions abort non-200: table-level `insert` denial (**403**, checked once up front); a JSON-array syntax error, a **truncated/unterminated array** (cut-off upload), or an oversized NDJSON line (**400**); the new **16 MiB body cap** (**413**, parity with `/v1/admin/query`); backpressure (**503** + `Retry-After`); publish/dedup-backend errors (**500**).

Formats sit behind a pull-based `recordReader` seam (`internal/api/record_reader.go` — object/array/line readers + a content sniffer) so **CSV and response-streaming slot in later** without reworking the handler; per-record vs whole-request handling stays in the shared `processRecord` helper.

## SDK

- `from(t).insert(rows[])` serializes to a **single** NDJSON request (was: one unbounded concurrent POST per row, returning only the first error). Per-record outcomes surface on `InsertResult` (`total`/`succeeded`/`failed`/`duplicates`/`results`); `ok` is `true` only when every record succeeded. The top-level Result `error` is reserved for whole-request failures.
- New `from(t).insertNDJSON(source)` for NDJSON you already have: `string | Uint8Array | Blob`/`File` | `ReadableStream`.
- Empty array → no-op (no request).
- New exported types: `InsertRecordResult`, `NDJSONSource`. The server-side decoder also accepts raw JSON arrays directly, so non-SDK clients can send whichever shape is convenient.

## Scope: why not #196?

#195 and #196 both touch `insert()` and the per-record result contract, but they split cleanly: **#195 = the wire format + server partial-failure contract; #196 = bounded-concurrency orchestration on top.** This PR resolves the in-code `insert()` TODOs and removes the unbounded fan-out, while designing the per-record contract so #196 layers on without rework. Keeping them separate follows #194's "one focused PR per sub-issue" rule. `insert([...])` currently sends one request regardless of size; chunking very large arrays into bounded-concurrency requests + backpressure-aware retry is #196.

## Tests

- **Go unit** (`ingest_test.go`): single-object + NDJSON (all-valid, partial-failure, malformed line, blank-line skip, empty body, dedup, backpressure 503, publish 500, per-line column-denied + check-clause +auto-inject, table-forbidden 403, charset, results truncation); plus **JSON array** (1-elem / N, per-element validation failure, scalar-element type-error recovery, **syntax error → 400**, **truncated/unclosed → 400**, sniff precedence, empty `[]` → 200, leading-whitespace, pretty-printed object, concatenated-objects-first-only); and **body-cap → 413** (single object / array mid-element / array between-elements / NDJSON).
- **SDK unit**: single NDJSON request shape, partial-failure surfacing (`results`), whole-request error arm, empty-array no-op, empty-200 tolerance, `insertNDJSON` string + non-string, `rawBody`/`contentType` transport.
- **E2E** (`ndjson.test.ts`): array→NDJSON lands in CH, per-record failure persists the good rows, `insertNDJSON(string)`, malformed-line reporting, plus a **raw JSON-array-over-the-wire** case exercising the server's array decoder.

`make ci` green locally: Go total ~88%, all coverage gates pass, e2e 10/10 files.

## Refactor

The format-detection / record-reader machinery was extracted from `ingest.go` (665 → 428 lines) into a new `record_reader.go` (the `recordReader` seam — object/array/line readers + the sniffing factory), and the whole-request abort response DRY'd into a `writeAbort` helper. Behavior-preserving (pure move + dedup).

## Docs

`api.md` (ingest section + Batch Ingest, error tables), `sdk.md`, `AGENTS.md` §3, `CHANGELOG.md`. (User-facing README docs land on a separate branch.)

## At-least-once note

A batch aborted partway (a 503/500, or a JSON-array syntax error, after some leading records were already published) re-publishes those leading records when retried — the same at-least-once property the single-object path already has. Enable dedup if duplicate suppression matters. Documented in `api.md`.

## Also: cache TTL test de-flake

`tests/e2e/sdk/cache.test.ts` ("expires cache naturally based on TTL") asserted `coldQueryMs > hotQueryMs` at millisecond resolution; on fast CI runners both the cold DB query and the hot cache hit measure ~2ms, so the strict `>` flaked (`expected 2 to be greater than 2`). Removed the racy timing comparison — the deterministic `x-cache: MISS → HIT → MISS` header assertions already prove caching works and the TTL expires, and `>=` wouldn't help (jitter can make a hot hit clock slower than a cold miss too). Folded in at the maintainer's request rather than as a separate PR.


taitelee (dismissed):
The current streaming approach for JSON arrays creates a dangerous edge case when a request is cut off and missing its closing `]`. Because we parse and publish records sequentially inside the same loop, any valid objects read before the truncation are immediately sent to NATS and permanently ingested. However, when the reader finally hits the unexpected end of the file, it aborts the HTTP request and returns a 400 Bad Request to the client. This creates a misleading situation where the client believes the entire batch failed, prompting them to retry the full payload. If the user has not configured the optional deduplication system perfectly, this retry will blindly insert duplicate rows for those initially successful records. We are essentially relying entirely on downstream deduplication as a required safety net to cover up HTTP-level network failures. A safer architectural alternative would be to deprecate JSON array batching entirely and strictly enforce NDJSON for all bulk uploads. Because NDJSON is simply line-delimited, it doesn't suffer from "unterminated array" structural errors; a cut-off network stream just drops the final partial line and can still cleanly return a 200 OK with a precise partial-success summary. This approach would give clients an accurate report of exactly which records succeeded, allowing them to retry only the missing rows and removing the strict dependency on background deduplication.

taitelee (approved):

PR #218, opened 2026-06-04
feat(observability): direct-to-cloud OTLP with TLS + auth headers

### LINKED ISSUE #97: feat(observability): direct-to-cloud OTLP — TLS + auth headers (no sidecar)
## Status

**Docs portion shipped in PR #116** — `docs/configuration.md` now has a full OTel section covering `WH_OTEL_ADDR` and every other env var, plus an explicit note on the `WithInsecure()` constraint and the sidecar workaround. `docs/deployment.md` covers the SigNoz/Honeycomb (OTLP-native) and Grafana Cloud / Alloy (Prometheus scrape + OTLP traces via Alloy) wiring patterns.

This issue is staying open and re-scoped to the **actual feature work** that the original docs note pointed at: making WaveHouse able to ship OTLP **directly** to TLS-protected cloud endpoints without an intermediate sidecar.

## Problem (re-scoped)

`internal/observability/provider.go` calls `otlptracegrpc.WithInsecure()` (and the equivalents for metrics + logs). That means WaveHouse can ship OTLP only to **plaintext** receivers — typically a local OTel collector or Grafana Alloy on `127.0.0.1:4317`. Operators who want to ship directly to:

- Grafana Cloud's OTLP gateway (`https://otlp-gateway-prod-us-east-0.grafana.net/otlp` — TLS + Basic auth)
- Honeycomb (`https://api.honeycomb.io:443` — TLS + `x-honeycomb-team` header)
- Datadog OTLP endpoint (TLS + API key)
- Any cloud OTLP endpoint that requires TLS or auth headers

…can't. The standard workaround is a sidecar collector that receives plaintext on localhost and re-exports with TLS + auth configured locally. Annoying for operators running on Fly/Railway/Cloud Run / etc. where adding a sidecar is non-trivial.

## Proposed Solution

Two related additions:

1. **Endpoint scheme sniffing** — if `otel.addr` starts with `https://`, configure the gRPC exporters with TLS credentials (`grpc.WithTransportCredentials(credentials.NewTLS(...))`) instead of `WithInsecure()`. If it starts with `http://`, behave as today. If no scheme, keep current behavior (plaintext) for backward compat.
2. **Headers env var** — `WH_OTEL_HEADERS` (yaml `otel.headers`) accepting comma-separated `key=value` pairs (or repeated env var pattern), wired to `otlptracegrpc.WithHeaders(map[string]string)` etc. on all three exporters.

Together these unlock `WH_OTEL_ADDR=https://otlp-gateway-prod-us-east-0.grafana.net/otlp WH_OTEL_HEADERS=Authorization=Basic\ <base64>` working end-to-end — no sidecar.

## Alternatives Considered

- **Separate `otel.tls.enabled` boolean** — explicit but more knobs than necessary. Scheme sniffing is the standard OTel-SDK pattern (some language SDKs do this natively).
- **A per-signal endpoint override** (different addr for traces vs metrics vs logs) — Grafana Cloud actually does this (different gateway URLs per signal). Worth scoping in: maybe `otel.{traces,metrics,logs}.addr` overrides the top-level `otel.addr` when set. Lower priority; can be follow-up.
- **mTLS** (client cert auth) — niche; defer until requested.

## Additional Context

- Docs from this PR (#116) explicitly call out the limitation in `docs/configuration.md` so operators aren't surprised, and link to this issue.
- Original issue (pre-rescoping) asked for *just* the docs change — that part is done. Re-scoping rather than closing+filing-new because the conversation is here.

taitelee:
## Summary

Exports traces/metrics/logs straight to a cloud OTLP gateway (Honeycomb, Grafana Cloud) with no sidecar collector to terminate TLS or inject auth. Closes #97.

The endpoint, TLS, custom/private CA, mutual TLS, and per-RPC auth headers are all configured through the OpenTelemetry SDK's **standard `OTEL_EXPORTER_OTLP_*` environment variables** — the same vars every Honeycomb / Grafana Cloud quickstart already documents. `InitProvider` passes **no** endpoint/header options and lets the SDK read the env, so WaveHouse owns essentially no config-translation surface of its own.

- **Endpoint + scheme-driven TLS** — `OTEL_EXPORTER_OTLP_ENDPOINT` (and the per-signal `_TRACES_`/`_METRICS_`/`_LOGS_ENDPOINT` overrides). An `https://`URL selects TLS with system root CAs; `http://` or a bare `host:port` stays plaintext. Default `localhost:4317`.
- **Custom CA / mutual TLS** — `OTEL_EXPORTER_OTLP_CERTIFICATE` trusts a private CA; `OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE` / `_CLIENT_KEY` enable mTLS.
- **Auth headers** — `OTEL_EXPORTER_OTLP_HEADERS` (SDK-parsed, percent-decoded per the OTLP spec) carries cloud auth as gRPC metadata on every signal.

### What this removed (vs. the earlier revisions of this PR)
The first cut shipped a WaveHouse-specific translation layer; review steered it toward delegating to the SDK instead. Gone now: the `otel.addr`/`WH_OTEL_ADDR` and `otel.headers`/`WH_OTEL_HEADERS` keys, the per-signal `WH_OTEL_{TRACES,METRICS,LOGS}_ADDR` overrides, the custom header parser and its fail-loud boot validation, the TLS 1.3 floor, and `endpoint.go` (`schemeURL`/`ParseEndpoint`). Set `OTEL_EXPORTER_OTLP_ENDPOINT` in place of `WH_OTEL_ADDR`. The per-signal **enable** gates and **sample-rate** knobs (`otel.enabled`, `otel.traces.sample_rate`, …) are unchanged.

**Trade-off — fail-soft, by design.** Delegating means we inherit the SDK's behavior: a malformed `OTEL_EXPORTER_OTLP_HEADERS` value is logged via the OTel error handler and skipped, not fatal at boot. We deliberately gave up the "refuse to start with bad auth config" property in exchange for zero parser-parity surface.

### The one piece of WaveHouse glue: `logExporterTLSOptions`
The pinned gRPC logs exporter (`otlploggrpc` v0.19) reads `OTEL_EXPORTER_OTLP_[LOGS_]CERTIFICATE` / `_CLIENT_CERTIFICATE` / `_CLIENT_KEY` into its config but **never wires that `*tls.Config` into the gRPC dial** — its
`newGRPCDialOptions` only honors the programmatic `WithTLSCredentials`, otherwise defaulting to system roots. So a custom CA or mTLS supplied via env would silently fall back to system roots **for logs only**, while traces/metrics
(`otlptracegrpc`/`otlpmetricgrpc` v1.43) apply the same env vars correctly. `logExporterTLSOptions` rebuilds the credentials from those standard env vars and hands them to the log exporter via `WithTLSCredentials`, so custom-CA + mTLS work
uniformly across all three signals. It returns `nil` when no custom trust material is set, leaving the public-CA path fully delegated to the SDK. (Confirmed still required as of `otlploggrpc` v0.20 — a version bump alone doesn't fix it; track upstream before dropping the shim.)

### Datadog

Datadog has no public direct-to-cloud OTLP endpoint; its supported path remains the local DDOT Collector embedded in the Datadog Agent, reached as a plaintext receiver. See the deployment guide.

## Related Issues

Closes #97


cmd/wavehouse/main.go
```diff
@@ -104,12 +104,33 @@ func run() int {
...
+		// Parse WH_OTEL_HEADERS here — this is the only validation point for
+		// the header map, deliberately in main rather than config.Validate()
+		// so internal/config stays free of the OTel SDK import graph. A
+		// malformed value is FATAL: the InitProvider error below is non-fatal
+		// (we fall back to stdout), so a bad header reaching the exporter would
+		// silently ship telemetry with no auth — fail the boot loudly instead.
+		// Skipped in Prometheus-only mode: no OTLP push, so headers are
+		// irrelevant and a stale value shouldn't block boot.
+		var headers map[string]string
+		if cfg.OTel.Enabled {
```
EricAndrechek: What's the point of this check, to make sure its otel and not prometheus? I feel like it'd be more clear to me if we check that prometheus was NOT enabled in that case?

internal/observability/endpoint.go
EricAndrechek, line 1: This all feels like something that a package or ideally even otel just bundles and has included/can do for us. Do we have to do all this manual code, is this really boilerplate code that every app that ships with otel options like this has to do?

internal/observability/endpoint.go
```diff
@@ -0,0 +1,111 @@
...
+
+// tlsConfigOrDefault returns c, or a TLS 1.3 floor config when c is nil.
+// credentials.NewTLS panics on a nil *tls.Config; the production default
+// uses system roots with MinVersion=TLS1.3.
+func tlsConfigOrDefault(c *tls.Config) *tls.Config {
+	if c != nil {
+		return c
+	}
+	return &tls.Config{MinVersion: tls.VersionTLS13}
+}
```
EricAndrechek: Is it safe to expect TLS 1.3? My understanding is that a lot of things might still be on older versions than that or that this may deploy on systems with self-signed certs or old protocols, etc, and there doesn't seem to be a way for them to use something else. Is the default go http connection stuff where (I assume) it uses TLS and tries for the highest version it can, etc, not good enough?

docs/src/content/docs/configuration.md
```diff
@@ -120,16 +120,22 @@ The master switch is `otel.enabled`. When `true`, each signal (traces/metrics/lo
...
 | `otel.enabled` | `WH_OTEL_ENABLED` | `false` | Master switch. When `false`, no signals are initialized regardless of the sub-toggles below. |
-| `otel.addr` | `WH_OTEL_ADDR` | `127.0.0.1:4317` | OTLP gRPC endpoint used by every enabled signal. Plain `host:port` — no scheme, plaintext gRPC only (see TLS note above). See `make help` for a local collector setup. |
+| `otel.addr` | `WH_OTEL_ADDR` | `127.0.0.1:4317` | Default OTLP gRPC endpoint for every enabled signal. Accepts `host:port` or `http://host:port` (plaintext) or `https://host:port` (TLS via system root CAs). A trailing URL path is tolerated and stripped — gRPC routes by service name. See `make help` for a local collector setup. |
+| `otel.headers` | `WH_OTEL_HEADERS` | *(empty)* | Comma-separated `key=value` pairs applied as gRPC metadata to **every** OTLP export — the standard auth knob for cloud endpoints (`authorization=Basic <b64>`, `x-honeycomb-team=<key>`). Values may contain `=`; only the first `=` per pair splits key from value, so base64 padding round-trips. Parsed and validated at boot — a malformed value fails startup rather than silently shipping unauthenticated telemetry. |
```
EricAndrechek: So we have a method to add headers to otel now, with the old pre-pr config only letting us have one addr to ship to. This PR adds the headers, but it also adds the ability to set/override delivery for components with custom addresses – but no way to then attach independent headers to each one?

I feel like either we remove the splitting otel per thing for now, or figure out how to wire in custom header options for each, which is a bit more involved and takes some thought to keep the config clean and as minimal as possible.

EricAndrechek (changes requested):

cmd/wavehouse/main.go
taitelee: Hm I think we keep `cfg.Otel.Enabled` here. This block runs whenever either OTLP push or Prometheus is on, because they are one `MeterProvider`. Header parsing only matters for the OTLP push path, so it's gated on `Otel.Enabled`. Switching to `Prometheus.Enabled` would skip header validation unless we're assuming `Otel.Enabled` is true? I think?

internal/observability/endpoint.go
taitelee: Yea I agree with this. I'm making fixes that use the default Go TLS to go straight to `credentials.NewTLS` (nil in prod): system roots, TLS 1.2 floor, negotiate up to 1.3, no pinned floor of our own.

docs/src/content/docs/configuration.md
taitelee: Yea I removed the per-signal addr stuff based on previous comments.

internal/observability/endpoint.go
taitelee: You're right that the SDK bundles most of this. The endpoint/TLS parsing this file used to do by hand is now the SDK's `WithEndpointURL`. It takes the raw `otel.addr`, selects TLS from the `https://` scheme (system root CAs), and strips any URL path itself. I deleted `ParseEndpoint`. The only thing left in this file on the endpoint side is a shim that prepends `http://` to a bare `host:port`, because `WithEndpointURL` needs a scheme to parse and we want a bare host to stay plaintext for backward compatability. N longer building TLS credentials at all. The SDK's default handles this (also relates to comment below).

The one thing I kept is `WH_OTEL_HEADERS` parsing. The SDK's `WithHeaders` wants an already parsed map and exposes no public parser for the `key=value` string, so to keep headers configurable via our yaml/WH_* surface (and validated fail loud at boot) something has to parse it. The alternative is to this is to drop `WH_OTEL_HEADERS` and have operators use the SDK's native `OTEL_EXPORTER_OTLP_HEADERS` env var. What do you think?

EricAndrechek:
Hey @taitelee — I went deeper on this one (plus some research into what the OTel Go SDK gives us out of the box) and want to capture it all in one place so we can steer toward as little custom code as possible. Your `0965a08` refactor was the right direction — dropping `ParseEndpoint`, the per-signal addrs, and the TLS 1.3 floor — this is just continuing that same line of thinking one more step.

**TL;DR:** Almost all of the endpoint/TLS/header plumbing is something the SDK already does natively via the standard `OTEL_*` env vars. We can likely delete `endpoint.go` entirely and shrink the config-translation layer to near-zero.

### What the SDK already does for us

`otlptracegrpc.New(ctx)` (and the metric/log equivalents) with **no options** automatically read the standard env vars — confirmed against our pinned `v1.43.0`:

- `OTEL_EXPORTER_OTLP_ENDPOINT` (+ per-signal `_TRACES_`/`_METRICS_`/`_LOGS_ENDPOINT`) — the scheme drives TLS: `https://` ⇒ TLS w/ system roots, `http://`/bare ⇒ plaintext. This is spec-mandated.
- `OTEL_EXPORTER_OTLP_HEADERS` (+ per-signal) — SDK-parsed.
- `OTEL_EXPORTER_OTLP_CERTIFICATE` / `_CLIENT_CERTIFICATE` / `_CLIENT_KEY` — custom CA + mTLS, for free.

Refs: [otlptracegrpc doc.go](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlptrace/otlptracegrpc/doc.go), [OTLP exporter spec](https://opentelemetry.io/docs/specs/otel/protocol/exporter/). (`go.opentelemetry.io/contrib/exporters/autoexport` goes even further and builds all three signals from env, if we ever want gRPC-vs-HTTP selection too.)

### Concrete items

**1. `schemeURL` → delete.** It only exists because we went through `WithEndpointURL`, which can't parse a bare `host:port`. The native `WithEndpoint("host:port")` + `WithTLSCredentials(...)`/`WithInsecure()` takes the bare form directly, and `OTEL_EXPORTER_OTLP_ENDPOINT` takes a scheme'd URL directly. Either way the shim isn't needed.

**2. `ParseOTelHeaders` has a real conformance bug, not just extra code.** The SDK percent-decodes header values (`url.PathUnescape` inside `stringToHeader`, in `.../otlptracegrpc@v1.43.0/internal/envconfig/envconfig.go`); ours stores them raw. So `WH_OTEL_HEADERS` *looks* exactly like `OTEL_EXPORTER_OTLP_HEADERS` but behaves differently. If an operator copies a vendor's value containing a `%xx` escape (or a percent-encoded comma), we ship the wrong header — silently, with no boot error, which defeats the whole point of the fail-loud parser. Values with a literal `,` also have no escape hatch in our parser. If we keep any custom parser it needs `url.PathUnescape` to match the spec — but better is to not have one at all (see Direction).

**3. Self-signed / custom-CA gap.** Prod TLS is now system-roots-only, and the trust-override hook is `-tags integration` only — so an operator with a private CA or self-signed gateway can't use the TLS path at all. That's the other half of my original TLS comment: the version-floor concern got fixed, but the trust concern is still open. `OTEL_EXPORTER_OTLP_CERTIFICATE` solves it for free. Net: the custom path currently costs more code *and* supports strictly less than delegating to the SDK.

**4. PR description is stale.** It still advertises per-signal `WH_OTEL_{TRACES,METRICS,LOGS}_ADDR`, which `0965a08` removed. Worth fixing so the squash message stays accurate.

### Direction

Let's lean on the standard `OTEL_*` env vars as much as we can. Target end state: `InitProvider` passes **no** endpoint/header/TLS options and lets the SDK read env; `endpoint.go` goes away; docs just point operators at the OTel spec vars (which is what every Honeycomb / Grafana Cloud quickstart already tells them anyway). That collapses the new surface area we have to maintain down to basically nothing.

**One trade-off to weigh:** delegating means we inherit the SDK's fail-*soft* behavior — a malformed header gets logged via the OTel error handler and skipped, not fatal at boot. We'd lose the "refuse to start with bad auth config" property you built. If you feel strongly about keeping fail-fast, the move is a *thin validated pass-through* — validate the value at boot, then hand it straight to the SDK with matching (percent-decode) semantics — rather than a separate reimplementation. I lean toward minimal code + accepting fail-soft, but I'm open if you want to keep the boot check. Your call on that one.

None of this touches the `provider.go` lifecycle work (globals snapshot/rollback, the `sync.Once` around `runtime.Start`, the private Prometheus registry) — that part's solid and stays. This is only about the config-translation layer sitting in front of it.

internal/observability/logtls.go
```diff
@@ -0,0 +1,71 @@
...
+// credentials.NewTLS(nil) (system roots). So a custom CA or mutual-TLS supplied
+// via env silently falls back to system roots for the logs signal, while the
+// trace/metric exporters (v1.43) apply the same env vars correctly.
+//
+// We rebuild the credentials from the standard env vars and hand them to the
+// log exporter via WithTLSCredentials — the path it does honor — so custom-CA
+// and mTLS work uniformly across all three signals. When no custom trust
+// material is configured this returns nil, leaving the public-CA path fully
+// delegated to the SDK (system roots). Drop this once the exporter honors the
+// env TLS config (track upstream before bumping otlploggrpc).
```
EricAndrechek: So is my understanding correct that this entire file is designed to fix an upstream issue in the package we are using? If so, is there an issue tracker for this upstream or anything?

tests/e2e/fixtures/config.yaml
```diff
@@ -41,4 +41,3 @@ policy:
 # so the absence of a collector at :4317 doesn't matter for the suite.
 otel:
   enabled: true
-  addr: "127.0.0.1:4317"
```
EricAndrechek, line 44: With this PR and this change, where will `make test-e2e` runs attempt to ship OTeL to, if its enabled but we don't set any address or OTeL ENV variables? This also means we've broken the `make obs-*` targets too, right? Can we modify the orchestrator in this PR to add the needed OTEL env variables so that they still will work properly?

EricAndrechek (changes requested):

internal/observability/logtls.go
taitelee: Do you mean like our own issue or an open telemetry issue. If the latter there is one: [open-telemetry/opentelemetry-go/issues/6661](https://github.com/open-telemetry/opentelemetry-go/issues/6661)

tests/e2e/fixtures/config.yaml
taitelee: Yea right now with `otel.addr` gone it just falls back to the SDK default, which is `localhost:4317` whici is the same place `otel.addr` used to point to. So make test-e2e itself still passes. Nothing's listening in CI, the gRPC exporter just backs off, and we still get coverage on the export paths.

The `obs-*` targets aren't really broken either. They only start the dashboard listening on `:4317`, they don't run WaveHouse themselves. But you're right that leaning on the implicit default is fragile.

So I had the orchestrator set `OTEL_EXPORTER_OTLP_ENDPOINT=http://127.0.0.1:4317` when it's not already set. That way `test-e2e` + a `make obs-*` dashboard running alongside it ship to the right place out of the box, and you can still override it to point somewhere else. Used `127.0.0.1` instead of `localhost` since `localhost` can resolve to ::1 and the dashboards only bind IPv4.

internal/observability/logtls.go
EricAndrechek: Yeah, the upstream one – my opinion here would be to NOT build this functionality in ourselves, and just acknowledge that this is broken but is not our responsibility, and is being tracked upstream. What do you think?

tests/e2e/fixtures/config.yaml
EricAndrechek: I am not seeing this change in orchestrator

taitelee: Will push after addressing comment above.

internal/observability/logtls.go
taitelee: Ok yea I agree. I think for now we can just acknowledge the issue and track opentelemtry's issue.

EricAndrechek: Yeah, do you want to subscribe to that issue so that if/when it has updates or is closed you know and can address it here?

taitelee: Yep subscribed

EricAndrechek (approved):

Issue #199, opened 2026-06-02
refactor(sdk)!: restructure query builder + command naming to match new API routes


EricAndrechek:
Restructure the structured builder and rename commands/functions to match the newly changed API routes and simplify the surface.

**Breaking** — coordinate a CHANGELOG entry + migration note.

**Acceptance**
- Builder maps cleanly onto the new API routes
- Renames documented with a migration note
- CHANGELOG `breaking-change` entry

Part of #194

EricAndrechek:
**SDK ergonomics input from nas-observability dogfooding (jfwoods, 2026-05-05)** — two friction points worth folding into the query-builder restructure:
1. **`liveQuery` discoverability** — it lives on `QueryBuilder` (`.select().liveQuery(...)`), not on `TableRef`, but live-streaming is the most common reason users reach for `.from(table)`. Consider a `TableRef.liveQuery(...)` shorthand, or at least document the chain in the `TableRef` docstring.
2. **`Database` constraint friendliness** — `DB extends Record<string, Record<string, unknown>>` rejects the natural `interface MyDB { table: MyRow }` (no index signature), forcing a manual `[t: string]: Record<string, unknown>`. Either relax to a structural shape explicit-key interfaces satisfy, or document the workaround prominently.

EricAndrechek:
**Triage note — one constraint to carry into this restructure:** #483 is an open, present-tense contract defect in the current builder. `selectAll()` preserves previously selected columns and `select()` preserves the `selectAll` state, so `select('page').selectAll()` silently changes the projection and `selectAll().count()` reaches the server and 400s. The docs already claim these are mutually exclusive.

#483 was considered for folding into this issue and **kept discrete** — it's live today, and this restructure has no timeline. But whichever lands first, the other has to agree: if this restructure ships before #483 is fixed, the new builder needs to enforce (or explicitly document) that exclusivity rather than inherit the current ambiguity.

Related: #483.

---
_Triage note from the pm-triage routine, 2026-08-25._

Issue #175, opened 2026-05-22
feat: embedded ClickHouse strict validation


EricAndrechek:
Currently our API's schema validation has a janky approach of attempting to do its best to validate JSON field's assumed types to the table's ClickHouse schema. However, given the number and complexity of ClickHouse types, building our own parser is deemed unrealistic and impossile to keep up to date or matching whatever clickhouse version our api is interacting with. As such, we have a few options (excluding building our own parser):

## Option 1: Fast and Permissive

We don't attempt to validate types – or at least not all of them. At most we do a small subset of them with highly permissive fallbacks if they fail softly. This approach would basically just be our API doing fast JSON validation, maybe checking to see that for that table all required columns were present, and that perhaps a small portion of the typed columns looked roughly correct.

### Pros

- Super fast
- Can be very light on memory if not serializing/deserializing JSON (although if in our custom format would need deserialized/reserialized anyway for proper ClickHouse insert format – unless our API/SDK build the ClickHouse JSON insert format by default anyway)
- Never rejects data ClickHouse would have accepted (no false negatives on schema validation)

### Cons

- Non-standardized data being given to clients (ie timestamp normalization, etc) that normally happens in ClickHouse
- Nullable columns with ClickHouse's schema setting defaults not being filled in for streams to clients
- Allows for false positives on schema validation (allowing in events that don't meet the schema requirements and are rejected by ClickHouse at buffered ingest), which leads to cascading side-effects including:
  - Invalid data being included in our streams to connected clients, which now must know how to handle errors
  - Slowing down the buffered ingest process with higher failure rates
  - Improperly communicating to a client sending data for insertion that it was accepted when it was rejected later

## Option 2: ClickHouse NULL Engine Validation

Our process takes the tables it detects from the schema refresh and builds NULL engine tables in ClickHouse for internal use by our process only, which a different NULL table for each schema. Our API then does its normal roles/permissions checks on ingest, but then attempts to insert the data into the ClickHouse NULL engine table matching its desired table. This (I think?) can be done quickly/synchronously compared to how normal buffered inserts/ingest need to work, so it could give us a semi-instantaneous response on if the data was accepted (which does nothing – the data is just discarded in ClickHouse (or eventually could flow to materialized views, etc, and we could skip our buffer/ingest step 👀 )) or rejected. We then return the reject/accept and if accepted pass it along to continue our normal process.

### Pros

- Tracks with the schema/versions of the actual version of ClickHouse a client is using
[...]

PR #272, opened 2026-06-05
fix(sdk): stop hardcoding received_timestamp in default query order

### LINKED ISSUE #270: bug(sdk): default ORDER BY / pagination column hardcoded to received_timestamp 500s on BYOS tables
**Area:** sdk · query — footgun (DX); breaks the zero-config query path on BYOS tables · found via WaveHouse-Stats dogfooding *(reproduced live on `71c2dc4`)*

**Expected:** `wh.from(table).fetch()` / `.select(…)` without an explicit `.orderBy()` works on any discovered table.

**Actual:** the SDK query builder injects `ORDER BY received_timestamp DESC` by default (also the pagination cursor: `orderCol = this._state.orderBy[0]?.column ?? "received_timestamp"`). On a table without that column it emits invalid SQL → ClickHouse `code 47, Unknown expression identifier received_timestamp` → HTTP 500. `received_timestamp` is the WaveHouse-Demo `clicks` column; for a bring-your-own-schema gateway it shouldn't be a hardcoded default.

**Impact (Stats):** `gh_events` uses `event_ts` (DateTime64), so every default-ordered query 500s naming a column they never referenced — looks like a setup bug. Workaround: always pass `.orderBy('event_ts','desc')`.

**Scope (DX):** derive the default order/cursor from the table schema (e.g. its `ORDER BY` key), or make it per-client/per-table configurable; at minimum omit the default `ORDER BY` when that column isn't present.

Related: SDK epic #194; query-builder restructure #199; the 500-vs-4xx error-semantics issue (filed alongside) masks this as a server error.

---
_From WaveHouse-Stats `WAVEHOUSE-FEEDBACK.md` dogfooding; reproduced live on dev `71c2dc4`, 2026-06-04._

EricAndrechek:
## Summary

`wh.from(table).fetch()` hardcoded `ORDER BY received_timestamp DESC` as both the default query order and the cursor-pagination column. On bring-your-own-schema tables without that column, the SDK emitted invalid SQL → ClickHouse `Unknown expression identifier received_timestamp` → HTTP 500.

This removes the hardcoded default. The SDK now sends `ORDER BY` **only** when the caller sets one via `.orderBy()`:

- `_buildAST` no longer injects the `received_timestamp` default — no `order_by` is sent unless `.orderBy()` was called.
- `_fetchNext` derives the keyset cursor from the explicit order (`this._state.orderBy[0]`), with no hardcoded fallback; it returns a terminal empty page if there is no cursor.
- `fetch()` attaches `next()` only when there is an explicit order, but still reports `hasMore` honestly from the row count — so `hasMore: true` with `next: undefined` is a valid, honest state.

Net: `.fetch()` works on any schema (no 500). Deterministic pagination requires an explicit `.orderBy()`. This also sidesteps the #175 default-cursor tie bug — there is no longer a default cursor to tie on.

**Soft behavior change:** `result.next` is now `undefined` (rather than always defined when `hasMore` is `true`) unless the query has an explicit `.orderBy()`. Callers paginating via `await result.next!()` should guard on `result.next` (the `sdk.md` / README examples already do).

### Why no client-side `defaultOrderBy` option?

An earlier draft of this PR added an opt-in `options.defaultOrderBy` to the SDK. It was dropped: a client-global default is the wrong granularity (one order for every table — and it re-introduces the invalid-SQL footgun for any queried table lacking that column), and it duplicates per-table sort-key knowledge in every client and language. The right home for a default sort order is the **backend**, per-table, designed together with pagination (ideally a server-side cursor token that would also lay the #175 tie bug to rest). Tracked as a follow-up.

## Tests

- `clients/ts/src/query-builder.test.ts`: a bare `.fetch()` now omits `order_by`; `.limit(n).fetch()` with no order → `hasMore: true, next: undefined`. The explicit-`.orderBy()` keyset-cursor test still pins the pagination path.
- `tests/e2e/sdk/query.test.ts`: regression test — a table created **without** `received_timestamp` returns `error === null` from a bare `.fetch()`. The previously-skipped #175 default-cursor test is repurposed to assert the new `hasMore`-without-`next()` contract.

## Docs

`docs/src/content/docs/sdk.md`: the Pagination section now states `next()` requires an explicit `.orderBy()` (and corrects a pre-existing "exactly limit" → "at least limit" inaccuracy). `CHANGELOG.md`: `### Fixed` entry under `## Unreleased`.

## Out of scope

`clients/ts/src/stream/live-query.ts` and `stream/sse.ts` use `received_timestamp` for stream dedup — left as-is (the live ingest stream always carries that column).

Closes #270. Relates to #175 (default-cursor tie); builds on #199 (query-builder structure). The backend-owned, per-table default sort order + server-side pagination is tracked as follow-up #274.

> **Branch history:** two commits — `eb236e4` (first cut, which added an `options.defaultOrderBy`) and `6eb4e26` (drops it, per the design discussion). Net diff vs `main` is the clean 5-file core fix; intended to squash-merge.


EricAndrechek (commented):
lgtm

Issue #12, opened 2026-04-13
Integrate Bento Stream Metrics into OTel Pipeline


taitelee:
## Problem

WaveHouse uses Bento (Benthos) for complex stream processing and batching to ClickHouse. Bento has its own internal metrics (error rates, batch sizes, latency), but these are currently trapped inside the Bento instance and do not appear in our central SigNoz dashboard.

## Proposed Solution

Map Bento's internal metrics system to our global OTel MeterProvider.

- Configure the Bento service.Resources to use an OpenTelemetry-compatible exporter.
- Capture key processor metrics: processor_errors, output_sent, and batch_latency_ms.
- Ensure these metrics share the same service.name resource attribute as the rest of WaveHouse to allow for correlation in SigNoz.

## Alternatives Considered

Integrating Bento's Prometheus Exporter by opening a separate port (e.g., 4195) for Bento stats. We decided to keep a single OTLP pipeline for all telemetry.

## Additional Context

None

PR #273, opened 2026-06-05
fix(query): enforce column allowlist on every structured-query clause

### LINKED ISSUE #223: [security] structured-query column allowlist not enforced on every column reference (omit/`*`, group_by, order_by, filters)
> **Scope update (2026-06-05):** Fixing this surfaced that the omitted-columns `SELECT *` case is **one of a family** of fail-opens with the same root cause — the structured-query handler authorized only the columns a caller *explicitly listed*, while `query.Build` validated every *other* column reference against the schema only. The same denied column also leaks (or is inferable) via `columns:["*"]`, `group_by`, `order_by`, and `filters`. All are fixed together in #273. Full breakdown below the original report.

---

**Area:** query · policy — **security** (data exposure) · found via WaveHouse-Stats dogfooding

**Expected:** a role's `allow_columns`/`deny_columns` is a hard cap on what columns it can read, on every read path.

**Actual:** `POST /v1/query` validates only the columns a caller *explicitly* lists (`internal/api/structured_query.go:86-91` loops over `sq.Columns`). Omit `columns` (or send `[]`) and the loop never runs, the builder emits `SELECT *`, and **every column is returned**, ignoring the allowlist. There is no output-side column filtering on this path. The SSE/stream path *does* filter output columns (`applyStreamPolicy` → `filterEventColumns`), so the structured-query path is the outlier.

**Impact (Stats):** an unauthenticated `{"limit":2}` against `gh_events` returned `payload` (raw webhook JSON), `source`, `actor_id`, `repo_id`, `event_id` — all meant to be hidden from the public role. For a private dataset this is a content leak.

**Why P0 / launch-blocking:** fail-open data exposure on the primary query path; the repo cannot be flipped public (or run an exposed alpha) with it live. Should be a precondition on the v0.1.0-alpha.1 release (#149).

**Note:** WAVEHOUSE-FEEDBACK.md marks this "patching upstream (Eric, 2026-06-04)" — if a fix is already in progress, link the PR here; this issue tracks it either way.

---

---

## Expanded scope & fix (#273)

**Root cause:** `internal/api/structured_query.go` checked only `sq.Columns` and `sq.Aggregations` against the allowlist; `internal/query/builder.go` (`Build`) validated `filters`, `group_by`, `order_by`, and `time_range` columns against the *schema* only — never the role's policy. So beyond the reported omitted-columns case, a denied column also leaks or is inferable via:

| Input | Leak |
| --- | --- |
| `columns: []` / omitted | `SELECT *` → every column (the originally-reported case) |
| `columns: ["*"]` under a deny-list | `IsColumnAllowed("*")` returned `true` when `allow_columns` was empty → `SELECT *` → denied columns returned |
| `group_by` on a denied column | `SELECT count(*) … GROUP BY salary` enumerates every distinct value of the denied column — **CRITICAL** |
| `filters` on a denied column | the returned row count becomes a value-inference oracle |
| `order_by` on a denied column | leaks ordering / extremes |

**Fix:** column authorization moves into `query.Build` — the single chokepoint that already enumerates every column reference (projection, aggregation args, `filters`, `group_by`, `order_by`, `time_range`) — so no clause can skip the allowlist. A column-restricted role's `SELECT *` expands to exactly its allowed columns (`policy.AllowedProjection`); unrestricted roles and admin keep `SELECT *`; a role with no readable columns fails closed with `403`. The per-column decision (`policy.IsColumnAllowed`) is now shared with the stream path (`filterEventColumns`), and role-specific SQL partitions the read cache by column visibility for free. Red-team tests cover every shape at unit / handler / e2e layers.

**Adjacent fix (separate vuln class, also in #273):** the aggregation `alias` was interpolated into the SELECT list verbatim (`fn(col) AS <alias>`) with no validation — a SQL-injection vector (e.g. `alias: "n FROM secrets --"` reparents the query past the table check). Now validated as a bare identifier. This was **not previously tracked**.

**Intended behavior change:** authorizing `order_by`/`group_by`/`filters` means a column-restricted role can no longer order/group/filter by a column it can't read — so the SDK's default `.fetch()` pagination (orders by `received_timestamp`) now needs a restricted role to allow that column or pass an explicit `.orderBy()`.

### Related findings (tracked separately — **not** in #273)
- **#224** — policy `check` honors only `_eq` (`_neq/_gt/_lt/_in` silently dropped); `_in` filter accepted but never enforced.
- **#270** — SDK's hardcoded `received_timestamp` default order/cursor (the column the behavior-change note above interacts with).
- **Latent / defense-in-depth (no dedicated issue):** the read cache key (`queryCacheKey(sql, params)`) omits the caller's role/scope. It's safe today only because column visibility (post-#273) and row filters both land in the SQL; if tenancy ever moves to `scope`-based isolation it would cross-serve. Should be folded into the scope/multi-tenant work (**#214**, **#235**) when `scope` is populated. (Distinct from **#261**, which is cache-key *determinism* for multi-column filters.)

EricAndrechek:
## Summary

Closes #223. The structured-query path (`POST /v1/query`) treated a role's `allow_columns`/`deny_columns` as a cap on only the columns a caller **explicitly listed**. Several inputs slipped past it and returned (or inferred) columns the role may not read — a fail-open data-exposure bug on the primary read path. While fixing the reported case I found the same root cause in three more clauses (a parallel security recon confirmed them), so this PR closes the whole family with one structural fix.

### What was exploitable

| Input | Result before this PR |
| --- | --- |
| Omit `columns` (or `[]`) | builder emits `SELECT *` → **every column returned** (the reported leak: an unauthenticated `{"limit":2}` returned hidden columns incl. raw webhook `payload`) |
| `columns: ["*"]` under a deny-list | `IsColumnAllowed("*")` returned `true` when `allow_columns` was empty → `SELECT *` → denied columns returned |
| `group_by: ["salary"]` (denied) | `SELECT count(*) … GROUP BY salary` → **enumerates every distinct value** of a denied column |
| `filters` on a denied column | row-count differences make it a **value-inference oracle** |
| `order_by` on a denied column | leaks ordering / extremes |

Only the SELECT list and aggregation arguments were ever checked. The SSE/stream path was already safe (`filterEventColumns`), so the structured path was the outlier.

## The fix — one chokepoint, not scattered checks

Column authorization now lives in the **single place that already enumerates every column reference**: the SQL builder (`query.Build`). It validates *and* authorizes the projection, aggregation args, `filters`, `group_by`, `order_by`, and `time_range` in one pass, so no current or future clause can silently skip the check — scattering the checks across the handler is precisely what produced #223.

- **`SELECT *` is resolved, never emitted blindly.** A full-row read (omitted `columns`, or `["*"]`) by a *column-restricted* role expands to exactly that role's allowed columns (`policy.AllowedProjection`), so denied columns never reach ClickHouse. *Unrestricted* roles and the admin role keep `SELECT *` unchanged (`policy.RestrictsColumns` decides) — zero behavior change on the hot path, and special-character/lagging-schema columns still work for them.
- **Fail closed.** A role allowed the table but no columns gets `403` (`query.ErrNoReadableColumns`), never a bare `SELECT *`.
- **Typed errors** (`ForbiddenColumnError`/`ForbiddenAggregationError`) map to `403`; malformed queries stay `400`.
- **Cache isolation for free.** Because restricted roles now produce role-specific SQL, the existing `queryCacheKey(sql, params)` partitions by column visibility automatically — one role can't be served another's cached columns. (No cache-key change needed; the SQL fully determines the result.)

### DRY / single source of truth

The decision stays in one function, `policy.IsColumnAllowed`, now consumed by both read surfaces:

- structured query → new `policy.AllowedProjection` / `RestrictsColumns` (batch form of the same decision)
- live stream → existing `filterEventColumns`

`IsColumnAllowed("*")` now returns `false` (a literal `"*"` is resolved by the projection layer, not waved through) — closing the deny-list footgun behind the `["*"]` variant.

## Tests (each bypass shape pinned as an explicit red-team case)

- **`internal/policy`** — `AllowedProjection`, `RestrictsColumns`, `IsColumnAllowed("*")`, and an invariant test that the "restricted?" and "allowed?" answers can never disagree.
- **`internal/query`** — a denied column rejected in *every* clause; wildcard expansion (restricted → allowed cols in schema order; unrestricted → `*`); the `["*"]`/deny-list sibling; empty-projection → `ErrNoReadableColumns`; `order_by` alias not over-restricted; `count(*)` still allowed.
- **`internal/api`** — handler-level with a SQL-capturing conn: omitted columns → `SELECT page, ts` (never `*`); `["*"]` expands; `group_by`/`filter`/`order_by` on a denied column → `403` and **never reaches ClickHouse**; unauthenticated `{"limit":2}` → only default-role columns; cache keys differ across roles.
- **`tests/e2e/sdk/query.test.ts`** — full pipeline (SDK → WaveHouse → ClickHouse) under a restricted `viewer`: omitted columns and `select("*")` return only allowed columns; `group_by`/`filter`/`order_by` on a denied column → `403`; no-readable-columns → `403`.

Unit coverage: `internal/query` 93.7%, `internal/policy` 94.6%, `internal/api` 81.8%. `make ci` green (unit + integration + e2e + merged coverage).

## No SDK code change

Purely a backend behavior fix — a `@wavehouse/sdk` user's code is unchanged; `.fetch()`/`.select("*")` now just get correctly-scoped results. Docs (`api.md`, `access-control.md`, `sdk.md`) and `AGENTS.md` (design invariant #12) updated; `CHANGELOG.md` under `### Security`.

## Bonus: a SQL-injection hole fixed in the same function

The pre-push review surfaced a **pre-existing** SQL-injection vector adjacent to #223 and living in the exact function this PR refactors: `aggregationExpr` interpolated the user-supplied aggregation `alias` into the SELECT list verbatim (`fn(col) AS <alias>`) with no validation. An alias like `n FROM secrets --` reparents the query past the table-policy check. Since it's in code this PR now owns and the same `validIdentifierRe` already guards ORDER BY aliases, I closed it here (validate the alias as a bare identifier) rather than leave a known injection. Red-team test covers comment-reparent / statement-break / subquery payloads. Flagging it prominently because it's technically orthogonal to the column-allowlist family — shout if you'd rather it be split into its own PR.

## Behavior change to note (intended)

Because `order_by`/`group_by`/`filter` columns are now authorized, a **column-restricted role can no longer order/group/filter by a column it can't read** (Hasura-style). The one user-visible edge: the SDK's bare `.fetch()` paginates by `received_timestamp` (its default keyset cursor → `ORDER BY received_timestamp`), so a restricted role must include `received_timestamp` in `allow_columns` or pass an explicit `.orderBy()` over a readable column. Unrestricted roles (`allow_columns: ["*"]`, the common case) and admin are unaffected. Documented in `sdk.md` and `access-control.md`; pinned by an e2e test.

## Out of scope — found during the recon, flagged for a decision (not fixed here)

Both are real but **not exploitable on `main` after this PR**; I did not file separate issues (happy to if you want):

1. **Read cache key omits role/scope** (`cache_key.go`, `scope` is hard-coded `""`). After this PR, isolation holds *by coincidence* — restricted SQL differs by columns, RLS differs by `WHERE`/params. If tenancy is ever moved to `scope` instead of RLS-in-`WHERE`, the cache would cross-serve. Defense-in-depth: fold the resolved role/scope into `queryCacheKey`.
2. **Policy `check` clauses silently ignore non-`_eq` operators** (`policy.go`). A `check` written with `_neq`/`_gt`/`_lt`/`_in` produces **no enforcement** and no error — a fail-open authoring footgun. Suggest rejecting non-`_eq` at `policy.Validate` time (or implementing them).


taitelee (dismissed):
Good to me. I spent most of my review looking over `builder.go`. 

- `SELECT *` is resolved, never emitted blindly `(resolveProjection` in `builder.go`). A column restricted role's full row read (omitted columns, or ["*"]) expands to exactly that role's allowed columns via `policy.AllowedProjection`. Unrestricted roles and admin still get a real `SELECT *`. `policy.RestrictsColumns` decides which case you're in, so there's zero behavior change on the hot path.
- Fail closed. A role allowed the table but no readable columns now gets a 403 (`ErrNoReadableColumns`), not a bare wildcard.
- `IsColumnAllowed("*")` now returns false. A literal `"*"` is no longer waved through. The all columns wildcard is resolved by the projection layer instead. That closes the `["*"]` deny list variant.
- Typed errors: HTTP status (`structured_query.go`): `ForbiddenColumnError` / `ForbiddenAggregationError` / `ErrNoReadableColumns` are 403. Malformed queries stay 400.
- The per-column rule stays in policy.IsColumnAllowed, now used by both the structured path (via `AllowedProjection`) and the live stream path (the existing `filterEventColumns`), so they can't drift apart.
- Cache isolation for free. Because restricted roles now produce role specific SQL, the existing `queryCacheKey(sql, params)` partitions by column visibility automatically. One role can't be served another's cached columns. I think we are deferring including roles in the cache or it's not necessary anymore because of the role specific SQL queries.
- Injection closed (`aggregationExpr` in `builder.go`): the alias is now backtick quoted as a bare identifier via `chsql.QuoteIdent`, the same treatment `ORDER BY` aliases already got.

Overall looks good to me. This seems to cover all cases with strict roles.

taitelee (approved):

PR #300, opened 2026-06-08
docs: fix stale error bodies and document missing internal packages

taitelee:
## Summary

- api.md: fixed the "validation failed: ..." error body. That string is never returned by the API (it's only a log line); replaced both occurrences (the ingest error table and the batch-results example) with the actual validator messages (unknown column "x" for table "y", etc.).
- configuration.md: documented the full prometheus.path validation. Added that it can't collide with reserved probe paths (/livez, /readyz, /healthz, /health, /ready) or shadow /v1 when port: 0, and that a bad value fails at startup.
- configuration.md: synced the Example Config File with config.yaml. Added the missing server.cors_allowed_origins and clickhouse.query_timeout.
- architecture.md: added the auth/ and chsql/ packages. Added both to the package tree, promoted auth/ into its own subsection (was a buried bullet), and added a new chsql/ subsection.
- development.md: added auth/, chsql/, and observability/ to the Project Structure tree (all three were missing).
- AGENTS.md: fixed the stale ingest-worker claim. "validates the table name" → "accepts whatever table name the envelope carries (existence is checked by the HTTP handler, not re-validated here)," matching the code.
- AGENTS.md: added auth/ and chsql/ to the package list (count "Eleven" → "Thirteen") and the File Structure tree, keeping it consistent with the docs trees.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,14 +26,16 @@ One binary:
 
 - **`cmd/wavehouse/`** — Standalone mode (all-in-one with embedded NATS, optional Pebble dedup)
 
-Eleven internal packages under `internal/` (plus `internal/testutil/` for shared test helpers):
+Thirteen internal packages under `internal/` (plus `internal/testutil/` for shared test helpers):
 
-- **`api/`** — Chi HTTP router, JWT/JWKS middleware, ingest/query/structured-query/SSE/schema/DLQ/policy/pipes handlers, Hub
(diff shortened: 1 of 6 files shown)


taitelee:
@EricAndrechek 

When I search something in the docs, click on whatever I found, and then click the search box again it comes up without the actual search bar. This is a frontend bug.

Also, there are still mentions of scope for the api. I know it’s still a todo but I don’t think we should keep references of it if it’s not implemented yet. I feel like as a reader I’d get confused trying to figure out what scope means just to find no core documentation or implementation on it. Should we either mark it as not yet implemented (currently always an empty field for a NATS message) or just get rid of its references in the doc?

EricAndrechek (approved):

PR #314, opened 2026-06-10
feat(cache): namespace-versioned cache keys for queries and pipes

### LINKED ISSUE #178: feat(cache): pipe cache invalidation via table/scope extraction from named queries
## Overview

Named pipe ("pipes") cache entries are currently namespaced by pipe name (`safePipeName`) rather than by the underlying tables they query. Because Bento ingest invalidates cache entries by table name after successful ClickHouse inserts, pipe cache entries are **not** invalidated on writes and can remain stale until TTL expiry, defeating read-your-writes for pipe consumers.

## Context

In the current implementation (`internal/api/pipes.go`), the cache namespace for a pipe execution is set to the pipe name:

```go
// TODO: scope impl
scope := ""
safePipeName := query.SafeEncodeNATS(name)
```

However, the Bento ingest worker invalidates cache by table name (via `InvalidateCache(ctx, table, scopes)`). Since the pipe cache entries are keyed under the pipe name namespace rather than a table namespace, ingest-triggered invalidation has no effect on them.

## Problem

Unlike structured queries (where the table is an explicit request parameter), named pipes contain arbitrary SQL that may reference one or more tables. There is currently no mechanism to:
1. Parse the tables referenced by a pipe's SQL at registration or execution time.
2. Map those table names to cache namespaces so ingest invalidation can evict stale pipe results.

## Proposed Solution (to be designed in follow-up)

Some approaches to consider:
- **Static analysis at pipe registration**: parse the SQL in `PUT /v1/admin/pipes/{name}` to extract referenced table names and store them alongside the pipe definition. Use these stored table names as cache tags on `Set` and ensure `InvalidateByTags` / `InvalidateCache` covers them.
- **Runtime extraction**: use the same naive SQL regex already used for structured queries (`FROM`/`JOIN` extraction) at pipe `Execute` time to derive table names dynamically and pass them as cache tags.
- **Explicit metadata on pipe definition**: allow pipe authors to declare `tables: ["clicks", "events"]` in the pipe JSON, which are used as cache invalidation tags.

## Acceptance Criteria

- [ ] Pipe cache entries are invalidated when a Bento ingest flush successfully inserts data into a table referenced by the pipe's SQL.
- [ ] Read-your-writes is upheld for pipe consumers after an ingest event.
- [ ] The solution handles pipes referencing multiple tables (multi-table invalidation).

## References

- Introduced in PR #177 (cache-invalidation): https://github.com/Wave-RF/WaveHouse/pull/177
- Specific comment: https://github.com/Wave-RF/WaveHouse/pull/177#discussion_r3293577785
- Related issues: #85, #73

/cc @EricAndrechek

taitelee:
Summary
Adds a versioned, namespace-based cache-key scheme to internal/cache that underpins cache invalidation for both structured queries and (eventually) named pipes. A cached result is keyed by every (table, scope) it depends on — one for a structured query, several for a pipe — so a write to any dependency invalidates it. Groundwork for per-pipe cache invalidation (#178).

Additive and dormant: it sits alongside the existing flat versions map (which still backs the live Cache.Get/Set path), and nothing in the read/ingest paths is wired to it yet, so no behavior change.

QueryKey folds the table version and namespace version of every (table, scope) a result depends on into its key, so a bump of any dependency misses it. A structured query passes one Namespace; a multi-table query (pipe) passes several — same QueryKey, same BumpTable/BumpNamespace.

Intended invalidation policy (once the ingest worker is wired): a write to a named scope → BumpNamespace(table, scope) (that scope + the whole-table view; other scopes survive); a no-scope / whole-table write → BumpTable(table) (whole table in O(1)).

Need todos:
- Wire the structured-query handler + ingest worker onto GetQuery / BumpTable / BumpNamespace; retire the legacy flat path.
- Pipe caching: resolve a pipe's table dependencies and build the []Namespace (lives in the pipes package).
- Unit tests for the new helpers + a CHANGELOG entry.

From Claude: Heads-up on scope: these changes are groundwork and not active on the live path yet. The new tableVersions/namespaceVersions registries and the BumpTable/BumpNamespace primitives exist, but nothing calls them on a real ingest — the worker still invalidates through the legacy flat versions map, and the query/pipe handlers still read through the legacy Cache.Get/Set. So a real write today does not update the new tables or namespace maps, and GetQuery isn't reachable from any endpoint. Making it effective is a follow-up that wires the handlers onto GetQuery and the worker onto BumpTable/BumpNamespace (choosing which by the ingest's scope), reaching into internal/api and internal/ingest. This PR is intentionally additive with no behavior change.


internal/cache/local.go
```diff
@@ -28,33 +28,44 @@ func NewLocal(maxCost int64) (*LocalCache, error) {
...
+// scope at once); a non-empty Scope bumps just that scope plus the whole-table
+// view. Returns the number of namespaces processed.
+func (l *LocalCache) Invalidate(_ context.Context, namespaces []Namespace) (uint64, error) {
+	for _, ns := range namespaces {
+		if ns.Scope == "" {
+			l.versionManager.BumpTable(ns.Table)
+		} else {
+			l.versionManager.BumpNamespace(ns.Table, ns.Scope)
+		}
 	}
```
EricAndrechek, line 73: So with how you've changed this, previously iirc it was smart enough/optimized enough such that when it built the version keys, they were as small as possible. Now though, with lots of namespaces to invalidate, we have a lot of extra work being done here. For instance, if namespaces includes `[{Table: "events", Scope: ""}, {Table: "events", Scope: "org1"}, {Table: "events", Scope: "org2"}]`, _technically_ if I'm understanding correctly, we should **only** have to invalidate the Table – the namespace bumps are meaningless when the table is being bumped anyway, no?

taitelee: Ok I see. I changed `Invalidate` to have two passes: bump the tables first, then bump only the scopes whose table wasn't entirely bumped. So `[{events,""}, {events,"org1"}, {events,"org2"}]` collapses to a single `BumpTable("events")`. It's order independent (all table bumps resolved before any scope decision), and a consistently scoped batch with no empty scope entry still bumps each scope as before. Added a test for the mixed batch case.

internal/api/structured_query.go
```diff
@@ -126,13 +126,15 @@ func (h *StructuredQueryHandler) Handle(w http.ResponseWriter, r *http.Request)
 	// TODO: impl scope
 	scope := ""
 	safeTableName := query.SafeEncodeNATS(table)
+	// A structured query reads one table, so it depends on a single namespace.
+	deps := []cache.Namespace{{Table: safeTableName, Scope: scope}}
```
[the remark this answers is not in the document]
taitelee: Agreed and fixed.

EricAndrechek (approved):

PR #357, opened 2026-06-29
fix(policy): case-normalize aggregation before denied_aggregations check

### LINKED ISSUE #318: security(policy): denied_aggregations bypassed by case (deny check runs before lowercasing fails open)
**Area:** policy · query — security (fail-open) · found via pre-launch audit

**Expected:** a `denied_aggregations` entry (e.g. `sum`) blocks the aggregation regardless of how the caller cases it (`SUM`, `Sum`, `sum`).

**Actual:** the deny check runs **before** lowercasing, so `SUM(...)` does not match the lowercase `sum` deny entry and is allowed — deny does not win (fail-open).

**Impact:** a policy that denies an aggregation is bypassed by changing case; a denied (expensive or sensitive) aggregation runs anyway. Same fail-open class as #224 — cheap to close, wrong direction to leave open.

Scope: normalize case before the deny comparison.

Related: #224 (policy `_in` fail-open), #168 (capability-based per-feature permissions), epic(security) DoS hardening (uncapped aggregations).

---
_From WaveHouse-Stats pre-launch security audit (`WAVEHOUSE-FEEDBACK.md` dogfooding), audited dev `60fed15` (2026-06-10). Filed via /pm-triage._

taitelee:
## Summary

The `denied_aggregations` policy check was fail-open against case variation. In `IsAggregationAllowed` (internal/policy/policy.go), the deny loop compared the lowercased deny entry against the raw caller-supplied function name, which was only lowercased later after the deny check and after the empty-allow-list early return. Since `isValidAggFn` already accepts any casing (`SUM`, `Sum`, `sum`), a denied aggregation like sum was bypassed simply by sending `SUM(...)`, and a denied (expensive/sensitive) aggregation ran anyway.

Fix was to normalize the function name to lowercase once, before the deny check, so both the deny and allow comparisons are case-insensitive. 

## Related Issues

Closes #318


EricAndrechek (approved):
lmao I can't believe we literally had the toLower case logic there for allowed case just not for the denied case, easiest fix ever

PR #358, opened 2026-06-29
fix(policy): enforce _in on filter and check paths

### LINKED ISSUE #224: Policy filter/check: `_in` accepted but never enforced; `check` honors only `_eq`
**Area:** policy — footgun (silent under-enforcement) · **security** (fail-open on the filter path)

**Summary:** the policy `Filter` schema accepts operators the engine never enforces. An `_in` clause is **accepted but produces no predicate on any path**, and on the `check` path *every* operator except `_eq` is silently dropped. No error is raised at config load or request time.

**Evidence (`internal/policy/policy.go`):**
- The `Filter` struct (`:42-49`) declares `_eq/_neq/_gt/_lt/_in`. **`In` is never read anywhere in the tree** — a grep for the field returns only the declaration.
- `resolveFilters` — the row-level `filter` / SELECT path (`:180-206`) — branches on `_eq/_neq/_gt/_lt` only; there is **no `_in` branch**, so an `_in` filter yields no `WHERE` clause.
- The `check` / INSERT path (`:167-174`) reads only `f.Eq`; `_neq/_gt/_lt/_in` are all dropped.
- `Validate` / `validateRolePerms` (`:300-335`) rejects none of these, so the misconfiguration is silent.

**Impact:**
- **Filter path = fail-open (the dangerous half).** A row-security filter like `tenant_id: { _in: [...] }` produces no predicate, so the role sees **all** rows instead of the intended subset — the same failure family as the SELECT-* column-allowlist leak #223, but triggered by an operator authoring an `_in` clause rather than by default config.
- **Check path** silently drops insert restrictions written with anything other than `_eq`.

**Fix (two phases):**
1. **Cheap pre-launch guard:** reject `_in` (and any operator the engine won't enforce) at `policy.Validate()` — turns a silent fail-open into a loud config error. Small; worth landing before the public flip even if full enforcement comes later.
2. **Full fix:** enforce `_in` → `col IN (?)` (array binding) on both paths, and decide whether `check` should honor the comparison operators or keep rejecting them.

**Docs:** keep the `access-control.md` operator table's note that `_in` is "not yet enforced" until phase 2 lands — and update/remove that caveat when it does. (The note was just added by a docs reviewer; it's accurate interim cover.)

Related: SELECT-* column-allowlist fail-open #223 (same family, default-config trigger).

---
_Originally filed from `WAVEHOUSE-FEEDBACK.md` (WaveHouse-Stats dogfooding) as "check clauses silently ignore non-`_eq` operators"; **broadened 2026-06-04** after a reviewer confirmed the `_in` gap also affects the `filter` path (the original write-up wrongly said `resolveFilters` honored all operators). Validated by code-read against `0f8826c`._

taitelee:
## Summary

`_in` was accepted by the policy `Filter` schema but never enforced: on the row-`filter`/SELECT path it produced **no `WHERE` predicate** (a fail-open — a `tenant_id: { _in: … }` filter let the role see *every* row instead of its tenant subset, same family as #223), and on the `check`/INSERT path only `_eq` was honored. This enforces it for real on both paths.

- **Filter:** `_in` takes a single claim that resolves to a JSON array (the multi-tenant case — a token's `tenant_ids` list) and emits `col IN (?, …)` with one bound param per element. A scalar claim is a one-element set; an empty/absent claim matches **no rows** (fail-closed) rather than widening.
- **Check:** an `_in` check requires the inserted column be present and one of the claim-derived set — no auto-inject (unlike `_eq`, there's no single value to stamp), so an omitted column is rejected. `check` now honors `_eq` + `_in`; `_neq`/`_gt`/`_lt` stay a loud config-load rejection (no insert-time meaning).
- Wire schema unchanged (`_in` stays a single templated string in Go + the SDK), matching the established "set = array" shape of the caller-query `in` operator. Supersedes the interim guard that merely rejected `_in` at `policy.Validate`.

## Related Issues
Closes #224


internal/policy/policy.go
```diff
@@ -254,6 +278,32 @@ func resolveTemplate(tmpl string, claims map[string]any) string {
...
+			for _, e := range v {
+				out = append(out, fmt.Sprint(e))
+			}
+			return out
+		default:
+			return []any{fmt.Sprint(v)}
+		}
+	}
+	return []any{resolveTemplate(tmpl, claims)}
+}
```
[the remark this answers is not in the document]
taitelee: Fail closed is right, but this path is already closed. `resolveInValues` returns `[]any`, so case `nil`: `return nil` is a typed-`nil []any`, not an untyped `nil`. It still asserts as `[]any` in `processRecord`, so an absent `_in` claim takes the membership branch and rejects (403), never the `_eq` auto-inject branch. Added `TestIngest_Policy_CheckIn_AbsentClaim_FailsClosed` to lock it in.

internal/policy/policy_test.go
```diff
@@ -639,6 +639,161 @@ func TestValidate_RejectsBindUnsafeFilterColumn(t *testing.T) {
...
+// widen to all rows — the fail-closed direction. `IN ()` is invalid SQL.
+func TestResolveFilters_InEmptyClaim_FailsClosed(t *testing.T) {
+	t.Parallel()
+	in := "{{ jwt.tenants }}"
+	filters := map[string]Filter{"tenant_id": {In: &in}}
+	clauses, params := resolveFilters(filters, nil)
+	require.Len(t, clauses, 1)
+	assert.Equal(t, "1 = 0", clauses[0])
+	assert.Empty(t, params)
+}
```
EricAndrechek: This does test absent, but if I'm not misunderstanding it doesn't test empty, right? Because this would hit the `nil` case/branch in the code, but for empty to be tested we'd need to give it an actual array just with no values, like `tenants: []any{}` right?

EricAndrechek (changes requested):
Would like to see that test case fixed up but other than that all good. Did have some thoughts on this policy stuff more broadly and SDK etc, but tracked in separate issue #371 to keep this one tighter

internal/policy/policy_test.go
taitelee: Fixed. The test is now table driven over both an absent claim (case nil) and a present but empty array tenants: []any{} (case []any, zero elements), so both paths are shown to fail closed to 1 = 0.

EricAndrechek (approved):

Issue #202, opened 2026-06-02
feat(sdk): auto-injected dedupe ID for retransmission detection


EricAndrechek:
Auto-inject a dedupe ID (not persisted into final DB row values) so network-issue retransmissions are caught automatically.

**Acceptance**
- SDK attaches a dedupe id per message
- API uses it for idempotency/dedupe without storing it in row values
- Test for a retransmit scenario

Part of #194

EricAndrechek:
**Cross-link from WaveHouse-Stats dogfooding:** this SDK auto-inject is the *client* half of dedupe-id robustness. Filed the *server*-side gaps it can't cover:
- **#219** — the server emits no log/metric when dedupe is enabled but a row lacks the id (silent loss of idempotency). This is the tripwire that would catch an SDK that forgets to inject.
- **#221** — let the dedupe id come from a request header (e.g. `X-GitHub-Delivery`) instead of only a body field.

Issue #222, opened 2026-06-04
Per-table dedupe id_field (+ fix cross-table dedupe keyspace collision)


EricAndrechek:
**Area:** dedupe — gap + latent correctness bug · found via WaveHouse-Stats dogfooding

**Expected:** each table dedupes on its own natural key, in its own keyspace.

**Actual:** two problems with the single global dedupe key:
1. **One global `id_field`** name applies deployment-wide (`internal/config/config.go:105-108`) — every table that wants dedupe must carry an identically-named column.
2. **Shared keyspace (latent bug):** `CheckAndMark` keys Pebble on the raw id value with **no table namespace** (`internal/dedupe/embedded.go:29`, `key := []byte(eventID)`). Equal id *values* across two different tables collide — the second is silently dropped as a "duplicate." Only bites when ids aren't globally unique across tables (e.g. per-table sequential ids), but it's silent data loss when it does.

**Impact (Stats):** we standardize on globally-unique `event_id`, so the collision doesn't bite us today — but it's a sharp edge for any multi-table deployment, and the single id_field forces a naming convention on every future table.

**Scope:** namespace the dedupe key by table (fixes #2 regardless), and allow a per-table id_field (fixes #1). The keyspace fix is the more urgent half.

Related: #219, #220, #221.

---
_From `WAVEHOUSE-FEEDBACK.md` (WaveHouse-Stats dogfooding); validated by code-read against `0f8826c` on 2026-06-04._

EricAndrechek:
Cross-linking #370 — it's a concrete instance of item 2 here (the shared, un-namespaced keyspace).

A present-but-`null` `id_field` dedupes on the literal string `"<nil>"` (`fmt.Sprint(nil)`), and because the key has no table prefix (`internal/dedupe/embedded.go:29`) that collision is global across every table — the same root cause as the cross-table keyspace bug in item 2. So namespacing the key by table fixes the cross-table dimension of #370 too; #370 additionally needs `null`/missing treated as "no id" in `processRecord`.

Flagging so the keyspace change is coordinated across both.

taitelee:
[...]

Issue #15, opened 2026-04-13
Dual-Mode Metric Export (OTLP Push vs. Prometheus Scrape)


taitelee:
## Problem

Different deployment environments have different requirements: SigNoz prefers OTLP Push, while existing enterprise environments often require a Prometheus /metrics scrape endpoint. We need the flexibility to support both without maintaining two separate instrumentation libraries.

## Proposed Solution

Use the go.opentelemetry.io/otel/exporters suite to provide a toggleable export layer:

- OTLP Push: The default behavior, pushing to WH_OTEL_ADDR.
- Prometheus Scrape: Use the OTel Prometheus Exporter (otelprometheus) to host a /metrics endpoint on port 8080.
- Logic: Use the config.yaml to determine which exporter to initialize. If WH_METRICS_EXPORT_TYPE=prometheus, start the HTTP server; otherwise, default to gRPC OTLP.

## Alternatives Considered

Using prometheus/client_golang directly: Rejected because it would require double-instrumenting the code. OTel's Prometheus exporter allows us to write code once and export it in any format.

## Additional Context

None

PR #364, opened 2026-07-01
feat(dedupe): observe or reject ingest rows missing the id_field

### LINKED ISSUE #219: Warn/metric (or strict mode) when dedupe is enabled but a row has no id_field
**Area:** ingest · dedupe — footgun (silent loss of a guarantee) · found via WaveHouse-Stats dogfooding

**Expected:** when dedupe is enabled, a row that can't be deduped (missing the configured id field) is at least observable.

**Actual:** if a row lacks the configured `id_field` it's published with no dedupe and **no log or metric** — there is no `else` branch on the id lookup (`internal/api/ingest.go:132-148`). A producer bug that drops the id silently disables idempotency for those rows.

**Impact (Stats):** the Worker/reconciler must *guarantee* `event_id` on every row, with no server-side tripwire if it ever doesn't.

**Scope:** a warn log + counter at minimum; optionally a strict mode that rejects rows missing the id when dedupe is on. The warn-vs-reject choice is the maintainer's.

Related: client-side mitigation is #202 (SDK auto-injects the id); this is the missing *server-side* signal.

---

taitelee:
## Summary

With dedupe enabled, `POST /v1/ingest` silently dropped its idempotency guarantee: a row missing the configured `id_field` was published un-deduped with **no log or metric** (`internal/api/ingest.go`), so a producer bug that dropped the id disabled dedupe for those rows unnoticed.

This makes that case observable and, optionally, enforceable:
- **Always-on:** a row missing `id_field` now emits a `WARN` and increments a new `wavehouse_ingest_dedupe_missing_id_total` counter (labeled by `table`).
- **Opt-in strict mode:** `dedupe.require_id` (`WH_DEDUPE_REQUIRE_ID`, default `false`) rejects such rows instead of publishing them — `400` for a single insert, a per-record failure in a batch. Default behavior is unchanged, so this is non-breaking.

Complements the client-side id auto-injection tracked in #202 with the server-side signal/guardrail that was missing.

## Related Issues

Closes #219
Relates to #202


internal/api/ingest_test.go
```diff
@@ -401,6 +402,58 @@ func TestIngest_Dedup_MissingIDField(t *testing.T) {
...
+	h := NewIngestHandler(testRegistry(), pub, testutil.NopLogger())
+	h.Dedup = testutil.NewMockDeduplicator()
+	h.IDField = "event_id"
+	h.RequireID = true
+
+	w := httptest.NewRecorder()
+	h.Handle(w, ingestRequest(t, "clicks", map[string]any{"page": "/home"}))
+	assert.Equal(t, http.StatusBadRequest, w.Code)
+	assert.Contains(t, w.Body.String(), "missing dedupe id field")
+	assert.Nil(t, pub.LastMessage(), "must not publish a row missing the dedupe id under require_id")
```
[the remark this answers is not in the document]
taitelee: Added `testutil.AssertJSONErrorResponse(t, w)` after the status/substring asserts in `TestIngest_Dedup_RequireID_Rejects`. Left the NDJSON batch test as-is: that path returns 200 with per-record errors in the results array rather than an HTTP error response, so the helper doesn't apply there.

internal/api/ingest.go
```diff
@@ -53,6 +55,11 @@ func NewIngestHandler(registry *discovery.SchemaRegistry, pub mq.Publisher, logg
 	return &IngestHandler{Registry: registry, Publisher: pub, logger: logger}
 }
 
+var dedupeMissingIDCounter, _ = otel.Meter("wavehouse-ingest").Int64Counter(
+	"wavehouse_ingest_dedupe_missing_id_total",
+	metric.WithDescription("Ingested records missing the configured dedupe id_field (idempotency skipped)"),
+)
+
```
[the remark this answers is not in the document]
taitelee: Keeping the _ here intentionally. It's the established pattern for instrument creation across the codebase — internal/stream/metrics.go (4 instruments) and internal/observability/metrics.go (4 gauges) all discard the creation error the same way — so this matches, rather than special-casing one counter.

On the specific risk: the OpenTelemetry Go API returns a usable no-op instrument alongside any error, so `dedupeMissingIDCounter.Add(...)` can't nil-panic even in the error case, and since this is a single package-level var created once there's no duplicate-name registration to collide with. `panic(err)` at `init` would also take the whole server down at startup over a metrics-registration hiccup, which runs against WaveHouse's fail-soft observability invariant (AGENTS.md Key Design Decision #15 — observability must never block startup; OTel exporters dial lazily for the same reason).

internal/api/ingest.go
```diff
@@ -382,7 +389,18 @@ func (h *IngestHandler) processRecord(
 
 	// Optional deduplication.
 	if h.Dedup != nil && h.IDField != "" {
-		if idVal, ok := data[h.IDField]; ok {
+		idVal, ok := data[h.IDField]
```
EricAndrechek, line 392: Out of scope for this PR, just flagging: this presence check treats an explicit `{"event_id": null}` as present, so a null id slips past `require_id` and then dedupes on the literal `"<nil>"` (via `fmt.Sprint` below) — every null-id row collides on one key. Pre-existing behaviour, but more relevant now that we reject missing ids. Worth a follow-up issue?

EricAndrechek (commented):
Clean change — non-breaking default, and both the observe and reject paths are covered. One doc-sync nit before merge:

The reference `.env` in `docs/src/content/docs/deployment.md` (the `# Optional dedup` block) still lists only `WH_DEDUPE_ENABLED` / `WH_DEDUPE_ID_FIELD`. Add `WH_DEDUPE_REQUIRE_ID=false` there too — otherwise an operator wiring up dedup from the deployment guide never discovers strict mode. `configuration.mdx`, `config.yaml`, and the compose block are all updated correctly, just this one got missed.

internal/api/ingest.go
EricAndrechek: Dug into this a bit more so it's grounded, not hand-waving:

**Reachability** — `discovery.Validate` only rejects a null when the column is non-nullable *and* has no default. If the `id_field` column is `Nullable(...)` or has a `DEFAULT`, `{"event_id": null}` passes validation and reaches this line with `ok == true`, so it falls into the `else` and dedupes on `fmt.Sprint(nil)` = the literal string `"<nil>"`. Config-dependent, but entirely reachable — every null row then collides on that one key and all but the first get dropped as duplicates.

**Why `null`, not empty string** — `null` is the *absence* of a value, semantically the same as the field being missing; `""` is a real value a client chose. And we're not even deduping on "null" here — we're deduping on the `"<nil>"` sentinel `fmt` happens to emit (which also collides with a client literally sending the string `"<nil>"`). So `null` provably shouldn't dedupe; `""` is at least debatable, so I'd leave it alone.

**Why it's arguably in scope for this PR** — it's a hole in `require_id`'s own guarantee: the check is key-presence only, so a null id sails past strict mode and the "every row has a usable id" invariant doesn't actually hold. Small fix — `if !ok || idVal == nil`, i.e. treat null exactly like a missing key (count it, reject under `require_id`). Scope it to null only. Note it's a real (small) behavior change — observe-mode would publish null rows un-deduped instead of collapsing them onto `"<nil>"` — so it wants its own test + a CHANGELOG line.

Happy to punt to a follow-up if you'd rather keep this PR tight — your call. Just wanted the full picture on the thread.

EricAndrechek (changes requested):
Requesting one change before merge.

The reference `.env` in `docs/src/content/docs/deployment.md` (the `# Optional dedup` block) still lists only `WH_DEDUPE_ENABLED` / `WH_DEDUPE_ID_FIELD`. Add `WH_DEDUPE_REQUIRE_ID=false` there per AGENTS.md §Documentation Sync — otherwise an operator wiring up dedup from the deployment guide never discovers strict mode. `configuration.mdx`, `config.yaml`, and the compose block are all updated correctly; just this one env reference got missed.

One line, e.g.:

```dotenv
WH_DEDUPE_REQUIRE_ID=false        # reject rows missing the id field instead of publishing un-deduped
```

Everything else looks good — non-breaking default, both observe and reject paths tested. The null-id gap is tracked separately in #370, so it's not blocking here.

internal/api/ingest.go
taitelee: Brought up in #370.

EricAndrechek (approved):
Approved, but I think we should consider moving the global strict mode config stuff to per-table config stuff when #222 is addressed too.

PR #368, opened 2026-07-03
feat(docs): override Starlight header with a top-level site nav

taitelee:
## Summary

Adds a curated top-level Docs link to the docs site header, with a custom header layout for the sidebar-less splash homepage. Starlight has no config hook for header links, so this overrides its Header component through the sanctioned
components map. The override is our own layout — brand + primary nav pinned left, controls pinned right — rather than a copy of Starlight's default row: doc pages stay full-bleed to the edges, while the splash homepage (the only page with a hero) holds the inner content in a centered 68rem column — deliberately narrower than the hero's 80rem band.

Every control (SiteTitle, Search, ThemeSelect, SocialIcons, LanguageSelect) is still Starlight's own built-in, re-rendered through its virtual:starlight/components/* import, so their overrides and Pagefind search keep working untouched. Search keeps Starlight's expanded "Search … ⌘K" box on every page at ≥50rem (including the splash) so the keyboard shortcut stays discoverable; below 50rem — where the nav and controls are hidden (doc pages' sidebar drawer carries navigation there; the splash's hero CTAs do) — it collapses to its icon.

The single Docs link (→ /getting-started) lives in headerNav in sidebar.ts, colocated with the sidebar so the header and sidebar share one navigation source of truth. It's the way into the doc tree from the splash homepage, which hides the sidebar; every doc page already carries the full sidebar, so the nav is hidden below the md breakpoint (50rem) there. The footer brand block reuses the same <Logo/> (labelled via a title prop mirroring WaveMark's API) so header and footer render one wordmark; the inline lockup is decorative inside the labelled home link, SSR and client normalize nav hrefs identically, and headerNav documents that component-rendered links sit outside starlight-links-validator's scope. env.d.ts gains type declarations for the built-in components the override re-renders so astro check passes.

## Related Issues

None


EricAndrechek:
The `docs` label in the header isn't immediately clear to me that it's a link, and it also isn't vertically centered well against the wavehouse logo: 
image

Also, in the old version, 
image
the search bar shows large and that I can keyboard shortcut to access it quickly, which now isn't as clear 
image

Also, with the header narrower now it introduces a new little UI thing: when it was full width, switching pages to the docs meant the header didn't need to re-render or change, it was the same width and stuff, but now the width changes which looks strange, but at the same time I agree the narrower looks nice on the landing page, so unsure how we should do it.

EricAndrechek:
I like the changes to the search bar, but doc link still isn't aligned quite right on my view: 
image

Is it that the like WaveHouse text in the WaveHouse logo isn't centered properly that makes the docs part look not centered?

taitelee:
Looking into why there are some inconsistencies with the header being displayed because on my end that's been fixed for quite some commits.

docs/src/components/Header.astro
```diff
@@ -0,0 +1,209 @@
...
+  .wh-lead {
+    display: flex;
+    /* Center the lockup vertically in the bar, matching the right-hand controls
+       (search, social, theme) — which are all centered too. The "Docs" link is
+       then dropped onto the lockup's wordmark baseline (see .wh-header-nav__link);
+       it can't be flex-baseline-aligned to a wordmark baked inside an image. */
+    align-items: center;
+    gap: 1.75rem;
+    min-width: 0;
+  }
```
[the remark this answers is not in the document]
taitelee: Removed the offset entirely instead of hoisting it to a variable. The wordmark's cap-height center sits at ≈y49 in the lockup's 0 0 508 100 viewBox — essentially the image center — so I switched the link to plain vertical centering against the lockup. It lands on the wordmark's optical center with no tuned constant, and it's stable across displays (the old 1px nudge was tuned to one screen's rasterization).

docs/src/components/SiteTitle.astro
```diff
@@ -1,51 +1,62 @@
...
+    brand. Two pre-themed exports from public/branding, swapped by CSS on the
+    site's `data-theme` (which defaults to dark); below 30rem it collapses to the
+    compact mark, like the old text-hidden state. alt="" because the link's
+    aria-label already names it — the images are decorative.
+  */}
+  <img class="wh-logo wh-logo--lockup wh-logo--on-dark" src={`${base}branding/lockup-dark.svg`} alt="" width="508" height="100" />
+  <img class="wh-logo wh-logo--lockup wh-logo--on-light" src={`${base}branding/lockup-light.svg`} alt="" width="508" height="100" />
+  <img class="wh-logo wh-logo--mark wh-logo--on-dark" src={`${base}branding/mark-dark.svg`} alt="" width="100" height="100" />
+  <img class="wh-logo wh-logo--mark wh-logo--on-light" src={`${base}branding/mark-light.svg`} alt="" width="100" height="100" />
 </a>
```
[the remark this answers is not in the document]
taitelee: Accepted tradeoff, agreed. The theme is a runtime data-theme toggle rather than a media query, so the dark/light pair can't be selected via <picture media> at request time without extra JS — and these are four tiny static SVGs. Keeping the CSS swap for simplicity and pixel-identical branding.

EricAndrechek:
Would love to see some changes, including:
- [x] Abstracting out the logo stuff into its own component (do we still use the `WordMark` or whatever it was one, or can we replace it?) for easier re-use elsewhere outside the header (ie anywhere we need to load the logo and have it properly show based on theme preferences) and to help make the header component more concise.
- [x] Removing magic numbers across the board, especially in aligning the docs/header links to the logo, so that future changes to fonts, font sizes, logo sizes/fonts, etc. don't break the alignment and the styles become easier to maintain
- [x] Figure out how to make the header width change from narrow on the homescreen to wide on the docs pages smoother (the logo moves from indented to left-aligned, which is abrupt when switching to docs pages compared to how it was before always left-aligned)

docs/src/assets/branding/lockup-themed.svg
EricAndrechek, line 1: This doesn't render in Github's preview, and locally when I try to open it I see: 

Image

I think it should (hopefully) be just as easy as removing the comment at the top?

docs/scripts/branding/generate.sh
```diff
@@ -202,6 +203,16 @@ cat > "$OUT_KIT/lockup-light.svg" <<EOF
 EOF
 outline_lockup_svg "$OUT_KIT/lockup-light.svg"
 
+# Themeable inline lockup for in-app components (Logo.astro inlines it via
+# set:html). Same OUTLINED art as lockup-dark, but colours swapped for CSS hooks
+# so ONE asset themes itself with no light/dark variants: mark → currentColor
+# (the consumer sets color: var(--wh-accent)), wordmark → var(--wh-ink). The
+# substitution is case-insensitive because usvg may lower-case the hex on outline.
+{
+  printf '%s\n' '<!-- Themeable inline lockup: mark=currentColor (set color:var(--wh-accent)), wordmark=var(--wh-ink). GENERATED from lockup-dark.svg by generate.sh — do not hand-edit. -->'
```
EricAndrechek: See comment on the `lockup-themed.svg` file – I think we need to remove the comment

docs/src/components/Logo.astro
EricAndrechek, line 1: I like the idea here, but I think it needs some tweaking and you testing it on your phone and whatnot. On my phone I see this: 

Image

which is not great imo, like we definitely could fit the words on that bottom logo, and should probably put more in the top header bar, whether it be the logo words too or the docs link or an expanded search button, or the theme/language toggles, etc.

EricAndrechek (changes requested):

EricAndrechek (approved):

Issue #294, opened 2026-06-08
epic(streaming): SSE delivery-path throughput — project/serialize once per role, not per subscriber


EricAndrechek:
**Area:** streaming · perf — throughput-ceiling epic · found via WaveHouse-Stats dogfooding (load tests in `docs/load-testing/`)

The live SSE fan-out re-does per-subscriber what could be done once per (role, table). Measured ceiling ≈ **2 270 deliveries/sec**, flat across event size — caps the public dashboard at ~`2270 / event_rate` clean concurrent viewers (~110 during a ~20 ev/s CI burst). Fine at today's traffic; the #1 lever before the streaming embed (LiveDemo) scales. Grouping the delivery-path optimizations so they're tracked without backlog spam (the #194 / #228 pattern); split into discrete issues when promoted to Ready.

**Root cause (shared):** projection + serialization live in the per-client read loop (`internal/api/stream.go:116-145`), not in the shared broadcast (`internal/api/hub.go:62-92`, which already marshals the envelope once and non-blocking-sends to N channels).

- [x] **Project + serialize once per (role, table, column-set), not per subscriber** — the big lever. `applyStreamPolicy` (`stream.go:136,162-193`) runs `json.Unmarshal` + `policy.Evaluate` + `filterEventColumns` (fresh map alloc) + `json.Marshal` per subscriber, though the result is byte-identical for a given (role, table). For a single-role audience (the public dashboard, every viewer `public`) this collapses N re-projections to 1, moving the ceiling from deliveries/s toward events/s. Memoize inside `Broadcast`, or pre-project in the `hub-bridge` handler keyed by role.
- [x] **Don't start an OTel span per event per subscriber on the delivery path** (done in #346 — per-delivery `SSE.PushEvent` span removed from the fan-out path) — `stream.go:129-143` runs `propagator.Extract(...)` + `tracer.Start("SSE.PushEvent")` … `pushSpan.End()` per fan-out leg (N×R spans/s), and the `Extract`/`Start` calls execute even with `WH_OTEL_ENABLED=false`. Trace once at ingest/broadcast, or gate the per-delivery span behind a sampling/debug flag.
[...]

Issue #239, opened 2026-06-04
feat(stream): re-validate JWT / enforce token expiry on long-lived SSE connections


EricAndrechek:
**Area:** streaming · auth — security gap · from Eric's notes

**Expected:** a long-lived SSE connection reflects the token's lifetime and current permissions — an expired/revoked token, or a policy change, eventually stops or re-authorizes the stream.

**Actual:** the JWT is validated once when the stream opens; after that the connection serves indefinitely. There's no periodic re-validation, no mid-stream `exp` enforcement, and a policy change isn't re-evaluated for an already-open subscription. A token that expires (or is revoked) an hour into a stream keeps delivering.

**Scope:** periodic re-check of the token (expiry + signature) and/or policy re-evaluation on a timer for open SSE connections; close or downgrade the stream when it no longer authorizes.

Related: auth-hardening epic #228 (exp, revocation); SSE keepalive #226 (same handler).

---
_From Eric's notes scratchpad; triaged 2026-06-04._

EricAndrechek:
Confirming this by code-read against current `main`, with the exact location in case it helps whoever picks it up.

`internal/api/stream.go` resolves the caller's role **once**, at connection setup:

```go
role := auth.RoleFromContext(r.Context())   // stream.go:48
...
h.Hub.Add(topic, role, sub)                 // stream.go:82
```

and then enters the `for { select { … } }` delivery loop, which never revisits it. The connection is pinned to whatever role and token state existed at `onopen`, for as long as the client holds the socket open — no `exp` check, no re-validation, no policy re-read.

**One consequence worth folding into the scope:** this defeats short-TTL tokens as a revocation strategy. A deployment that deliberately mints short-lived tokens — so that the revocation window is bounded by the TTL — gets that property on queries and **not** on streams. A single long-lived SSE connection outlives every token rotation, indefinitely. That's a surprising asymmetry if you've reasoned about your exposure window from the token lifetime, and it's invisible from the outside.

**Cheap first step:** even without timer-based re-validation, a configurable **maximum connection lifetime** — close the stream, let the client's `EventSource` auto-reconnect, and re-run the full auth path on reconnect — would bound the exposure with very little machinery, and it reuses the reconnect behavior that already exists for slow-consumer disconnects. The full re-validate-in-place version could follow without the interim being wasted work.

Related: #427 (introspection mode) would make this sharper still — a stream that never re-checks would silently ignore a revocation the introspection endpoint had already reported.

PR #381, opened 2026-07-08
fix(stream): apply policy row-filter per subscriber on SSE

### LINKED ISSUE #319: security(streaming): SSE applies the column allowlist but not the policy row-filter — query/stream RLS drift
**Area:** api · policy · observability (streaming) — security (RLS bypass on the stream path) · found via pre-launch audit

**Expected:** the SSE delivery path applies the **same** policy as the structured-query path — both the column allowlist **and** the row-filter (RLS predicate) that the query path injects.

**Actual:** the SSE stream applies the column allowlist but **not** the policy row-filter, so the two read paths drift: a subscriber receives rows that the structured-query path would filter out for that role.

**Impact:** any deployment that combines row-policies with a public or role-scoped SSE stream leaks rows over the stream that RLS should remove — a data-exposure on the streaming surface. Live signal on Stats is limited (the public table carries no restrictive row-policy), but the gap is real for any future private/PII table fronted by a stream.

Scope: apply the query-path row-filter inside the stream policy evaluation (`applyStreamPolicy`).

Related: #294 (SSE delivery path), #214 (per-table policy scopes), the `applyStreamPolicy` unfiltered-passthrough issue.

---
_From WaveHouse-Stats pre-launch security audit (`WAVEHOUSE-FEEDBACK.md` dogfooding), audited dev `60fed15` (2026-06-10). Filed via /pm-triage._

taitelee:
## Summary

The Server-Sent Events stream applied a role's column allowlist but never its row-level `filter`, so a subscriber could receive rows the structured-query path would have hidden for that same role — a row-level-security bypass on the streaming surface. This applies the same row-filter on the stream:

- The filter resolves **once** into predicates shared by both read paths — rendered to SQL on the query path, evaluated in memory (`ResolvedPermissions.RowVisible`) on the stream — so the two can't drift.
- Fan-out is now claims-aware: a role **without** a filter keeps the once-per-role projection fast path unchanged (zero change for the public stream); a **filtered** role keeps the shared column projection but delivers each row only to the subscribers whose JWT claims admit it (evaluated against the full event). Replay does the same per connection.
- Ordering predicates (`_gt`/`_lt`) are schema-informed — numeric columns compare numerically via the schema registry now wired into the hub — while equality/set (`_eq`/`_neq`/`_in`) stay exact. Ambiguous values fail closed; the best-effort ordering boundary is documented in access-control docs.
## Related Issues

Closes #319


internal/stream/hub_test.go
```diff
@@ -148,9 +160,132 @@ func TestHub_ProjectsPerRole_DistinctRolesGetDistinctFrames(t *testing.T) {
...
+
+	sub := NewSubscriber() // constant filter value ⇒ no claims needed
+	hub.Add(topic, "viewer", sub)
+
+	hub.Broadcast(topic, rawEvent(t, "clicks", "t1", map[string]any{"amount": float64(9), "page": "/a"}))
+	assertNoFrame(t, sub) // 9 is not numerically > 100
+
+	hub.Broadcast(topic, rawEvent(t, "clicks", "t2", map[string]any{"amount": float64(250), "page": "/b"}))
+	assert.Equal(t, float64(250), frameData(t, recvFrame(t, sub))["data"].(map[string]any)["amount"])
+}
```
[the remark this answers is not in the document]
taitelee: They compile. Go 1.26 generalizes the build in new operator to accept a value expression, not only a type.

internal/stream/hub.go
```diff
@@ -125,20 +133,70 @@ func (h *Hub) Broadcast(topic string, raw []byte) {
...
+			subPerms := policy.Evaluate(p, rb.role, evt.TableName, "select", sub.claims)
+			if !subPerms.RowVisible(evt.Data, numericCols) {
+				continue // this row is filtered out for this subscriber
+			}
 			if !sub.Send(frame) {
 				h.metric.FrameDropped(KindEvent)
 			}
 		}
 	}
 }
```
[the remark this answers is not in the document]
taitelee: Agreed the observation is correct, and we're keeping it as-is for now. The claims-independent fast path (`!HasRowFilter()`) means the common, high-fanout public stream pays nothing — the per-subscriber Evaluate only runs on topics that actually carry a row-filter, where visibility is inherently per-connection (different JWT claims → different rows), so the decision can't be hoisted out of the loop without losing correctness. Memoizing per distinct claim set would rarely hit (subscribers sharing a row-filtered bucket typically have distinct tenant claims — that's the point of the filter) and map[string]any claims aren't cheaply hashable. Correctness is unaffected either way, so we'd rather add a benchmark and optimize if profiling on a real filtered-high-fanout topic shows it matters, than add the complexity speculatively.

internal/stream/hub_test.go
```diff
@@ -287,7 +422,7 @@ func TestHub_ReplayFrame(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-			f, ok := hub.ReplayFrame(tt.role, raw)
+			f, ok := hub.ReplayFrame(tt.role, nil, raw)
```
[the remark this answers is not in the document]
taitelee: Added `TestHub_ReplayFrame_RowFilter` with matching-claims (row replayed, projected to allowed columns, denied column stripped) and non-matching-claims (row withheld) subtests. Also dropped the `const topic` from the suggested snippet since `ReplayFrame` takes role/claims/raw directly and doesn't use a topic.

internal/stream/hub_test.go
```diff
@@ -343,6 +506,44 @@ func TestWireFrame(t *testing.T) {
...
+	const topic = "ingest.clicks"
+	raw := rawEvent(b, "clicks", "2026-06-26T00:00:00Z",
+		map[string]any{"tenant_id": "acme", "page": "/a", "secret": "x"})
+
+	for _, n := range []int{100, 1_000, 10_000} {
+		b.Run(fmt.Sprintf("subscribers=%d", n), func(b *testing.B) {
+			hub := NewHub(policy.NewMemoryStore(rowFilterPolicy()), nil, nil)
+			// Half the subscribers share the event's tenant (row visible), half don't
+			// (row withheld); either way each still pays the per-subscriber Evaluate,
+			// which is the cost under measurement.
```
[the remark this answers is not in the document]
taitelee: Fixed. The comment predated the compiled refactor; there's no per-subscriber Evaluate anymore. Updated it to "each pays the per-subscriber `RowVisible` check against the once-per-bucket compiled filter, which is the cost under measurement," which matches the benchmark header. The half-visible/half-withheld split stays — it's there to exercise both `RowVisible` branches.

internal/policy/rowfilter.go
```diff
@@ -0,0 +1,149 @@
...
+	if p == nil {
+		return true
+	}
+	// A denied role sees no rows — fail closed, mirroring the !Allowed guard on
+	// IsColumnAllowed, so a denied receiver never reads as "no filter ⇒ all visible".
+	if !p.Allowed {
+		return false
+	}
+	for _, pred := range p.rowFilter {
+		if !pred.matches(row, numericCols[pred.Column]) {
```
EricAndrechek: `numericCols[pred.Column]` collapses three distinct states into `false`: a known text column, a column absent from the schema, and *no schema at all* — `Hub.numericCols` returns nil when the registry is nil or when `registry.Get(table)` misses. The first is correct. The other two silently downgrade `_gt`/`_lt` to lexicographic comparison, which admits rows ClickHouse would exclude: `"9" > "100"` is true as text, false as numbers.

This contradicts the contract stated at lines 31–33 of this file — "Every ambiguous or uncomparable case fails closed — the row is hidden, never leaked" — and `rowfilter_test.go:64` currently pins the leak as expected behavior.

The no-schema state is reachable in production, not just in the nil-registry test hub: `cmd/wavehouse/main.go:198` treats a failed boot-time `registry.Refresh` as non-fatal and serves while retrying in the background, so every table reads as unknown for that window. A table dropped or renamed since the replayed events were published lands in the same place.

internal/stream/hub_test.go
```diff
@@ -300,9 +435,35 @@ func TestHub_ReplayFrame(t *testing.T) {
...
+		t.Parallel()
+		_, ok := hub.ReplayFrame("viewer", map[string]any{"tenant": "globex"}, raw)
+		require.False(t, ok, "row must be withheld when claims don't satisfy the filter")
+	})
+}
+
 func TestHub_ConcurrentAddRemoveBroadcast_Race(t *testing.T) {
 	t.Parallel()
-	hub := NewHub(nil, nil)
+	hub := NewHub(nil, nil, nil)
```
EricAndrechek, line 657: This is the only test that drives the hub from multiple goroutines, and it builds a passthrough hub. With a nil policy store `projectColumns` returns a nil `perms`, so `HasRowFilter()` is false and the new per-subscriber `sub.claims` read at `hub.go:169` never executes here.

The tests that do reach that line (`TestHub_RowFilter_*`) call `SetClaims`, `Add` and `Broadcast` from a single goroutine, so the cross-goroutine read this PR introduces is never observed by the race detector anywhere in the suite. `-race` covers unit and integration only (`Makefile:666`, `:680`), and the e2e binary is built `-cover -covermode=atomic` under `CGO_ENABLED=0` (`scripts/build.sh`), so the new e2e row-filter coverage doesn't close the gap either.

docs/src/content/docs/access-control.mdx
```diff
@@ -370,12 +370,12 @@ The same policy drives every data path, but not every field is meaningful on eve
...
...row-level `filter` predicates admit, evaluated per subscriber against their JWT claims. Equality and set predicates (`_eq`, `_neq`, `_in`) — the usual tenant/user scoping — are enforced exactly: numeric ties resolve at full precision (an ID beyond `float64`'s 2^53 never falsely matches a neighbor) and an unparseable or `NaN` operand withholds the row. One representation caveat: a `Bool` (or `DateTime`) value compares by its canonical text form, without ClickHouse's cross-representation coercion — write the filter value the way your events carry it (`true`, not `1`). Ordering predicates (`_gt`, `_lt`) are best-effort: the stream compares numeric columns numerically (matching ClickHouse) but lacks ClickHouse's full per-type coercion, so an ordering filter on an exotic type (a `Decimal`/`Int128` beyond `float64` precision, or a `DateTime` ingested as a Unix number) can admit or withhold a row the query path wouldn't. The resource limits (`max_rows`, `max_execution_time`, `max_rows_to_read`, `max_memory_usage`) remain a property of the SQL query path and are **not** applied to the live event stream; if those caps are part of a role's isolation story, don't rely on them over the stream.
```
EricAndrechek: Two problems with the representation caveat.

**Scope.** Numeric comparison is selected only when `discovery.IsNumericType` is true, and that covers `Int*`/`UInt*`/`Float*`/`Decimal*` only (`internal/discovery/validation.go:161`). So *every* other type compares as text — `Enum`, `UUID`, `IPv4`/`IPv6`, `Date`, `String` — not just `Bool` and `DateTime` as written here.

**Direction.** The failure isn't uniformly fail-closed. A representation mismatch withholds the row under `_eq`/`_in`, but under `_neq` the text compare returns "not equal" and *admits* a row the query path would have excluded — an uppercase UUID carried in a claim, or `1` for a `Bool`.

docs/src/content/docs/architecture.md
```diff
@@ -85,7 +85,7 @@ The API layer uses [Chi](https://github.com/go-chi/chi) for routing with Request
...
...ach event once, applies each subscribed role's column policy once, builds one SSE frame per role, and fans it to every member of that role's `Bucket` — collapsing the prior per-subscriber `unmarshal → evaluate → filter → marshal` into one pass per distinct `(role, table)` output shape (the [#294](https://github.com/Wave-RF/WaveHouse/issues/294) lever; the measured ceiling was ~2 270 deliveries/s from re-projecting per subscriber). The column projection is claims-independent, so it is shared across a role's whole bucket; the role's row-level `filter` predicate is not — it is resolved against each subscriber's JWT claims, so for a role that carries a filter `Broadcast` keeps the shared column projection but delivers it only to the subscribers whose claims admit each row (`ResolvedPermissions.RowVisible`, evaluated against the full event via the numeric-aware comparison seeded from the schema registry). This is the [#319](https://github.com/Wave-RF/WaveHouse/issues/319) fix that closes the query/stream row-level-security drift; roles without a filter keep the pure once-per-role fast path. `ReplayFrame` shares the same projection and per-connection row check for the handler's gap-fill.
```
EricAndrechek: This bullet is updated, but the Streaming Path flow block at lines 261–263 still reads "Per-role policy filtering (historical + live): denied tables skipped, denied columns stripped. Live projection runs once per role (`Hub.Broadcast`)". It doesn't mention row filtering, and "once per role" now holds only for roles without a filter. The two descriptions in this file disagree.

docs/src/content/docs/access-control.mdx
```diff
@@ -370,12 +370,12 @@ The same policy drives every data path, but not every field is meaningful on eve
...
 :::
```
EricAndrechek, line 405: The caution doesn't cover the fail-closed case a policy author is most likely to hit. `rowfilter.go:56-58` withholds the row whenever the filter's column is absent from the event payload, and the stream evaluates against the ingested JSON rather than the stored row. So a filter keyed on a column clients don't send — or on a ClickHouse `DEFAULT`/`MATERIALIZED`/`ALIAS` column, which never appears in an SSE payload at all — silently withholds every event for that subscriber while the query path returns those rows normally.

internal/stream/subscriber.go
```diff
@@ -41,6 +48,12 @@ func newSubscriber(size int) *Subscriber {
 	return &Subscriber{out: make(chan Frame, size), evict: make(chan struct{})}
 }
 
+// SetClaims attaches this connection's JWT claims, which the Hub evaluates against a
+// role's row-level-security filter to decide, per subscriber, whether each event row
+// is visible. Call it before registering the subscriber with the Hub; the claims are
+// then read-only for the subscriber's lifetime, so the fan-out reads them race-free.
+func (s *Subscriber) SetClaims(claims map[string]any) { s.claims = claims }
```
EricAndrechek: Claims should be taken at construction and immutable, rather than settable after the fact.

Worth being explicit about why the hot-reload work we're moving toward doesn't argue for a setter here. What a policy/role reload changes is the **policy**, and that path is already fully dynamic and completely independent of this field: `Broadcast` calls `h.policy.Get()` on every event (`hub.go:134`), and `Store` *replaces* the whole `*Policy` on every update (`store.go:109`, `store.go:158`) rather than mutating in place. Reload a policy and the very next event picks it up for every subscriber, with no change to any `Subscriber`. Claims are a property of the authenticated connection — a policy reload changes what a caller may see, not who they are.

Meanwhile the field is read from the fan-out goroutine at `hub.go:169` with no synchronization. That's safe today only because `internal/api/stream.go:85` calls `SetClaims` before `Add` on line 86, and the happens-before edge comes from `subscriberSet`'s mutex inside `Add` — nothing in this file establishes it. The doc comment above states the consequence ("read-only for the subscriber's lifetime, so the fan-out reads them race-free") rather than the mechanism that guarantees it, so a future caller that sets claims after `Add` introduces a real data race on a security decision, with nothing at compile time or in the test suite to catch it (see my note on `hub_test.go:466`).

Constructor injection makes the invariant structural instead of documentary, and costs two mechanical call-site changes.

(Scoping out mid-stream re-auth / credential revocation on a long-lived SSE connection — tracked separately, wider blast radius.)

internal/stream/hub.go
```diff
@@ -125,20 +133,70 @@ func (h *Hub) Broadcast(topic string, raw []byte) {
...
+		// Row-level security applies. The column projection is claims-independent and
+		// shared, but whether each subscriber may see THIS row depends on its claims, so
+		// evaluate visibility per subscriber. Predicates read the full event data (a
+		// filter may key on a column the role can't SELECT), not the projected columns.
+		if !numericResolved {
+			numericCols = h.numericCols(evt.TableName)
+			numericResolved = true
+		}
 		for _, sub := range rb.bucket.Snapshot() {
+			subPerms := policy.Evaluate(p, rb.role, evt.TableName, "select", sub.claims)
```
EricAndrechek: Filed as #435 — not asking for it in this PR, but flagging it here so the cost is recorded rather than silently absorbed.

Nothing in this `Evaluate` call depends on the event. `role` and `table` are fixed for the connection (one subscriber = one topic = one table), `claims` are fixed at connect, and `p` changes only on policy reload — yet it runs once per event per subscriber. A subscriber connected across 10k events recomputes an identical `ResolvedPermissions` 10k times, each one paying a `resolveTemplate` regexp pass per filter value plus a `predicatesToSQL` render whose `WhereClause`/`WhereParams` the stream never reads.

Worth being explicit in the PR description or CHANGELOG that this trades some of the #294/#353 per-role fan-out gain for the RLS fix — right now the tradeoff isn't recorded anywhere, and `af7afb9`/`8987480` (which had solved it) were reverted in `2379bc4` with an empty commit body.

internal/stream/hub.go
```diff
@@ -125,20 +133,70 @@ func (h *Hub) Broadcast(topic string, raw []byte) {
...
+			if !subPerms.RowVisible(evt.Data, numericCols) {
+				continue // this row is filtered out for this subscriber
```
EricAndrechek: A row withheld here is invisible to operators. `Metrics` exposes only `ConnOpened` / `ConnClosed` / `FrameSent` / `FrameDropped` (`internal/stream/metrics.go`), and `FrameDropped` means a full queue — an RLS withhold isn't counted anywhere.

That matters more than it would for an ordinary filter, because this path fails closed silently in several distinct ways: a filter column absent from the event payload (`rowfilter.go:56`), a non-scalar value, an unparseable or `NaN` operand, and — today — any table the registry has no schema for. All of them produce the same outcome: the subscriber quietly receives nothing.

An operator debugging "this subscriber isn't getting events" has no way to separate *no matching rows* from *every row withheld by a misconfigured filter*, and no signal at all when a policy edit starts withholding everything for a role. A counter here (ideally labelled by table/role) would make a fail-closed security feature observable.

internal/stream/hub.go
```diff
@@ -149,50 +207,60 @@ func (h *Hub) snapshotPolicy() (p *policy.Policy, filter bool) {
...
 	decoded := json.Unmarshal(raw, &evt) == nil && evt.TableName != ""
 	p, filter := h.snapshotPolicy()
-	wire, ok := project(p, filter, role, &evt, raw, decoded)
+	wire, perms, ok := projectColumns(p, filter, role, &evt, raw, decoded)
 	if !ok {
 		return Frame{}, false
 	}
+	if perms.HasRowFilter() {
+		subPerms := policy.Evaluate(p, role, evt.TableName, "select", claims)
+		if !subPerms.RowVisible(evt.Data, h.numericCols(evt.TableName)) {
```
EricAndrechek: `h.numericCols(evt.TableName)` runs once per replayed event. `replayFromNATS` loops (`internal/api/stream.go:168`), so a gap-fill of N events does N registry lookups plus N `make(map[string]bool, len(schema.Columns))` allocations, each walking the full column list.

`Broadcast` hoists exactly this behind `numericResolved` (lines 164-167) and reuses it across every role and subscriber for the event. Replay doesn't, even though a `Last-Event-ID` reconnect after an outage is precisely the case where N gets large. Low stakes — replay is one-time and bounded — but the pattern is already sitting next door in the same file.

internal/stream/hub.go
```diff
@@ -5,25 +5,29 @@ import (
...
-// transform is column filtering, which depends solely on the role+table policy
-// entry — never on JWT claims (claims feed only row-level WHERE/CHECK, which the
-// stream path does not apply). See projectData.
+// Row-level security is the exception: a role's row-filter (RLS predicate) is
+// resolved against each subscriber's claims, so two subscribers of the same role can
+// be entitled to different rows. For a role that carries a row-filter, Broadcast
+// therefore keeps the shared column projection but evaluates row visibility PER
+// subscriber (ResolvedPermissions.RowVisible) before delivering — closing the
+// query/stream RLS drift in #319. Roles without a row-filter keep the pure
+// once-per-role fast path unchanged. See projectColumns.
```
EricAndrechek, line 25: This doc comment is updated, but the AGENTS.md invariant index isn't — and per AGENTS.md §Documentation Sync an architecture change maps to both `architecture.md` and `AGENTS.md`. (Commenting here because AGENTS.md isn't in this diff, so it can't be line-anchored.)

`AGENTS.md:44` still describes the package as "`Broadcast` projects + serializes each event once per role, the #294 delivery hot path" — now true only for roles *without* a filter.

`AGENTS.md:61` (Key Design Decision #12) closes with "Structured and live-stream (`stream.filterColumns`) reads share the one per-column decision `policy.IsColumnAllowed`, so column visibility can't drift." This PR builds the row-level twin of exactly that — one `resolvePredicates` call feeding both `predicatesToSQL` and `RowVisible` — and it isn't recorded anywhere in the index. That index is what the next agent touching `internal/stream` or `internal/policy` reads to learn what has to stay true, and "row visibility can't drift either" is precisely the kind of invariant it exists to hold.

internal/stream/hub_test.go
EricAndrechek, line 1: `BenchmarkBroadcast_RowFilteredFanout` existed at `8987480` and was removed by `2379bc4` along with the optimization it measured. HEAD has no benchmarks in this package at all.

This connects to how the performance thread on `hub.go:178` was closed. The reply there was: *"we'd rather add a benchmark and optimize if profiling on a real filtered-high-fanout topic shows it matters, than add the complexity speculatively"* — and CodeRabbit recorded a persistent learning from that exchange, so the flag is unlikely to resurface on its own in future reviews.

The branch's final state has neither the optimization nor the benchmark, so the condition that thread was resolved on can't actually be evaluated by anyone. Restoring just the benchmark — without the compiled-filter machinery that was reverted — would settle the question on data and guard the path against future regressions. See #435.

EricAndrechek:
Two doc-sync findings that land on lines this PR doesn't touch, so they can't be line-anchored as review comments.

**`docs/src/content/docs/api.md:542`** — the `GET /v1/stream` note still reads:

> **Note:** When access control policies are active, streamed events are filtered per the caller's role — denied columns are removed and tables without select permission are skipped.

That enumeration is now incomplete in a security-relevant way. `Hub.Broadcast` calls `policy.Evaluate(...)` + `RowVisible(evt.Data, numericCols)` per subscriber whenever `perms.HasRowFilter()`, and `ReplayFrame` does the same per connection for gap-fill — so a reader of the endpoint reference alone would conclude rows are *not* filtered, which is exactly what `access-control.mdx` said before this PR. AGENTS.md §Documentation Sync maps "add/modify API endpoint" to `api.md`, and the file isn't in this diff at all.

Worth noting here too that the stream now depends on claims, so a connection authenticated via `?token=` supplies the claims the filter binds against.

**`docs/src/content/docs/access-control.mdx:181`** — the canonical section for the feature still describes `filter` as a SQL-only mechanism:

> `filter` restricts *which rows* a role can read by injecting a `WHERE` clause into the generated SQL.

The column section immediately above it (line 177) already names both read surfaces — "the structured-query and live-stream paths defer to the **same** per-column decision" — and after this PR the row filter has the identical property: one `resolvePredicates` call in `Evaluate` feeds both `predicatesToSQL` (query) and `RowVisible` (stream). The row-filter paragraph should say so, rather than leaving the enforcement caution further down the page as the only place it's mentioned.

EricAndrechek (changes requested):
Nice work. Reviewing this PR took a lot longer than I expected/wanted it to – it really made me realize how complex our policy engine and stuff is starting to become... I originally had a BUNCH of things I was adding to this PR review about policy stuff I didn't like, but I think it's more important to get this security fixing PR merged and then to refactor and make policy more maintainable separately. Also some efficiency/perf type things, especially in the claim evaluation predicate stuff, SQL building, etc – like I noticed in this review just how much of that is wasteful recomputation and could be memoized or smth, and how much happens on the hot path, etc., but same idea: let's tackle that later in another PR so this one can just get shipped. I also forgot about our discussion about the row policy logic and how it being done correctly would technically require a CH SQL engine executing, which is what the query path does, but iirc what we decided on and you built was to just do a best effort of it now that fails closed until we can figure that out so we at least don't have a gaping security hole, as we'd rather that failed closed behavior occasionally accidentally blocking data from being streamed that should have been allowed as opposed to the inverse.

My notes on this PR include:
- [ ] a case in `policy/rowfilter.go` where I think we actually fail open? Would love a test case to confirm that first, and then if my theory is correct, for it be closed and test case to enforce it so it can't regress. So like `"9" > "100"` being a test that should NOT pass I think, right? Also need to make sure `_neq` properly works and fails closed too for instance, where like UUIDs that have different caps applied but that CH might see as the same etc need to be equal but the current code would pass them through to trivial string comparisons I think.
- [x] a reference to some of the perf / memoization stuff in the code to a new issue I opened (#435) for it just to document
- [ ] a small req for us to add in some o11y for when rls filters/withholds data and some other minor o11y fixes
- [ ] test case expansion, specifically for the hub stuff: I don't think that the dual goroutine interaction was properly being tested with race coverage for sub claims?
- [ ] a minor immutability req on `stream/subscriber.go`'s claims since those (I don't _think_ need to be mutable? Feel free to pushback if I missed something though...) shouldn't change/hot-reload or at least have no mechanism to mid-stream right now (which kind of leads into the broader #239 issue...)
- [ ] a bunch of smaller stale comment or docs type things – not sure if you have been or not, but HIGHLY recommend/suggest you get in the practice of running `/prepush` or `/docs-review` for just docs (or `/pr-review-locally` for other PR reviews) in `claude` on your local worktree/branch before requesting human review etc as it does a good job at catching low hanging fruit, especially these sorts of doc inconsistencies for you

internal/policy/rowfilter.go
taitelee: Fixed by replacing the bool map with `policy.ColumnKind`, whose zero value acts as a strict fallback. A nil map (no registry / no schema), a column missing from the map, and an unknown type all land on `ColumnOpaque`, where ordering and != refuse to compare (row withheld) and only byte-equality admits. There is no lexicographic fallback left to downgrade into. A known String column is the only place bytewise ordering survives, and there it's exact (byte order is ClickHouse's String order).

The boot-time discovery failure window is now fail-closed. When `Hub.columnKinds` returns nil, every column is Opaque, meaning equality scoping keeps working while ordering and != withhold until a schema lands. Addressed in `TestRowVisible_Ordering_SchemaInformed` (nil-map, unknown-column, and Opaque-type cases all asserted withheld, and the old "9" > "100" assertion is inverted) and at the hub level in `TestHub_RowFilter_NumericOrdering_SchemaInformed`, which now also runs the same policy against a hub with no registry and asserts both rows are withheld.

internal/stream/hub_test.go
taitelee: Right, the only concurrent test was passthrough, so `sub.claims` was never read under -race. Added `TestHub_ConcurrentRowFilteredBroadcast_Race`: 4 goroutines broadcasting on a row-filtered role while 4 more construct/Add/Remove claims-bearing subscribers, so the fan-out's per-subscriber claims read runs concurrently with subscriber publication in the unit suite's -race run. Combined with the constructor-injection change, the happens-before edge is bucket-mutex-in-Add by construction, and this test is the tripwire if anyone reintroduces post-Add mutation.

docs/src/content/docs/architecture.md
taitelee: The Streaming Path flow block now says the fan line delivers per subscriber for a role carrying a row filter (claims must admit the row), and the policy-filtering stanza names the row filter alongside table/column enforcement, with "column projection runs once per role" scoped to exactly that.

docs/src/content/docs/access-control.mdx
taitelee: Fixes:
- Scope: The caution now states the real classes. Numeric types compare numerically, String compares bytewise (exact for equality and ordering), and every other type (Enum, UUID, Date/DateTime, Bool, IPs, FixedString, etc.) is equality-only, not just Bool and DateTime.
- Direction: `_neq` no longer fails open. On anything but numeric or String columns, it withholds every row because inequality is unprovable from bytes. This means the uppercase-UUID and `1` for `Bool` cases now withhold instead of admitting. The caution leads with the invariant that ambiguity only ever withholds a row the query path would return, and never delivers one it would hide.

docs/src/content/docs/access-control.mdx
taitelee: Added to the caution: the stream evaluates the ingested event payload, not the stored row. This means a filter keyed on a column the payload doesn't carry withholds every event for that subscriber, while the same filter matches normally on the query path. This bites MATERIALIZED and ALIAS columns (which are never in a payload) or a DEFAULT column that clients omit.

internal/stream/hub.go
taitelee: Now recorded. The CHANGELOG's #319 entry carries an explicit perf note. On a topic whose role has a row filter, part of the once-per-role gain is traded back (`Evaluate` runs per subscriber per event, the column projection stays shared, and unfiltered roles are unaffected). Issue #435 tracks memoization and uses the restored benchmark as the measuring stick. I can change the PR description.

internal/stream/hub.go
taitelee: Added `wavehouse_sse_rows_withheld_total`, labeled by table and role (bounded by the policy, not by data). This is incremented on every row-level-security withhold for live fan-out and replay alike. This means all the silent fail-closed shapes listed (absent filter column, non-scalar, unparseable, no schema, plus the new ordering/_neq-on-opaque withholds) are in one visible counter. Docs point at it from the access-control caution and the architecture metrics bullet. `TestHub_RowFilterWithheldIncrementsMetric` covers both paths.

internal/stream/subscriber.go
taitelee: `NewSubscriber(claims)` now takes them at construction and `SetClaims` is deleted, so the invariant is structural. There is no API through which claims can change after publication, and the field's doc comment now states the mechanism (publication happens-before the fan-out read via the bucket mutex in Add) rather than just the consequence. The handler passes claims straight through, and mid-stream re-auth/revocation stays scoped out.

internal/stream/hub.go
taitelee: Fixed with the pattern from next door, using per-connection instead of per-event. `ReplayFrame` became `ReplayProjector(role, claims)`, which returns the projection closure the handler now creates once per gap-fill. The closure caches the column-kind map for the table and re-resolves only if the table name changes (which a single-topic stream never does). An N-event replay is now one registry lookup plus one map build.

internal/stream/hub.go
taitelee: Made changes to both places. The stream/ bullet now qualifies once-per-role with the filtered-role per-subscriber delivery (and notes claims are fixed at construction). Key design decision #12 gains the row-level same thing: one `resolvePredicates` resolution feeding both `predicatesToSQL` (query) and `RowVisible` (stream, fail-closed `ColumnKind` comparison). This makes sure "row visibility can't drift either", and is tagged to #319.

internal/stream/hub_test.go
taitelee: Restored just the benchmark, with no compiled-filter machinery. `BenchmarkBroadcast_RowFilteredFanout` tests 100, 1k, and 10k subscribers (half visible and half withheld) to measure the per-subscriber `Evaluate` and `RowVisible` cost that the thread was about. First numbers on my box are ~99µs/event at 100 subs, ~1.44ms at 1k, and ~16ms at 10k (roughly 1.6µs/subscriber, linear, with ~16 allocs each). So the condition the thread was resolved on is evaluable again, and #435 has a baseline.

internal/api/stream.go
```diff
@@ -78,7 +81,7 @@ func (h *StreamHandler) Handle(w http.ResponseWriter, r *http.Request) {
 	// Register for live events before gap-fill so events arriving during replay
 	// buffer in the subscriber queue instead of being missed (an overlap with
 	// replay yields duplicates, deduped client-side by id: — at-least-once).
-	sub := stream.NewSubscriber()
+	sub := stream.NewSubscriber(claims)
```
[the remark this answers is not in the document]
taitelee: `NewSubscriber` does take ownership with a deep copy of the JSON-like claims tree (`cloneClaimsMap`, where nested maps and slices are copied, and immutable scalars are shared). There is deliberately no setter, so the subscriber holds an immutable snapshot fixed at construction.

Publication happens under the bucket mutex in `Hub.Add`, making the fan-out's unsynchronized read race-free structurally. Added `TestHub_RowFilter_ClaimsSnapshotImmuneToCallerMutation`, which mutates a nested claim in the caller-retained map after `Hub.Add` and asserts that visibility follows the snapshot. Also added a -race concurrency test covering Add, Remove, and Broadcast churn.

Do you see it?

internal/policy/rowfilter.go
```diff
@@ -0,0 +1,205 @@
...
+		return 0, false
+	default:
+		return 0, false // unknown future kind: refuse to compare, fail closed
+	}
+}
+
+// compareExact compares two numeric strings at arbitrary precision — the tie-break
+// for operands float64 cannot tell apart. ok=false when either side isn't an exact
+// rational (±Inf, malformed), failing the predicate closed.
+func compareExact(a, b string) (int, bool) {
```
EricAndrechek: **A subscriber can be delivered another tenant's rows.** This fails open, and it's the one direction the CHANGELOG entry says is impossible.

`compareExact` compares the event's exact digits against the filter constant — but the constant was already rounded before it got here:

- `internal/auth/auth.go:206` calls `jwt.Parse` without `jwt.WithJSONNumber()`, so a numeric JWT claim decodes as `float64`.
- `resolveTemplate` (`internal/policy/policy.go:328`) stringifies it with `fmt.Sprint`, turning `10000000000000001` into `"1e+16"`.
- `compareExact` then does an exact comparison against a value whose digits were already lost.

Reproduced with a real signed HS256 token carrying `tenant: 10000000000000001`:

```
decoded claim: float64 1e+16 -> fmt.Sprint = "1e+16"

tenant 10000000000000001  ->  WITHHELD    (the subscriber's own rows)
tenant 10000000000000000  ->  DELIVERED   (a different tenant's rows)
```

The query path binds the same constant and ClickHouse rejects it outright (`WHERE tenant_id = '1e+16'` → `Code: 53`, cannot convert to `UInt64`), so the stream is strictly more permissive than the query path here.

Needs a bare-JSON-number integer claim above 2^53 against a numeric column — Snowflake-style tenant IDs, which is exactly the case the `math/big` tie-break was written for.

`jwt.WithJSONNumber()` on the parser fixes the stream, the query path and the `check`-clause auto-inject together, but it lands in `internal/auth` — outside this PR's footprint, so worth a call on scope. Failing the predicate closed when a claim arrives as a `float64` at or above 2^53 would keep it in-package.

internal/stream/hub_test.go
```diff
@@ -323,6 +535,193 @@ func TestHub_ConcurrentAddRemoveBroadcast_Race(t *testing.T) {
...
+		Tables: map[string]policy.TablePolicy{
+			"clicks": {Select: map[string]policy.RolePermissions{
+				"viewer": {Filter: map[string]policy.Filter{"tenant_id": {Eq: new("{{ jwt.tenant }}")}}},
+			}},
+		},
+	}
+	hub := NewHub(policy.NewMemoryStore(p), reg, nil)
+	const topic = "ingest.clicks"
+
+	neighbor := NewSubscriber(map[string]any{"tenant": "10000000000000000"}) // float64-equal neighbor
```
EricAndrechek: These claims are strings, and `jwt.Parse` never produces a string for a bare JSON number — it yields `float64`. So this test passes while the production path it stands for does not (see my note on `rowfilter.go`).

Constructing the subscriber from claims that came out of `jwt.Parse`, rather than a literal map, would have caught it. Worth doing for any test that asserts a claims-derived guarantee.

docs/src/content/docs/access-control.mdx
```diff
@@ -370,12 +370,27 @@ The same policy drives every data path, but not every field is meaningful on eve
...
+- **No usable schema** — the table is unknown to schema discovery, or discovery is still failing at boot (the server serves while retrying in the background): every column is treated as the "other" bucket above. Equality scoping keeps working; ordering and `_neq` withhold until a schema is available.
+
+A few more edges worth knowing when you write a policy — the stream evaluates the **ingested event payload**, not the stored row, and each of these follows from that:
+
+- **A filtered column the payload doesn't carry withholds *every* event** for that subscriber, even though the same filter matches normally on the query path. That bites a `MATERIALIZED`/`ALIAS` column (never part of an ingest payload) or a `DEFAULT` column your clients omit. The recommended [`check` + `filter` pairing](#insert-checks) is unaffected: an `_eq` insert `check` auto-injects its claim value into any payload that omits the column *before* the event is published, so the streamed event carries it and the matching row filter evaluates normally.
```
EricAndrechek: This guarantee doesn't hold for `DateTime`/`DateTime64` columns since the #402 merge, and the problem is wider than the `check` + `filter` pairing.

The check runs at `internal/api/ingest.go:374` and injects the raw claim value; `CanonicalizeTimestamps` rewrites it at `:406`, before the publish. The stream then compares the canonicalized payload against the un-canonicalized claim under `ColumnOpaque` byte-equality, so every row is withheld. Auto-injection isn't required — a client-sent value takes the same path, because the check compares pre-canonicalization and the filter compares post. In practice **any `_eq`/`_in` row filter on a timestamp column** is affected, since ingest rewrites every producer spelling.

The harder part: there is no constant that works on both surfaces. Verified against ClickHouse 25.10:

```
WHERE created_at = '2026-06-21 04:00:00'   -> 1 row     (stream withholds every row)
WHERE created_at = '2026-06-21T04:00:00Z'  -> Code: 53 TYPE_MISMATCH
```

`date_time_input_format=best_effort` doesn't help; it governs insert parsing only. So the advice two bullets up to "pick a value that works on both surfaces" isn't actionable here — the two spellings are disjoint. Replay is affected identically, and no test currently uses a `DateTime`-typed filter column.

It fails closed, so this is availability rather than exposure. Two ways out, and the choice is worth making explicitly:

1. Give `DateTime`/`DateTime64` their own `ColumnKind` and run the filter constant through the same parser before comparing. That makes `_eq`/`_in` spelling-independent and would also unlock `_gt`/`_lt` on timestamps — the natural shape for a time-window policy.
2. If that's out of scope here, say plainly that a timestamp row filter can't be enforced on the stream, and drop the "pairing is unaffected" sentence. `AGENTS.md:68` points at #381 for fail-closed enforcement on this type, so that pointer would need qualifying too.

docs/src/content/docs/access-control.mdx
```diff
@@ -370,12 +370,27 @@ The same policy drives every data path, but not every field is meaningful on eve
...
+- **`String` columns** (again under any `Nullable`/`LowCardinality` wrapping): byte comparison *is* ClickHouse's String comparison — equality and ordering are both exact.
+- **Every other type** (`Enum`, `UUID`, `Date`/`DateTime`, `Bool`, `IPv4`/`IPv6`, `FixedString`, …): only byte-equality is trusted. `_eq`/`_in` admit exactly the event's own text rendering — write the filter value the way your events carry it (`true`, not `1`; a lowercase UUID if that's what clients send). The same constant is bound into the query path's SQL, where ClickHouse compares it against the column's declared type rather than the event's text rendering — so pick a value that works on both surfaces, and verify the query path returns what you expect before relying on the filter. `_neq`, `_gt` and `_lt` withhold **every** row: a text difference can be pure representation (an uppercase UUID, an alternate date format, an Enum name vs. its number), so inequality and order are unprovable without ClickHouse — on these columns, use the query path for ordering/exclusion filters.
```
EricAndrechek: "Write the filter value the way your events carry it" reads as "the way your clients send it," which is no longer true for timestamps: since #402, ingest rewrites every parseable top-level `DateTime`/`DateTime64` value to canonical RFC 3339 UTC before publishing (`internal/discovery/timestamp.go:36`). WaveHouse decides the spelling, not the producer.

This is the page policy authors actually write filters on, and it never mentions canonicalization — the only statement of the rule is in `api.md`, in future tense (separate comment).

The bullet also groups `Date`/`DateTime`, but `timestamp.go:25-27` excludes `Date`/`Date32` from canonicalization, so the two need opposite advice: `Date` keeps the producer's spelling, `DateTime` doesn't.

docs/src/content/docs/api.md
EricAndrechek, line 1: `api.md:282` still describes this feature as unshipped:

> …and the form the stream row-filter **will require** for timestamp comparisons **once row-level enforcement lands** ([#381])

This PR is #381 — `internal/stream/hub.go:171-176` evaluates `RowVisible` per subscriber today. #402 anticipated this interaction and left the line as a placeholder for whoever landed row-level enforcement, so it's (y)ours to close out.

Present tense, and it should say what actually happens: `_eq`/`_in` match the canonical form, `_neq`/`_gt`/`_lt` withhold every row on a timestamp column. Worth repointing the link at the access-control enforcement caution rather than the issue.

(File-level because line 282 isn't in this diff.)

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
...eld_total` counter (labeled by table and role, on the live and replay paths alike), separating "no matching rows" from "a fail-closed filter is withholding everything". Subscriber claims are fixed at construction (`stream.NewSubscriber(claims)` — no setter), so the fan-out's unsynchronized claims read is race-free structurally, with a dedicated `-race` test driving concurrent row-filtered broadcasts. Replay (gap-fill) applies the same per-connection row check via `Hub.ReplayProjector`, which caches the per-table column-kind lookup across the replay loop. **Perf note:** on a topic whose role carries a row-filter, part of the #294/#353 once-per-role fan-out gain is traded back for correctness — `policy.Evaluate` runs per subscriber per event there (the column projection stays shared; roles without a filter are unaffected); #435 tracks memoizing it, and `BenchmarkBroadcast_RowFilteredFanout` exists to measure it. The resource limits (`max_rows`, `max_execution_time`, …) remain a query-path property and are still **not** applied to the stream (a separate, documented boundary). Supersedes the "stream path applies no row-level filter" invariant noted in the #294/#353 Changed entry below.
```
EricAndrechek: Two things:

**Direction.** This entry closes with "the #294/#353 Changed entry **below**". That entry is at line 28, above — `### Changed` precedes `### Security`.

**The superseded entry still reads as current.** Line 28 says "the stream path applies no row-level filter (a documented invariant — if row-level filtering is ever added to streaming, the key must take claims into account)". Both halves are now wrong, and they ship in the same `Unreleased` block as this entry. The stream does apply a row filter; and the prescription is the design this PR deliberately rejected — `hub.go:51` still keys buckets on `(topic, role)`, with claims on the `Subscriber` gating delivery per subscriber. Someone reading only the Changed entry walks away with a false invariant and a stale instruction.

Worth a forward pointer on line 28 rather than leaving it to be inferred.

docs/src/content/docs/access-control.mdx
```diff
@@ -370,12 +370,27 @@ The same policy drives every data path, but not every field is meaningful on eve
...
+- **A non-scalar event value** (array/object/null) under a filtered column withholds the row.
+- **Insert-time numeric narrowing bounds the fail-closed guarantee itself** — this one edge can fail *open*. The guarantee holds for the *payload* value, and the one place that differs from the stored row is a payload with more precision than the column's declared type (a `Decimal`'s scale, `Float32` width), which is rounded on insert — so a numeric `_neq`/`_gt`/`_lt` filter on such a column compares the pre-rounding payload, and can deliver an event whose stored row lands on the other side of the comparison: a `_gt: "1.004"` on a `Decimal(10, 2)` column delivers a payload of `1.005` on the stream, while the query path compares the stored `1.00` and hides the row. `_eq`/`_in` cannot fail open here: ClickHouse narrows the bound filter constant to the column's declared type just as insert narrowed the payload, so an equality that matches on the stream matches the stored row too. Columns that store exactly (integer IDs, `String` tenants) are unaffected under every operator.
```
EricAndrechek: ClickHouse truncates decimals, it doesn't round them — verified on 25.10, where `1.005`, `1.006` and `1.009` all store as `1.00` in a `Decimal(10,2)`.

That truncation is what makes this bullet's own example work. A reader who takes "rounded" literally predicts `1.01` for `1.005` and concludes the example is broken.

`Float32` genuinely does round to the nearest representable value, so the two cases differ in direction as well as name and are worth stating separately.

docs/src/content/docs/api.md
```diff
@@ -577,7 +577,7 @@ Each SSE connection is bound to a single `?table=`; to consume multiple tables,
...
 
-**Note:** When access control policies are active, streamed events are filtered per the caller's role — denied columns are removed and tables without select permission are skipped.
+**Note:** When access control policies are active, streamed events are filtered per the caller's role — tables without select permission are skipped, denied columns are removed from each event, and the role's row-level `filter` is evaluated per subscriber against the caller's JWT claims — so, for a filter constant written to be valid on both read surfaces (see the enforcement caution), a connection is never delivered a row the query path would hide for that role, with one documented exception (a numeric `_neq`/`_gt`/`_lt` filter over a column type that rounds the payload on insert). The claims come from the connection's token (the `Authorization` header, or the `?token=` fallback above), and replayed gap-fill events are filtered the same way. See [Access control — row-level security](/access-control#row-level-security) and the enforcement caution there for the stream's fail-closed comparison boundary.
```
EricAndrechek: This note opens with a 93-word sentence carrying three enforcement rules, a conditional ("for a filter constant written to be valid on both read surfaces"), and a parenthetical exception. It's the primary user-facing statement of a security guarantee and shouldn't need a second read.

It also says "see the enforcement caution" without a link, then links that same caution two sentences later.

Three sentences would do it: the enumeration; then the guarantee and what it depends on; then the exception, with the link attached at first mention.

Note the conditional will need revisiting either way — see my comment on `access-control.mdx:387` about there being no filter constant valid on both surfaces for timestamp columns.

docs/src/content/docs/access-control.mdx
EricAndrechek: New comment in new review, closing this one

internal/stream/hub.go
EricAndrechek: Up to you, I don't really care either way

EricAndrechek (changes requested):
Nice work, basically everything from the previous round is fixed/addressed. Two new things:
- [ ] I _think_ (Claude tested it for me and "confirmed" it, but obviously take that with a grain of salt) that the JWT claim mechanism may lead to improper number comparisons when really large
- [ ] A few fixes need to be made to timestamp filtering now that #402 has been merged in

internal/policy/rowfilter.go
taitelee: Fixing so that the middleware now parses with `jwt.WithJSONNumber()` so numeric claims keep their exact digits, and `claimString` refuses any float64 at or above 2^53 as an in-package backstop, so the predicate matches no rows instead. Smaller floats render positionally now too, which also fixes the check clause auto-inject for those claims.

internal/stream/hub_test.go
taitelee: Ok. Now I have the row filter hub testst build claims by running a real signed token through `auth.Middleware` (`jwtClaims` helper) instead of literal maps, which is what surfaced the float64 issue.

docs/src/content/docs/access-control.mdx
taitelee: Timestamps moved out of that bullet entirely since they compare as instants now, and the byte-equality bullet calls out that `Date` keeps the producer's spelling while `DateTime` doesn't.

docs/src/content/docs/api.md
taitelee: Will update on next push.

CHANGELOG.md
taitelee: Will fix in current.

docs/src/content/docs/access-control.mdx
taitelee: Right, verified the same on 25.10. The two cases are stated separately now: Decimal truncates, Float32 rounds.

docs/src/content/docs/api.md
taitelee: Rewrote it as three sentences with the link at first mention. The conditional got revisited as part of the timestamp fix since a zone-less constant now works on both surfaces.

EricAndrechek (changes requested):
So I am going to send this back for changes before finishing reviewing it all so that the larger requirements can start while I do more reading, but I think some changes larger than you may have been anticipating are going to be required here, meaning I will have to pretty heavily re-review after them anyway.

Basically, a LOT of the code I am reading and reviewing now is giving me flashbacks to the PR we JUST finished on (#457) and cases we already spent many iterations going back and forth on there, etc finding so so many niche edge cases and closing them, etc. It feels like we now have a lot of duplicate code or code attempting to do basically the exact same things in two different places – half of it pre-existing or already a WIP in this PR prior to #457, and the other half a more robust version iterated on and approved in #457.

For instance, in #457 we spent a lot of time working on coercing and comparing types and canonicalizing them safely, like in `internal/policy/policy.go` with the `CanonicalScalar` function that had that digit bound check for loop and max digits const and everything – yet (as far as I can tell) we are doing something very very similar in the new `internal/policy/rowfilter.go` file and its `scalarString` function. Obviously there are some differences across the code, but you can immediately see how much less robust the code is in `rowfilter.go` compared to the new code from #457, and how similar many of the functions are in what they are trying to do.

I think that the merge in from #457 was more complicated than we'd think it was going to be, and that we need to do a significant amount of refactoring what was already in flight in this PR to adopt/extend the work done in #457 and build off of it and its more rigorously tested type coercions and comparison logic.

taitelee:
@coderabbitai review

internal/stream/hub.go
```diff
@@ -122,23 +132,128 @@ func (h *Hub) Broadcast(topic string, raw []byte) {
...
+					spec.Numeric = policy.NumericSpec{Family: policy.NumericDecimal, Precision: st.Precision, Scale: st.Scale}
+				}
+			}
+			m[c.Name] = spec
+		case discovery.IsStringType(c.Type):
+			m[c.Name] = policy.ColumnSpec{Kind: policy.ColumnText}
+		}
+	}
+	return m
+}
```
EricAndrechek, line 229: I feel like this function is out of place being defined and used only in `stream/hub.go`, it feels more related to the discovery of tables and their schemas (and could then be done ONCE per table when its schema changed and save there) or in the policy potentially too – but I can also understand it here as we really only use the native go type comparisons like this on the stream data. Not going to make this a blocker for this PR, fine here, but may want to tweak or refactor in a later PR, just commenting.

internal/stream/hub.go
```diff
@@ -149,50 +264,75 @@ func (h *Hub) snapshotPolicy() (p *policy.Policy, filter bool) {
...
+// event), so a large Last-Event-ID gap-fill doesn't pay one registry lookup and map
+// build per event. The closure is for a single goroutine — each connection makes
+// its own. The live path uses Broadcast.
+func (h *Hub) ReplayProjector(role string, claims map[string]any) func(raw []byte) (Frame, bool) {
+	var colSpecs map[string]policy.ColumnSpec
+	specsFor := "" // table name colSpecs was resolved for ("" ⇒ not yet resolved)
+	return func(raw []byte) (Frame, bool) {
+		var evt ingest.EventMessage
+		decoded := decodeEvent(raw, &evt)
+		p, filter := h.snapshotPolicy()
```
EricAndrechek: Didn't realize it until reviewing this PR, but I *think* this could go up out of the returned function, otherwise its re-acquiring a cache read lock on the policy on every event replayed which seems unnecessary, especially seeing as we use the same auth from the connection for the entire length of the SSE stream anyway...

Not really blocking, just something that you should probably at least mention/flag in the new issue we opened on one of the general SSE-related perf issues, like #435

internal/stream/hub.go
```diff
@@ -122,23 +132,128 @@ func (h *Hub) Broadcast(topic string, raw []byte) {
...
 		frame := Frame{Kind: KindEvent, Data: wire}
+
+		if !perms.HasRowFilter() {
+			// No row-level security for this role: one projection serves the whole
+			// bucket, regardless of per-subscriber claims (the #294/#353 fast path).
+			for _, sub := range rb.bucket.Snapshot() {
+				if !sub.Send(frame) {
+					h.metric.FrameDropped(KindEvent)
+				}
+			}
```
EricAndrechek: Why not use the bucket's `Push` function for this, and then just move the frame dropped metric call into that too?

EricAndrechek: Actually, why not move the frame drop metric caller fully into the subscriber's `Send` implementation, so it's always called on drops without us having to think about remembering to add it everywhere like this?

internal/stream/hub.go
```diff
@@ -122,23 +132,128 @@ func (h *Hub) Broadcast(topic string, raw []byte) {
...
+			specsResolved = true
+		}
 		for _, sub := range rb.bucket.Snapshot() {
+			subPerms := policy.Evaluate(p, rb.role, evt.TableName, "select", sub.claims)
+			if !subPerms.RowVisible(evt.Data, colSpecs) {
+				h.metric.RowWithheld(evt.TableName, rb.role)
+				continue // this row is filtered out for this subscriber
+			}
 			if !sub.Send(frame) {
 				h.metric.FrameDropped(KindEvent)
```
EricAndrechek: See above, if we move this to `Send` we can forget about adding this code everywhere – wasn't a big thing before since we only had 1 or 2 call sites, but with sending per subscriber in different places now and the broadcasts to all of them etc it's starting to get more mentally taxing to remember to add imo

internal/stream/hub.go
```diff
@@ -149,50 +264,75 @@ func (h *Hub) snapshotPolicy() (p *policy.Policy, filter bool) {
...
+				specsFor = evt.TableName
+			}
+			subPerms := policy.Evaluate(p, role, evt.TableName, "select", claims)
+			if !subPerms.RowVisible(evt.Data, colSpecs) {
+				h.metric.RowWithheld(evt.TableName, role)
+				return Frame{}, false // this row is filtered out for these claims
+			}
+		}
+		return Frame{Kind: KindReplay, Data: wire}, true
 	}
```
EricAndrechek, line 317: This function is nearly identical to the check that happens in the `Broadcast` loop – only difference is this doesn't actually `Send` (or `Push`), but rather returns the frame for the actual stream handler to write to the wire and flush since it happens prior to the stream subscriber's loop – but I feel like it could be abstracted out for both to use, even if not the whole thing at least some of the:
```go
subPerms := policy.Evaluate(p, role, evt.TableName, "select", claims)
if !subPerms.RowVisible(evt.Data, colSpecs) {
	h.metric.RowWithheld(evt.TableName, role)
```
logic. Thoughts? Genuinely possible that it's just not worth keeping DRY here but lmk what you think

internal/policy/rowfilter.go
```diff
@@ -0,0 +1,551 @@
...
+	m := make(map[int]struct{ sMin, sMax, uMax string }, 6)
+	one := big.NewInt(1)
+	for _, bits := range []int{8, 16, 32, 64, 128, 256} {
+		uMax := new(big.Int).Sub(new(big.Int).Lsh(one, uint(bits)), one)
+		sMax := new(big.Int).Sub(new(big.Int).Lsh(one, uint(bits-1)), one)
+		sMin := new(big.Int).Neg(new(big.Int).Lsh(one, uint(bits-1)))
+		m[bits] = struct{ sMin, sMax, uMax string }{sMin.String(), sMax.String(), uMax.String()}
+	}
+	return m
+}()
```
EricAndrechek: I'm sure this is performant and whatnot, but it is abhorrently unreadable… like I'm sure given enough time I could make sense of it. I understand it is likely a quite common filter in the hot path for ALL subscribers on streamed data though, so perf is important for it. Could we at least comment each line to explain what it's doing/means and/or break some of it up more or something?

internal/policy/rowfilter.go
```diff
@@ -0,0 +1,551 @@
...
+func (n NumericSpec) integerInRange(c string) bool {
+	b, ok := intBounds[n.Bits]
+	if !ok {
+		return false
+	}
+	if n.Unsigned {
+		return compareCanonicalDecimals(c, "0") >= 0 && compareCanonicalDecimals(c, b.uMax) <= 0
+	}
+	return compareCanonicalDecimals(c, b.sMin) >= 0 && compareCanonicalDecimals(c, b.sMax) <= 0
+}
```
EricAndrechek: Why only one of these for integer range, are other numeric types like floats not subject to this same issue?

internal/policy/policy.go
```diff
@@ -502,6 +589,46 @@ func canonicalDecimal(lit string) (string, bool) {
...
 }
 
+// compareCanonicalDecimals orders two canonical decimal forms (CanonicalScalar
+// output) as numbers, by digit-string arithmetic alone — the comparison twin of
+// canonicalDecimal, sharing its invariants: an optional leading '-' (never on
+// zero), no leading integer zeros except a lone "0", no trailing fraction
+// zeros, no exponent. Those invariants are what make the string operations
+// sound: with no leading zeros a longer integer part IS the larger magnitude,
+// and with no trailing zeros a fraction that is a proper prefix of another IS
+// the smaller. Never a float round-trip, so 64-bit-plus IDs order exactly.
```
EricAndrechek: Can floats not have scientific notation that would impact magnitude/comparisons not based on length stuff? Like `1e-3` or something?

internal/policy/rowfilter.go
EricAndrechek: Coming back to this, yeah I think it is – I think Copilot's comment is flagging the same thing

internal/policy/rowfilter.go
```diff
@@ -0,0 +1,551 @@
...
+		// rendering FormatFloat emits that canonical form forbids ("-0").
+		s, ok := CanonicalScalar(x)
+		if !ok {
+			return "", false
+		}
+		return CanonicalNumericLiteral(s)
+	default:
+		return "", false
+	}
+}
```
EricAndrechek: This function is really important to be called, especially the sub-functions it calls within `policy.go`, but I am having an INCREDIBLY hard time tracking all the function names and call sites and returns, and following data through various paths confirming it goes through the proper variety of various canonicalization functions throughout this file and the policy file. Apart from making more classes for pseudo-types and abstracting more to consistently used paths, and/or some function renaming/merging (into classes maybe), I don't really know what a great solution to this would be, apart from stronger types like we'd get in Rust or C++ – all the more reason to try vendoring ClickHouse's types properly into their own library so we don't need to manage this stuff...

internal/stream/hub.go
EricAndrechek: Don't need to necessarily fix this in this PR, so scope stays smaller and we can get this shipped, although it was introduced here so does make sense here. Your call either way, but if not in this PR make an issue for that sort of cleanup+metrics to track it at least.

EricAndrechek (changes requested):
Much better use of what we merged in from the prior PR, but `policy/policy.go` and `policy/rowfilter.go` are quickly become massive files that are very difficult to read. I think the only BLOCKING changes to this PR I have are on these files.

internal/stream/hub.go
taitelee: Ok I agree with this I think it's better for organizaiton anyways. I'm going to make the `Evaluate`, `RowVisible` and `RowWithheld` part be consolidated into a single `Hub.rowAdmitted` check used by both paths, so the two delivery surfaces can't differ on how rls is evaluated.

internal/stream/hub.go
taitelee: It is definitely schema adjacent. But I have it here because it renders discovery's type knowledge into `policy.ColumnSpec`, and only the stream takes that in right now. But precomputing it per table at schema refresh belongs with the #435 memoization batch stuff, so I can add it to that issue (Claude pointed this out on my end). One thing for now is that the numeric half is now `stream.NumericSpecOf`. The integration oracle uses this exact same exported mapping, so if we ever need to move it elsewhere, it's just a one line change.

internal/policy/rowfilter.go
taitelee: Added some comments. Basically, it's a range checker for int types. It maps each ClickHouse integer width (8, 16, 32, 64, 128, and 256 bits) to the smallest and largest values that width can store. These limits are saved as three decimal strings per width: signed min, signed max, and unsigned max. For Int8, that means `-128`, `127`, and `255`. It uses `big.Int` to compute these because `Int128` and `Int256` exceed native Go integer types. The limits are stored as strings since the row filter compares numbers as canonical decimal strings via `compareCanonicalDecimals`. This lets us compare the string bounds directly against the operands.

The only consumer is `integerInRange`. This function checks if a value actually fits in a column before allowing any comparison. If the event value or the filter constant falls outside the column range, the comparison is refused and the row is withheld. We have to refuse these cases because ClickHouse is inconsistent with out of range constants. Depending on the situation, ClickHouse might throw a query error, promote the type and compare, or wrap around and compare a completely different number. Modeling just one of those behaviors could result in the stream delivering a row that the SQL path would hide. Refusing the comparison is the only safe approach across all of those scenarios.

internal/stream/hub.go
taitelee: Just ended up making a change for it. I have `ReplayProjector` now taking a policy snapshot for the whole gap-fill instead of re-acquiring the store lock for every replayed event. This also matches the one-snapshot-per-event semantics in `Broadcast`. If a reload lands mid-replay, it will apply starting from the first live event. Had Claude add documentation for this update in the projector comment, the changelog, and the three docs pages that mentioned the old timing.

internal/stream/hub.go
taitelee: Will move `FrameDropped` so it's now inside `Subscriber.Send` and is labeled by the frame's kind. I'll have the fast path back to `bucket.Push`. However this does mean that a keepalive dropped by a full queue now counts too under `kind="keepalive"`, whereas it used to be silent. The test I have written will assert both kinds through the real paths.

internal/stream/hub.go
taitelee: This is covered by the same change. Call sites no longer touch the metric. Send is now the only `FrameDropped` caller.

internal/policy/rowfilter.go
taitelee: Floats face the same issue and are also gated, just not by a table. The `narrowFloat` function parses at the exact bit width of the column. Go reports overflow as an error, so if a magnitude is too large for the column to hold, the comparison is refused and the row is withheld. This is covered by the `1e39` into `Float32` and the `Inf` spelling tests in `rowfilter_test.go`. Decimals get their range gate from `decimalInPrecision`. Integers alone need the explicit bounds table because their comparison is digit string arithmetic with no inherent width. I added a comment for `integerInRange` to explain this. Copilot's thread was about a related but different precision narrowing bypass. That was addressed in a previous commit by narrowing both operands. The differential oracle pins both behaviors against a live ClickHouse database.

internal/policy/policy.go
taitelee: String containing exponent can't reach this function. Both operands are canonicalized before any comparison, and the canonical form is positional by construction. `canonicalDecimal` expands `1e-3` to `0.001` and `1e3` to `1000`. I have exponent tests in `policy_test.go` that cover this. This is why having no exponent is one of the listed invariants that the digit length trick relies on. Every call site routes through `CanonicalScalar`, `CanonicalNumericLiteral`, or `numericCanonical` first.

internal/policy/rowfilter.go
taitelee: Reorganized some of the functions. All three rendering entry points now are in a new file `canonical.go`. I agree that the durable fix is a vendored ClickHouse type library. But for now this.

CHANGELOG.md
```diff
@@ -31,19 +31,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 - **Go toolchain requirement bumped to 1.26.5** (`go.mod`): the `go` directive moves from `1.26.4` to `1.26.5` so local builds (via `GOTOOLCHAIN=auto`) and CI's `setup-go` (which reads `go.mod` via `go-version-file`) install a Go whose standard library clears the `govulncheck` findings GO-2026-5856 (`crypto/tls`) and GO-2026-4970 (`os`), both fixed in 1.26.5 — those findings were failing `make verify`'s `vulncheck` leaf (and with it the pre-commit hook) on every tree. Patch-level toolchain bump only — no source changes — and the released binaries pick up the patched stdlib too.
```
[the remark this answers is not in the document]
taitelee: Fair catch, but the 1.26.5 entry is main's history — this PR only added the 1.26.6 line (the bump itself is load-bearing: the pre-commit govulncheck gate requires the patched stdlib). Consolidating superseded dependency-bump entries inside Unreleased is a changelog-hygiene sweep we handle separately rather than folding into a security PR, so deferring this one; it'll collapse naturally in that pass before the next release cut.

taitelee:
@coderabbitai review

internal/policy/policy.go
EricAndrechek: Is it that they "can't reach this function" because there are currently no paths that call this function without having already turned their strings into canonicalized strings, or because this function either canonicalizes itself or guards against this case with an assertion or something? Because if it doesn't guard or ideally call canonicalization itself, and depends on the inputs having already been canonicalized, Go doesn't know the difference, they're both strings, so this will pass type checking and be allowed and not detected until it fails in prod.

taitelee: So the comparison function itself doesn't guard against exponents at all, it just trusts what it's handed. The reason `1e-3` never shows up is that every caller canonicalizes first (`numericCanonical` / `CanonicalNumericLiteral` expand it to `0.001` or refuse, and the bounds/truncation strings are canonical by construction). So it's a convention held up by the call sites, not enforced anywhere. You're right that Go can't tell the difference, they're both strings, so a future caller that skips the gate would compile fine and just compare wrong in prod. Right now the risk is contained because it's unexported and `compareScalar` is the only real entry point, plus the doc comment spells out the precondition. If we want it actually enforced I can add a small unexported canonical string type that only the canonicalizers produce and the compare functions accept, so skipping the gate won't compile without an obvious cast. Your call on this part.

EricAndrechek (approved):
This is MUCH better, much cleaner with the `canonical.go` and `numeric.go` broken out. Nothing significant enough to block merge of this, lgtm – but I'm triggering a final copilot run – make sure it doesn't flag anything and that all the comments are indeed resolved first before hitting merge. Thanks!

Issue #226, opened 2026-06-04
SSE: emit periodic keepalive heartbeats so idle streams survive proxy timeouts


EricAndrechek:
**Area:** streaming (api) · ops — reliability gap · found via WaveHouse-Stats dogfooding *(reproduced live)*

**Expected:** a long-lived SSE stream survives through a CDN/proxy even with no events flowing.

**Actual:** the handler writes `: connected` once (`internal/api/stream.go:67`) then sends nothing until an event arrives — the select loop (`:116-145`) has only context-done and incoming-event cases, **no heartbeat ticker**. On a quiet stream an intermediary (Cloudflare's edge idle-timeout, ~2 min) resets the idle HTTP/2 stream with `INTERNAL_ERROR`. Browser `EventSource` auto-reconnects so it's masked there, but `curl`/server-side consumers drop repeatedly and each reconnect re-runs gap-fill.

**Impact (Stats):** `curl` against `/v1/stream?table=gh_events` dies ~every 2 min while the table is quiet; we wrap it in a reconnect loop.

**Scope:** periodic SSE comment heartbeat on a ticker, added to the select loop.

Related: SSE per-subscriber channel buffer sizing #152 (same handler).

---
_From `WAVEHOUSE-FEEDBACK.md` (WaveHouse-Stats dogfooding); validated by code-read against `0f8826c` on 2026-06-04._

EricAndrechek:
**Reproduced live (Eric, 2026-06-04)** on the public demo, server-side:

```
curl -N "https://stats.wavehouse.dev/v1/stream?table=gh_events"
: connected
curl: (92) HTTP/2 stream 1 was not closed cleanly: INTERNAL_ERROR (err 2)
(after 2m 5s, EC2 → Cloudflare Tunnel)
```

Confirms the ~2-minute idle reset — and note it's via a **Cloudflare Tunnel** here, not just the CDN edge, so the keepalive is needed for tunneled/proxied server-side consumers generally (not one specific CDN). Browser `EventSource` masks it via auto-reconnect; `curl`/server-side consumers don't. Trivial fix (a `:\n\n` heartbeat ticker in the select loop) and it's visible to anyone curling the demo, so it's a cheap launch-polish win.

PR #402, opened 2026-07-09
fix(ingest): canonicalize DateTime column values to RFC 3339 UTC

### LINKED ISSUE #372: bug(streaming): SSE /v1/stream emits row DateTime columns zone-less while /v1/query uses RFC3339 Z
**Area:** streaming · query · sdk — bug (timestamp serialization inconsistency) · found via WaveHouse-Stats dogfooding

**Expected:** a `DateTime` column serializes identically on every read path — one unambiguous zoned RFC3339 form (`…Z`) — so a client stitching backfill (`/v1/query`) + live (`/v1/stream`) parses both on the same clock, and it matches the SSE frame's own `received_timestamp`.

**Actual:** the same `event_ts` (`DateTime64(3,'UTC')`) comes back two ways:
- `POST /v1/query` → `"event_ts":"2026-07-06T14:42:36Z"` — RFC3339 `Z`.
- `GET /v1/stream` frame → envelope `"received_timestamp":"…Z"` (zoned) **but** the wrapped row column `"data":{…,"event_ts":"2026-07-06 14:42:36.000"}` — ClickHouse-native (space separator, **no zone**). Same frame, envelope zoned / row column not.

The stream marshals the row `data` verbatim (`internal/stream/hub.go` `project()` → `json.Marshal({… "data": evt.Data})`), so a row `DateTime` keeps its native form, while `/v1/query` formats through the CH driver to RFC3339. The SDK `SSETransport` (`clients/ts/src/stream/sse.js`) passes `data` through untouched, and JS `Date.parse` reads a zone-less string as **local** time.

**Impact:** any SDK `liveQuery` consumer (backfill `.fetch()` + SSE live in one list) lands the two paths on different clocks — in `America/New_York` a live row's `event_ts` resolves **+4h** vs the same instant from `/v1/query`, sign-flipping by the viewer's UTC offset. Broke both ordering and the "N min ago" label in the Stats live feed (a 2-min-old event rendered "just now"). Every consumer must normalize today. Envelope-keyed logic (e.g. ingest latency off `received_timestamp`) is unaffected — the bug is row `DateTime` columns inside `data`.

Scope: render row `DateTime`/`DateTime64` columns in the SSE `data` payload with the same formatter `/v1/query` uses (RFC3339 `Z`) rather than the driver-default `YYYY-MM-DD HH:MM:SS.fff`.

Related: #294 (SSE delivery path), #319 (stream-vs-query policy drift — same query/stream-parity theme), WaveHouse-Stats#55 (client-side workaround shipped there)

---
_From WaveHouse-Stats `WAVEHOUSE-FEEDBACK.md` dogfooding (live `curl` capture + SDK-path analysis, dev `08c6f19`); validated by code-read against `ccd7549` (main) on 2026-07-06._

taitelee:
## Summary

The SSE stream fans out the pre-insert payload verbatim, so a row `DateTime` reached subscribers in whatever spelling the producer sent (typically zone-less), while `/v1/query` renders the stored value as RFC 3339 `Z` — JavaScript parses a zone-less string as local time, landing backfill and live on different clocks.

Ingest now canonicalizes every `DateTime`/`DateTime64` column value to RFC 3339 UTC before the NATS publish, so the one payload every consumer shares (SSE, the ClickHouse insert, the DLQ) matches the query path:

- Inputs stay liberal: RFC 3339 with any offset, `YYYY-MM-DD[ T]HH:MM:SS[.fff]`, `YYYY-MM-DD`, and Unix seconds (number or digit-string) all convert. Zone-less strings are interpreted in the column's time zone, else the server's (discovered via `SELECT timezone()`) — the same rule ClickHouse applies, so the stored instant never changes. Per-column specs (precision + resolved zone) are precomputed at schema refresh, so the per-record path parses no type strings and loads no zones.
- Canonicalization is fail-open: an unparseable value — or a column whose zone the binary can't resolve (no tzdata embedded; `Etc/UTC` maps to UTC without it) — publishes verbatim, and ClickHouse's more liberal parser stays the arbiter of insertability, with failures landing in the DLQ as before. Ingest never rejects a record over its timestamp spelling; fail-closed enforcement of the canonical form is the stream row-filter's (#381).
- The worker's insert sets `date_time_input_format=best_effort`: the default `basic` parser rejects the canonical form's zone suffix (verified live); `best_effort` is a superset, so pre-upgrade messages still in NATS parse unchanged.
- `Date`/`Date32` pass through untouched (day precision, no zone ambiguity on this path).

This is the data-plane foundation for the row-filter timestamp enforcement landing in #381.

## Related Issues

Closes #372


internal/api/ingest_test.go
```diff
@@ -1379,3 +1379,104 @@ func TestIngest_BodyCap_413(t *testing.T) {
...
+	h := NewIngestHandler(tsRegistry(), pub, testutil.NopLogger())
+
+	req := ingestRequest(t, "events", map[string]any{"name": "e", "ts": "banana"})
+	w := httptest.NewRecorder()
+	h.Handle(w, req)
+
+	assert.Equal(t, http.StatusBadRequest, w.Code)
+	assert.Contains(t, w.Body.String(), `column \"ts\"`)
+	assert.Nil(t, pub.LastMessage(), "nothing published for a rejected record")
+}
```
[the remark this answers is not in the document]
taitelee: Added `testutil.AssertJSONErrorResponse(t, w)` alongside the status assert.

internal/discovery/discovery.go
```diff
@@ -67,6 +68,18 @@ func (sr *SchemaRegistry) Refresh(ctx context.Context) error {
...
+	// timestamp strings on columns without an explicit one; ingest canonicalization
+	// must apply the same rule so it never changes which instant is stored (#372).
+	var tzName string
+	if err := sr.conn.QueryRow(ctx, "SELECT timezone()").Scan(&tzName); err != nil {
+		return fmt.Errorf("query server timezone: %w", err)
+	}
+	serverTZ, err := time.LoadLocation(tzName)
+	if err != nil {
+		return fmt.Errorf("load server timezone %q: %w", tzName, err)
+	}
```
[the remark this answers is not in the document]
taitelee: Real availability risk, but the proposed fallback trades it for silent data corruption: on a non-UTC server, interpreting zone-less strings as UTC changes which instant gets stored — the one thing this PR promises never to do. Fixed the availability half without that trade: `cmd/wavehouse` now embeds time/tzdata (~450 KB), so `LoadLocation` resolves even on images with no system zone database. A refresh failure is now only reachable when ClickHouse reports a zone newer than the binary's own tzdata, and there failing loudly is correct — boot stays non-fatal, `RetryRefresh` keeps retrying and surfaces the error via `/livez`, and no wrong instants get written. Zone resolution also moved off the hot path entirely: each timestamp column's spec (precision + zone) is resolved once per refresh and cached on the column, and an unresolvable column zone just warns and keeps rejecting that column's records per record (behavior unchanged) rather than touching the refresh.

AGENTS.md
```diff
@@ -65,6 +65,7 @@ The invariant index — what must stay true. Full narrative and rationale live i
...
+19. **Canonical timestamp wire form** — the HTTP ingest handler rewrites every `DateTime`/`DateTime64` value to RFC 3339 UTC (`discovery.CanonicalizeTimestamps`; per-column precision + resolved zone precomputed at schema refresh; `time/tzdata` embedded in `cmd/wavehouse`, an unresolvable zone fails the refresh loudly) after validation + policy checks and **before** the NATS publish, so the one payload every consumer shares — SSE subscribers (the #294 hot path), the ClickHouse insert, the DLQ — carries the same spelling `/v1/query` renders (its `transformRow` also forces UTC): live and query reads of a row can't drift (#372). Zone-less inputs are read in the column's declared zone, else the discovered server default — ClickHouse's own rule, so the spelling changes but never the instant. Don't publish an ingest payload that skipped canonicalization, and don't re-spell timestamps downstream. Preserve when touching `internal/discovery`, the ingest handler, or the SSE fan-out. Detail: architecture.md § `discovery/` + §Ingest Path.
```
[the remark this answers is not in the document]
taitelee: Half agreed. The blanket "an unresolvable zone fails the refresh loudly" was imprecise, but the suggested replacement ("warned at refresh and rejected per-record") is imprecise in the opposite direction — the behavior is 
- Server default zone (discovery.go:87-95): time.LoadLocation failure fails the refresh, deliberately (commented inline). On a non-UTC server, a silent UTC fallback would change which instant a zone-less string stores — the exact failure mode #372 exists to eliminate. Per the earlier thread on this file, boot stays non-fatal (BootState + RetryRefresh serve 503 with the diagnostic), but the refresh itself fails loudly.
- Column-declared zone (timestamp.go:97-102): warn + nil spec at refresh, per-record 400s on just that column at ingest — what your suggestion describes.

So the retrieved learning ("timezone-resolution failures during refresh should be non-fatal") applies to column-declared zones only, not the server default — please scope it that way.

Reworded the AGENTS.md line (and the matching CHANGELOG sentence) to state both tiers explicitly.

internal/discovery/discovery.go
```diff
@@ -202,13 +229,16 @@ func isNullable(chType string) bool {
...
 	for _, t := range tables {
+		resolveTimestampSpecs(t, time.UTC, logger)
 		m[t.Name] = t
 	}
 	return &SchemaRegistry{
 		tables: m,
-		logger: slog.New(slog.NewTextHandler(io.Discard, nil)),
+		logger: logger,
 	}
 }
```
EricAndrechek: Can we move this to mocking/test utils instead?

internal/discovery/timestamp.go
```diff
@@ -0,0 +1,189 @@
...
+		return timestampSpec{}, fmt.Errorf("unknown time zone %q in type %q: %w", name, chType, err)
+	}
+	return timestampSpec{precision: precision, loc: loc}, nil
+}
+
+func loadLocation(name string) (*time.Location, error) {
+	if name == "Etc/UTC" {
+		return time.UTC, nil
+	}
+	return time.LoadLocation(name)
```
EricAndrechek, line 135: In briefly looking at this function on my machine, it seems like it treats "" as UTC, and uses env variable reading for things like "Local", etc, and really only errors with invalid IANA names or if the timezone info isn't on your unix install – so the error return path from this probably doesn't run in a lot of cases, which has me wondering if that was how you intended for this to work, like if an `""` timezone should error out ( and then throw the warning that it was zone-less) or should be treated as UTC?

internal/discovery/discovery.go
```diff
@@ -61,12 +67,29 @@ func NewSchemaRegistry(conn driver.Conn, database string, refreshInterval time.D
...
+	}
+	var serverTZ *time.Location
+	if loc, err := loadLocation(tzName); err == nil {
+		serverTZ = loc
+	} else {
+		// Unresolvable — warn, not fatal, and no UTC fallback (that could move
+		// instants). A nil server zone means zone-less values pass through.
+		sr.logger.Warn("cannot resolve server timezone; zone-less timestamps will pass through un-canonicalized",
+			"timezone", tzName, "error", err)
+	}
```
EricAndrechek, line 90: See my comment in `timestamp.go` on `loadLocation` – this reads to me as if `""` or "zone-less" as I take it to mean should warn, but it doesn't appear as though that is what happens? (Plus many other timezone oddities)

EricAndrechek (changes requested):
Ok, I'm sending this back without detailed comments on much else because timezones and time parsing scares me, and I see exactly 0 tests that actually run e2e with a full clickhouse instance and various variations of datetime w/ and w/out timezone, varying degrees of precision, etc columns in their schema that we test against. For instance, what happens if you send "20260711", or "202607111500"? Does ClickHouse, with `best_effort` turned on now, interpret these as `YYYYMMDD` and `YYYYMMDDHHMMSS`, or as unix seconds?

Also, worth reading [falsehoods programmers believe about time](https://infiniteundo.com/post/25326999628/falsehoods-programmers-believe-about-time) and  [more falsehoods programmers believe about time](https://infiniteundo.com/post/25509354022/more-falsehoods-programmers-believe-about-time) – basically, it scares me a lot to do this much, and I am POSITIVE that if we started fuzzing this end-to-end with random inputs, we wouldn't have to wait very long before the time we parsed it as differed from the time clickhouse parsed it as...

So in essence, I really think we need to figure out how to be FAR more strict about what the time coming in is when parsing it or its timezone, and only use that if we are ABSOLUTELY sure we know the format they used, etc. Good luck, times are hard. :)

taitelee:
@coderabbitai review

internal/discovery/discovery.go
EricAndrechek: Confirmed moved to `testutil.NewTestSchemaRegistry` via a mock conn + real `Refresh` — good, keeps the production discovery/spec path under test rather than faking it away. Resolving.

internal/discovery/timestamp.go
EricAndrechek: `loadLocation` now rejects `""`/`"Local"` explicitly and warns + passes through un-canonicalized instead of silently guessing UTC/Local — that's exactly the behavior I was asking about here. `Etc/UTC` and genuine IANA/UTC still resolve. Resolving.

internal/discovery/discovery.go
EricAndrechek: Same as the paired `loadLocation` thread — an unresolvable server-default zone now warns at refresh and leaves `serverTZ` nil with no UTC fallback (a UTC fallback would move which instant a zone-less string stores), and column-declared zones warn per-column. Resolving.

tests/integration/timestamp_canonicalization_test.go
```diff
@@ -0,0 +1,160 @@
...
+		"2026-06-21",                 // date-only
+		"2026-06-21 04:00:00.123456", // zone-less with fraction
+		"2026-06-21T04:00:00.9999Z",  // fraction beyond column precision
+		"2026-06-21T04:00:00,999Z",   // comma fraction: ISO 8601 yes, ClickHouse no
+		"1750478400",                 // 10-digit Unix string
+		"999999999",                  // 9-digit Unix string
+		"1750478400.5",               // fractional Unix string: DateTime64-only to ClickHouse
+		"1750478400.123456789",       // ns-exact fraction (float64 would corrupt it)
+		"1750478400.9999999995",      // >9 fraction digits: ClickHouse truncates, never rounds
+		float64(1750478400),          // integer number: seconds to DateTime, *ticks* to DateTime64
```
EricAndrechek: The differential oracle is exactly what I wanted here — but it's feeding numbers as Go `float64` (lines 64–66), while the real ingest path sets `json.Decoder.UseNumber()`, so production numbers arrive as `json.Number`. That matters because `parseTimestamp` handles the two types differently:

- `float64` branch rejects `|x| > 2^53` → pass-through
- `json.Number` branch canonicalizes *any* all-digit integer up to int64 via `unixNumber`

At/below 2^53 the two branches converge (shared `unixNumber` arithmetic), and `float64(1750478400500)` validates that. But **above 2^53 they fork**: a bare nanosecond epoch like `1750478400123456789` into a `DateTime64(9)` is (a) production-reachable, since that's the JSON type we actually decode, and (b) canonicalized rather than passed through — yet the oracle never checks it against ClickHouse. That's the one number class not already covered by the string digit-length gating, and it's precisely the "does our number reading match CH's" question this test exists to pin.

It very likely agrees (same `unixNumber` path), but "likely" against a live server is what we're trying to eliminate. Can you add `json.Number` entries to the corpus — including at least one `> 2^53` epoch into a `DateTime64` column — so the oracle exercises the type production actually sends? Happy to approve on that.

docs/src/content/docs/api.md
```diff
@@ -217,6 +217,17 @@ The body is a **flat JSON object** whose keys must match column names in the tar
...
... [SSE subscribers](#get-v1stream--server-sent-events-stream), the DLQ) sees the same spelling `/v1/query` renders. **Fail-open**: a value in none of those forms — or whose instant lies outside the column type's range, where ClickHouse *saturates* spelling-dependently instead of erroring — is published verbatim: ClickHouse's more liberal parser decides insertability, and a value it too rejects surfaces via the DLQ, as before. Pass-through likewise applies when a time zone doesn't resolve at runtime — the binary embeds no tzdata, so named zones resolve from the runtime's zone database (the bundled distroless images ship one; a stripped-down custom runtime, or a server zone newer than the image's snapshot, may not resolve): an unresolvable *column* zone skips canonicalization for that column entirely, an unresolvable *server* default skips only zone-less values of columns without a declared zone — warned at schema refresh either way, and never guessed as UTC, which could move the stored instant. The accepted grammar is differentially tested against a live ClickHouse (raw and canonicalized spellings must insert identically, or both fail). `Date`/`Date32` columns pass through untouched.
```
EricAndrechek: Nit (non-blocking): this "Timestamp canonicalization:" lead is accurate but has grown into a ~120-word sentence — the subject is separated from "is rewritten to one canonical wire form" by four input forms with nested em-dash/semicolon parentheticals. Consider pulling the accepted input forms into a short bulleted list (RFC 3339 / zone-less / 9–10-digit Unix string / integer number), mirroring the "canonical form, precisely" bullets right below it. Same content, much easier to scan.

CHANGELOG.md
```diff
@@ -50,6 +50,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
... zones resolve from the runtime's zone database, which the bundled distroless images ship) warns and passes through, never a failed refresh or a silent UTC reinterpretation that would move instants (`Etc/UTC` is special-cased to UTC); ingest never rejects a record over its timestamp spelling — fail-closed enforcement of the canonical form is the stream row-filter's (#381). The worker's `INSERT … FORMAT JSONEachRow` pins `date_time_input_format=best_effort` — the server default since ClickHouse 26.5, and on older servers the `basic` default rejects the canonical form's zone suffix (verified live); a superset of `basic` either way, so pre-upgrade messages still in NATS — zone-less strings, Unix numbers — parse exactly as before. Boundaries: `Date`/`Date32` columns pass through untouched (day precision, no zone ambiguity on this path); events published before the upgrade replay in their original spelling until the sweeper retires them; a column with an explicit non-UTC zone renders as `Z` on both paths (`/v1/query` likewise normalizes `DateTime` values to UTC), so the declared zone affects how zone-less inputs are read but never the output spelling — byte-identical regardless of zone.
```
EricAndrechek: Nit: "~35k generated inputs × **five** column shapes" — the integration test now has six shapes (`datetime`, `datetime_nyc`, `datetime64_3`, `datetime64_6_tokyo`, `datetime64_9`, `datetime64_0_utc`). Bump to six.

taitelee: Bumped to six. The sentence predated the `DateTime64(9)` / `DateTime64(0, 'UTC')` shapes.

docs/src/content/docs/api.md
taitelee: The lead sentence now states the rewrite up front, and the four accepted input forms are a bulleted list mirroring the "canonical form, precisely" bullets just below. Same content.

tests/integration/timestamp_canonicalization_test.go
taitelee: Ok the test now includes entries alongside the float64 ones. There's integer seconds, a non-integer, and the 19-digit ns epoch above 2^53 (which rewrites only on `DateTime64(9`), everywhere else it's out of range and passes through).

docs/src/content/docs/api.md
```diff
@@ -217,7 +217,14 @@ The body is a **flat JSON object** whose keys must match column names in the tar
...
+
+- RFC 3339, any offset (`.`-fractions only — ClickHouse has no `,` separator).
+- `YYYY-MM-DD[ T]HH:MM:SS[.fff]` or `YYYY-MM-DD`, zone-less — interpreted in the column's time zone, else the ClickHouse server's, exactly as ClickHouse itself would.
+- A Unix-seconds string of exactly 9–10 digits (a `.fff` fraction is honored only for `DateTime64` columns, as ClickHouse does). Other digit lengths are ClickHouse's own forms — calendar shapes like `YYYYMMDD`, or its 13/16/19-digit ms/µs/ns epochs — so they pass through untouched.
+- An **integer** JSON number — read the way ClickHouse reads numbers: Unix seconds for `DateTime`, **ticks at the column's scale** for `DateTime64` (`1750478400500` into a `DateTime64(3)` is the millisecond epoch). Non-integer numbers pass through; ClickHouse rejects them.
```
EricAndrechek: Why to `Non-integer numbers pass through; ClickHouse rejects them.`?

EricAndrechek (dismissed):

EricAndrechek:
**Live confirmation of the pre-26.5 case, on a production tenant — this is actively losing data today.**

The PR notes that on servers older than ClickHouse 26.5 the `basic` default rejects the canonical form's zone suffix (verified live). Hit exactly that on a **WaveHouse Cloud** tenant while validating an unrelated control-plane change, before finding this PR. Adding the reproduction as corroboration for merge urgency — every WHC tenant cell currently runs **ClickHouse 24.8.14.39**, i.e. the pre-default case.

Engine `ghcr.io/wave-rf/wavehouse:dev-774faec013649f2adad6646ad6eb20bb08cada13`, table:

```sql
CREATE TABLE gh_events (
  event_id String, event_type LowCardinality(String), event_ts DateTime64(3, 'UTC')
) ENGINE = ReplacingMergeTree ORDER BY (event_type, event_ts, event_id)
```

Ingesting the format `/v1/query` itself emits:

```
POST /v1/ingest?table=gh_events
{"event_id":"evt-1","event_type":"star","event_ts":"2026-08-05T18:00:00.123Z"}
→ 200 {"total":2,"succeeded":2,"failed":0,"duplicates":0}
```

Then, ~5s later:

```
WARN  bulk insert failed, falling back to 1-by-1 isolation  component=ingest_worker table=gh_events
      error="HTTP 400: Code: 27. DB::Exception: Cannot parse input: expected '\"' before:
      'Z\",\"event_type\"...': (while reading the value of key event_ts) (at row 1)
      ... (CANNOT_PARSE_INPUT_ASSERTION_FAILED)"
ERROR isolated bad row, sending to DLQ  component=ingest_worker table=gh_events
```

`SELECT count()` → **0**. Same row with `"2026-08-05 18:02:00.750"` persists fine, and reads back as `"2026-08-05T18:02:00.750Z"` — the string that fails.

Two things worth underlining for reviewers:

1. **The failure is invisible to the caller.** Single-row ingest answers `{"ok":true}`; a batch answers `succeeded:2, failed:0`. The documented "branch on `failed`, not status" guidance doesn't help — the rejection happens after the response, in the worker. Only the DLQ and a log line show it.
2. **The idiomatic input is the failing one.** RFC 3339 with `Z` is what the engine emits, what `toISOString()` produces, and what any SDK will naturally send.

Both of which this PR fixes. No change requested — just evidence that the `best_effort` pin is load-bearing for deployments on the current CH version, not only a tidiness fix.

docs/src/content/docs/api.md
taitelee: ClickHouse has lenient datetime parsing (`date_time_input_format=best_effort`). It applies only to quoted strings. A bare JSON number is read by an integer-only parser (Unix seconds for `DateTime`, raw ticks for `DateTime64`). So for a timestamp `ts: 1750478400.5`, the reader consumes let's say `1750478400`, stops at the `.`, and the leftover `.5` fails the row's delimiter check — rejected for every timestamp column type.

So based on that, pass-through is the only option that keeps the #372 invariant. Rewriting `1750478400.5` → `"2025-06-21T04:00:00.5Z"` would have the canonicalizer accepting a row ClickHouse rejects: raw value fails the insert, rewritten value succeeds. The differential test pins exactly this — for each corpus value it inserts the raw and the canonicalized spelling side by side and requires both to succeed or both to fail — so that rewrite would fail the oracle on its `json.Number("1750478400.5")` entry. A non-integer number therefore keeps the producer's spelling, gets ClickHouse's accept-or-reject decision, and surfaces via the DLQ, as pre-#372.

Also reworded the bullet so it carries the mechanism (and the ticks thing).

docs/src/content/docs/development.md
```diff
@@ -428,16 +428,20 @@ WaveHouse/
 │   ├── pipes/              # Named query pipes (NATS KV + .sql bootstrap)
 │   ├── policy/             # Access control policies (evaluation + NATS KV store)
 │   ├── query/              # Structured query AST + SQL builder
+│   ├── stream/             # SSE fan-out: Hub, Subscriber queue, Bucket, keepalive wheel
```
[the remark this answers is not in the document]
taitelee: The tree entry is correct. Hub lives in internal/stream/hub.go on this branch. The #294 changelog entry higher up documents the move (it removed internal/api/hub.go); the line-59 entry you're citing is the earlier #226 keepalive entry, written before the move, describing it as future work. Changelog entries record history at the time they landed, so that phrasing stays.

docs/src/content/docs/architecture.md
```diff
@@ -113,13 +113,14 @@ The SSE fan-out, factored out of `api/` so the delivery hot path ([#294](https:/
...
+- **timestamp.go** — `CanonicalizeTimestamps(schema, data)` rewrites every parseable value in a top-level `DateTime`/`DateTime64` column to the canonical RFC 3339 UTC wire form before the event is published ([#372](https://github.com/Wave-RF/WaveHouse/issues/372)): zone-less values are interpreted in the column's declared zone, else the discovered server default — ClickHouse's own rule, so the spelling changes but never the instant. Fail-open: an unparseable value or unresolvable zone passes through verbatim for ClickHouse's own parser to judge; ingest never rejects a record over its timestamp spelling.
```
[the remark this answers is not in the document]
taitelee: All three points trace to learnings recorded against a previous commit, an intermediate design this PR later reversed. I dropped the `time/tzdata` embed and made zone resolution fully fail-open. I also extended that (Etc/UTC without `tzdata`). In the current code, no `time/tzdata` import exists in the tree; an unresolvable server zone warns and leaves zone-less values un-canonicalized (in `discovery.go:83-90`, refresh continues; only the SELECT `timezone()` query error itself fails refresh, which is ClickHouse-unreachable, not zone resolution); an unresolvable column zone warns and stores no spec: pass-through, never a per-record rejection. There is no refresh-failing or per-record-rejecting tier to document. The docs describe the shipped behavior accurately, so no change is needed. @coderabbitai please update the PR #402 learnings from 2026-07-09 that describe the embedded-tzdata / fail-refresh design — they're superseded.

EricAndrechek (approved):

PR #414, opened 2026-07-20
feat(config): hot-reload whitelisted fields via SIGHUP or admin endpoint

### LINKED ISSUE #48: Hot-Reloading Configuration
> **Status (2026-08-20):** being delivered via the settings-directory design — validation landed in PR #500; boot loading + key migration, then reload triggers, follow. JSON Schema follow-up in #506. See the status comment below for the full sequencing.

## Problem

Changing a log level or a metric interval currently requires a full application restart, causing a small window of downtime and re-sharding in NATS.

## Proposed Solution

Implement a watcher for the configuration file or handle the SIGHUP signal. When triggered, the application will re-parse the config and update internal components (like the LogLevel variable) in real-time without dropping active connections.

## Alternatives Considered

Easy endpoint to change LogLevel using a pointer to point at the current log level and change it via that in the logger. This is only for log level though. Probably not the ideal way to approach this.

## Additional Context

Any mockups, examples, or references that help explain the request.

taitelee:
## Summary

Adds config hot reload: `SIGHUP` or the new admin-gated `POST /v1/admin/config/reload` (both serialize on one code path) re-runs the full config load — file, `WH_*` environment overrides, validation — and applies an explicit whitelist to the running process, currently `dedupe.id_field` and `dedupe.require_id`. The ingest handler reads them through an atomically swapped snapshot: one lock-free load per record, so a reload lands at a record boundary and never mixes old and new settings within one record.

Every other changed section is reported as `restart_required` (reload log line + endpoint response) instead of silently ignored. A file that fails to parse or validate changes nothing — the previous config stays active. `dedupe.enabled` stays restart-only (it owns the Pebble store lifecycle), env-pinned keys honestly report no change, and a drift-guard test forces every future config field to be classified hot or restart-only.

## Related Issues

Closes #48 — delivers the reload mechanism (SIGHUP + admin endpoint) with an explicit hot-field whitelist; log-level hot reload split out to #415. Groundwork for the per-table dedupe overrides of #222.


taitelee:
`RestartRequired` is technically unnecessary apart from logging. Making note to bring up in discussion.

taitelee:
Closing as superseded by the settings-directory design that starts in #500.

Issue #475, opened 2026-08-13
npm: the dev channel's versions don't order by recency, so a range resolves to a two-month-old build


EricAndrechek:
**Area:** infra / npm publishing — correctness of the published version graph · related to #268 (which describes the content-addressed `dev` channel as designed)

**Expected:** installing from the `dev` channel gets you a recent build, and keeps getting you recent builds.

**Actual:** it can silently pin a build from two months ago, because **the `dev` channel's version numbers do not order by recency.**

Every dev publish is `0.0.0-dev.h<dist-hash>` (#268: *"every push to `main` builds the SDK and publishes `0.0.0-dev.h<dist-hash>` to the `dev` dist-tag"*). A content hash is deliberate and reasonable for identity — but semver compares alphanumeric prerelease identifiers **lexically**, so a hash makes the ordering arbitrary with respect to time.

Today's published set, in publish order:

```
2026-06-04   0.0.0-dev.0f8826c          ← also the `latest` dist-tag
2026-06-04   0.0.0-dev.h7d02b5f9a5e4
2026-06-05   0.0.0-dev.h825a5d7fa405
2026-06-05   0.0.0-dev.h97a11a9f5a0e
2026-06-08   0.0.0-dev.hab0797ca5a74
2026-06-08   0.0.0-dev.hff4a2a48d0da   ← SEMVER-HIGHEST of all ten
2026-06-08   0.0.0-dev.h514be5522d89
2026-06-12   0.0.0-dev.h8ea15a023aa2
2026-08-11   0.0.0-dev.hb0c3a577b2f8
2026-08-12   0.0.0-dev.h04a85f20bb96   ← current `dev` dist-tag
```

The newest build sorts **sixth**. A June 8 build is the highest version in the package. `hff…` > `h04…` lexically, and that is the whole story.

### Why that bites in practice

Installing from the tag resolves correctly *at that moment* — but what gets recorded in a consumer's `package.json` is a **range**, not the tag. `^0.0.0-dev.h04a85f20bb96` means `>=0.0.0-dev.h04a85f20bb96 <0.0.1`, and `0.0.0-dev.hff4a2a48d0da` satisfies it. The next resolution takes the highest match, which is the June 8 build. The dist-tag pointer is correct and gets thrown away one step later.

[...]

PR #479, opened 2026-08-17
refactor(api)!: merge all admin-gated endpoints under /v1/ops

taitelee:
## Summary

**BREAKING**: every admin-gated endpoint now lives under a single `/v1/ops/*` prefix, behind one tree-level `RequireAdmin` gate. Previously the admin surface was split between the `/v1/admin/*` subtree (raw SQL, policy CRUD, pipe CRUD) and three individually-gated top-level routes.

- `/v1/admin/{query,policy,pipes...}` → `/v1/ops/{query,policy,pipes...}`
- `/v1/schema` and `/v1/schema/refresh` → `/v1/ops/schema[/refresh]`
- `/v1/dlq/stats` → `/v1/ops/dlq/stats`

No aliases kept (pre-1.0); old paths return `404`. Gate semantics unchanged (`policy.admin_role`, operator-key break-glass included). SDK path constants updated in lockstep — its method surface is unaffected, but an older SDK 404s against a renamed server. Docs swept throughout, including a new "Fencing the admin surface" reverse-proxy section (one proxy rule now covers the whole management surface).


taitelee:
@coderabbitai review

taitelee:
@coderabbitai full review

taitelee:
@coderabbitai review

internal/api/router_test.go
```diff
@@ -313,17 +321,24 @@ func TestNewRouter_RoutesRegistered(t *testing.T) {
...
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.method+" "+tt.path, func(t *testing.T) {
 			t.Parallel()
-			req := httptest.NewRequestWithContext(context.Background(), tt.method, tt.path, nil)
+			// The admin role is inert on the public rows (no gate reads it)
+			// and lets the /v1/ops rows through to their handlers.
+			ctx := auth.WithRole(context.Background(), "admin")
+			req := httptest.NewRequestWithContext(ctx, tt.method, tt.path, nil)
```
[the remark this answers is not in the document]
taitelee: Added the two missing rows /v1/admin/policy and /v1/admin/pipes now join /v1/admin/query, /v1/schema, /v1/schema/refresh, and /v1/dlq/stats, so every removed path is pinned to 404. One deliberate deviation from the suggested shape: the rows stay roleless rather than running as admin. Roleless is the stronger polarity — an accidentally re-registered alias fails the row whether it comes back gated (403 ≠ 404) or, the real hazard, ungated (200 ≠ 404) — and it additionally proves the row carries no gate at all. The test's comment block documents this.

docs/src/content/docs/sdk/admin.md
```diff
@@ -4,11 +4,24 @@ description: "Schema introspection, access-control policy, DLQ stats, and health
...
 
 Operational surfaces of `@wavehouse/sdk`. Everything here except
-`wh.sys.health()` requires the admin role (`policy.admin_role`) — see
-[Access Control](/access-control) for how roles resolve.
+`wh.sys.health()` sits behind the server's admin gate: the caller must
+resolve to the admin role (`policy.admin_role`) or present the non-JWT
+[operator key](/api#authentication) — the SDK has no first-class
+operator-key option, but [`options.headers`](/sdk#custom-headers) can carry
+the `X-Operator-Key` header. See [Access Control](/access-control) for how
+roles resolve.
```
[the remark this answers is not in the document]
taitelee: Bruh. Please keep your comments to the scope of this PR. We will be doing a doc review again in another scoped PR. Partly declined, partly adopted. The contradiction described here didn't exist: the sentence reads "Everything here except `wh.sys.health()` sits behind the server's admin gate" — line 7 in isolation is the middle of that wrapped sentence, and the proposed deletion target ("`wh.sys.health()` sits behind the server's admin gate.") never appears in the file. §System on the same page already documents the public route. That said, the misparse is fair evidence the exception was easy to skim past, so the intro now states the gate rule first and gives the exception its own sentence — "The exception is `wh.sys.health()`, which calls the public, content-free /v1/health route and needs no credentials" — which also surfaces the no-credentials fact in the intro instead of only in §System.

clients/ts/README.md
```diff
@@ -10,6 +10,8 @@ npm install @wavehouse/sdk
 
 Requires Node 22 or newer — the only line this SDK is tested against; Node 18 and 20 are past upstream end-of-life. Browsers need no polyfill, and Node needs an `EventSource` polyfill only for streaming — see [Runtime support](https://wavehouse.dev/sdk/#runtime-support).
 
+Until the next tagged release, the server's admin surfaces (raw SQL, schema, policy, DLQ, pipes admin, codegen) live at `/v1/ops/*` and only the `dev` build targets them — `npm i @wavehouse/sdk@dev` alongside a server built from `main`. See [Versioning](#use-without-a-build-step-cdn) below.
+
```
EricAndrechek: Why add this?

clients/ts/README.md
```diff
@@ -30,7 +32,7 @@ No bundler, no `npm`, no framework required — drop the SDK straight into an HT
 </script>
 ```
 
-Pin a version for production (`https://esm.sh/@wavehouse/sdk@0.1.0`). jsDelivr (`https://cdn.jsdelivr.net/npm/@wavehouse/sdk/+esm`) and unpkg (`https://unpkg.com/@wavehouse/sdk?module`) serve the same ES module.
+Pin a version for production once the first tagged release exists (`https://esm.sh/@wavehouse/sdk@<version>` — see Versioning below). jsDelivr (`https://cdn.jsdelivr.net/npm/@wavehouse/sdk/+esm`) and unpkg (`https://unpkg.com/@wavehouse/sdk?module`) serve the same ES module.
```
EricAndrechek: Undo/delete this change

clients/ts/README.md
```diff
@@ -42,7 +44,7 @@ Pin a version for production (`https://esm.sh/@wavehouse/sdk@0.1.0`). jsDelivr (
 </script>
 ```
 
-**Versioning.** A bare CDN URL serves the latest published **release**; pin for production (`@wavehouse/sdk@0.1.0`) or float on a range (`@0` for the newest 0.x, `@0.1` for 0.1.x). Builds from `main` are published under the `dev` tag — `@wavehouse/sdk@dev` — for trying unreleased changes.
+**Versioning.** A bare CDN URL serves the `latest` dist-tag — today a `0.0.0-dev.*` snapshot, since no tagged release exists yet; once one does, pin for production (`@wavehouse/sdk@<version>`) or float on a range (`@0` for the newest 0.x). Builds from `main` are published under the `dev` tag — `@wavehouse/sdk@dev` — for trying unreleased changes. Note the current path skew: the `latest` snapshot predates the server's `/v1/admin` → `/v1/ops` rename, so its admin surfaces (raw SQL, schema, policy, DLQ, pipes admin, codegen) get `404` from a renamed server — use `@wavehouse/sdk@dev` alongside a server built from `main`.
```
EricAndrechek: Undo/delete this change

clients/ts/README.md
```diff
@@ -144,13 +146,13 @@ if (error) {
 
 ## Codegen
 
-Generate TypeScript types from your live WaveHouse schema. The package ships a `wavehouse-codegen` bin, so after installing you can run it with `npx`:
+Generate TypeScript types from your live WaveHouse schema. The package ships a `wavehouse-codegen` bin — until the first tagged release, select the `@dev` build explicitly:
```
EricAndrechek: Undo/delete this change

clients/ts/README.md
```diff
@@ -144,13 +146,13 @@ if (error) {
...
 
 ```bash
-npx wavehouse-codegen --url http://localhost:8080 --out ./src/db.d.ts
+npx --package @wavehouse/sdk@dev wavehouse-codegen --url http://localhost:8080 --out ./src/db.d.ts
 ```
 
-This introspects `/v1/schema`, maps ClickHouse column types to TypeScript, and outputs a `Database` interface you can pass to `createClient<Database>()`. `/v1/schema` is **admin-only** — pass an admin-role token with `--auth <jwt>` (or `-a`) against any non-dev policy.
+This introspects `/v1/ops/schema`, maps ClickHouse column types to TypeScript, and outputs a `Database` interface you can pass to `createClient<Database>()`. `/v1/ops/schema` is **admin-only** — pass an admin-role token with `--auth <jwt>` (or `-a`) against any non-dev policy. The `/v1/ops/*` paths are unreleased on both sides: until the next release, the published `latest` build (a pre-rename `0.0.0-dev.*` snapshot) targets the old `/v1/admin`-era paths and gets `404` from a renamed server — use the `@dev` build (`npm i @wavehouse/sdk@dev`) alongside a server built from `main`.
```
EricAndrechek: Remove the bit at the end:

```suggestion
This introspects `/v1/ops/schema`, maps ClickHouse column types to TypeScript, and outputs a `Database` interface you can pass to `createClient<Database>()`. `/v1/ops/schema` is **admin-only** — pass an admin-role token with `--auth <jwt>` (or `-a`) against any non-dev policy.
```

docs/src/content/docs/sdk/admin.md
```diff
@@ -3,12 +3,26 @@ title: "SDK Admin & System"
...
 
+:::caution[Admin paths moved to `/v1/ops/*`]
+Server-side, these endpoints moved from `/v1/admin/*` and the top-level
+`/v1/schema`/`/v1/dlq/stats` to `/v1/ops/*` with no aliases — upgrade
+`@wavehouse/sdk` alongside the server, or every admin call on this page
+returns `404` against a renamed server. The ops-path client is unreleased —
+see the [Installation note](/sdk#installation) for the `@dev`-tag build
+(`npm i @wavehouse/sdk@dev`) that targets the new paths.
+:::
+
```
EricAndrechek: We haven't released the tagged version yet, so this breaking change doesn't need to be documented so loudly here, just remove all this.

docs/src/content/docs/sdk/index.mdx
```diff
@@ -49,19 +49,34 @@ import { Tabs, TabItem, LinkCard, CardGrid } from "@astrojs/starlight/components
     </script>
     ```
 
-    Pin a version for production (`https://esm.sh/@wavehouse/sdk@0.1.0`);
+    Pin a version for production (`https://esm.sh/@wavehouse/sdk@<version>`,
+    once the first tagged release exists — see
+    [Releasing the SDK](/development#releasing-the-sdk));
```
EricAndrechek: Undo/delete this change

docs/src/content/docs/sdk/index.mdx
```diff
@@ -49,19 +49,34 @@ import { Tabs, TabItem, LinkCard, CardGrid } from "@astrojs/starlight/components
...
     (`WaveHouse.createClient({ … })`) for a classic `<script src>` tag.
     Streaming uses the browser's native `EventSource`, so it needs no polyfill
-    either way. A bare CDN URL tracks the latest published release; use a
-    range (`@0`, `@0.1`) to float within a major or minor, or the `@dev` tag
-    for unreleased builds from `main` (see
-    [Releasing the SDK](/development#releasing-the-sdk)).
+    either way. A bare CDN URL tracks the `latest` dist-tag — today a
+    `0.0.0-dev.*` snapshot, a tagged release once one exists; after that
+    first release, use a range (`@0`, `@0.1`) to float within a major or
+    minor, or use the `@dev` tag for unreleased builds from `main`.
```
EricAndrechek: Undo/delete this change

docs/src/content/docs/sdk/index.mdx
```diff
@@ -49,19 +49,34 @@ import { Tabs, TabItem, LinkCard, CardGrid } from "@astrojs/starlight/components
...
+build targeting the new paths is unreleased: against a renamed server, the
+currently published build (the `latest` dist-tag — today a pre-rename
+`0.0.0-dev.*` snapshot; see
+[Releasing the SDK](/development#releasing-the-sdk)) returns `404` on every
+admin surface — `wh.sql()`, `wh.schema`, `wh.from(t).schema()`, `wh.policy`,
+`wh.dlq`, the `wh.pipes` admin methods, and `npx wavehouse-codegen`. Until
+the next release, install from the `@dev` tag (`npm i @wavehouse/sdk@dev`)
+alongside a server built from `main`.
+:::
+
```
EricAndrechek: Undo/delete this change

docs/src/content/docs/sdk/index.mdx
```diff
@@ -108,7 +123,9 @@ each runtime.
     import { createClient } from 'https://esm.sh/@wavehouse/sdk';
     ```
 
-    Pin a version for production: `https://esm.sh/@wavehouse/sdk@0.1.0`.
+    Pin a version for production once the first tagged release exists
+    (`https://esm.sh/@wavehouse/sdk@<version>` — see
+    [Releasing the SDK](/development#releasing-the-sdk)).
```
EricAndrechek: Undo/delete this change

docs/src/content/docs/sdk/pipes.md
```diff
@@ -47,7 +47,11 @@ Open a live stream. See [Streaming](/sdk/streaming).
 
 ## Pipes Admin — `wh.pipes`
 
-Manage named query pipes. Requires the admin role (`policy.admin_role`).
+Manage named query pipes. Requires the admin gate — the admin role
+(`policy.admin_role`) or the [operator key](/api#authentication).
+These methods hit `/v1/ops/pipes*`; the client build targeting `/v1/ops` is
+unreleased — see the [Installation note](/sdk#installation) for the
+`@dev`-tag build.
```
EricAndrechek: I like the change to add that it requires the operator key above, but this bit about the ops path being unreleased can/should be deleted.
```suggestion
```

docs/src/content/docs/sdk/queries.md
```diff
@@ -69,7 +69,13 @@ await clicks.insertNDJSON(await openAsBlob('events.ndjson'));
...
 ### `.schema(opts?)`
 
-Fetch the table's column definitions from ClickHouse.
+Fetch the table's column definitions from ClickHouse. `.schema()` hits
+`/v1/ops/schema`, an **admin-only** endpoint: the caller must pass the admin
+gate — resolve to the policy admin role or present the non-JWT
+[operator key](/api#authentication) via [`options.headers`](/sdk#custom-headers)
+— or this returns `403`. The
+client build targeting `/v1/ops` is unreleased — see the
+[Installation note](/sdk#installation) for the `@dev`-tag build.
```
EricAndrechek: Remove the bit about this being unreleased:
```suggestion
— or this returns `403`.
```

docs/src/content/docs/sdk/queries.md
```diff
@@ -255,7 +261,7 @@ while (result.hasMore && result.next) {
 
 ## Raw SQL — `wh.sql(query, opts?)`
 
-Execute a raw SQL query. `/v1/admin/query` is admin-only: the caller's JWT must resolve to the policy admin role (`admin_role`, `"admin"` by default). A request with no token, or an invalid/expired one, falls back to the `default_role` and is rejected.
+Execute a raw SQL query. `/v1/ops/query` is admin-only: the caller must resolve to the policy admin role (`admin_role`, `"admin"` by default). A tokenless request falls back to the `default_role`, so it is rejected with `403` on any policy that doesn't deliberately set `default_role` to the admin role (a loudly-warned dev-only setting); an invalid or expired token is rejected with `401`. The SDK has no first-class option for the server's non-JWT [operator key](/api#authentication) — an operator can send its `X-Operator-Key` header via [`options.headers`](/sdk#custom-headers). The client build targeting `/v1/ops` is unreleased — see the [Installation note](/sdk#installation) for the `@dev`-tag build.
```
EricAndrechek: Remove the bit about it being unreleased:
```suggestion
Execute a raw SQL query. `/v1/ops/query` is admin-only: the caller must resolve to the policy admin role (`admin_role`, `"admin"` by default). A tokenless request falls back to the `default_role`, so it is rejected with `403` on any policy that doesn't deliberately set `default_role` to the admin role (a loudly-warned dev-only setting); an invalid or expired token is rejected with `401`. The SDK has no first-class option for the server's non-JWT [operator key](/api#authentication) — an operator can send its `X-Operator-Key` header via [`options.headers`](/sdk#custom-headers).
```

docs/src/content/docs/sdk/reference.md
```diff
@@ -94,16 +94,16 @@ StreamController (NOT thenable)
 
 ## Codegen CLI
 
-Generate TypeScript types from a running WaveHouse instance. The package ships a `wavehouse-codegen` bin, so after installing `@wavehouse/sdk` you can run it with `npx`:
+Generate TypeScript types from a running WaveHouse instance. The package ships a `wavehouse-codegen` bin — until the first tagged release, select the `@dev` build explicitly (the published `latest` CLI still targets the removed pre-`/v1/ops` paths and gets `404`s; see the [Installation note](/sdk#installation)):
```
EricAndrechek: Undo/delete this change

docs/src/content/docs/api.md
```diff
@@ -601,14 +601,20 @@ curl -N "http://localhost:8080/v1/stream?table=clicks&since=2026-03-24T11:00:00Z
...
 
-:::note[Admin only]
-The schema and DLQ endpoints in this section require the `admin_role` (like [`/v1/admin/query`](#post-v1adminquery--query-clickhouse)); other callers get 401 (bad token) / 403. The quickstart's trial `public` role cannot call them.
+:::note[Renamed from `/v1/admin/*`]
+These endpoints previously lived at `/v1/admin/*` (raw SQL, policy, pipes) and at the top level (`/v1/schema`, `/v1/schema/refresh`, `/v1/dlq/stats`). They all moved under `/v1/ops/*` in a single pre-1.0 rename with **no aliases kept** — a request to an old path returns `404 {"error":"not found"}`. See the [CHANGELOG](https://github.com/Wave-RF/WaveHouse/blob/main/CHANGELOG.md) for the full old-path → new-path mapping. If you fenced `/v1/admin/` at your reverse proxy, that rule silently stops matching — move it to `/v1/ops/` (see [Fencing the admin surface](/reverse-proxy#fencing-the-admin-surface)).
```
EricAndrechek: I don't think this is necessary since we are pre-release so no one needs docs for this "breaking" change

docs/src/content/docs/api.md
```diff
@@ -809,12 +810,13 @@ The message format used on NATS JetStream between ingest and the batch consumer:
 | Field | Type | Description |
 | ----- | ---- | ----------- |
 | `table_name` | string | Target ClickHouse table (from URL). |
+| `scope` | string | Optional cache-invalidation scope, omitted when empty. The ingest handler currently always publishes it empty (see the TODO in `internal/api/ingest.go`); consumers must tolerate its presence. |
```
EricAndrechek: Remove this until we actually properly support it

docs/src/content/docs/api.md
```diff
@@ -830,9 +832,9 @@ Same as the wire format — events are passed through directly:
...
+When a batch insert to ClickHouse fails (e.g., type errors, connection issues), the worker re-inserts the batch row by row: rows that succeed are acked, and only the rows that fail again are published to the DLQ NATS stream (`WAVEHOUSE_DLQ`) under subjects `dlq.{table}`. This prevents infinite retry loops — those messages are ACKed from the main stream and moved to the DLQ for inspection. The DLQ message body is the published `EventMessage` envelope (`{"table_name":…,"received_timestamp":…,"data":{…}}` — as published, so it would also carry `scope` if that were ever set; the failed row is under its `data` key, its `DateTime`/`DateTime64` values as published: canonicalized where WaveHouse could parse them, otherwise the producer's original spelling — see [timestamp canonicalization](#timestamp-canonicalization)); the failure reason, table, and time travel in the `X-DLQ-Table` / `X-DLQ-Error` / `X-DLQ-Timestamp` message headers.
```
EricAndrechek: Don't write about scope until we actually have a path for it to be set and used.

docs/src/content/docs/architecture.md
```diff
@@ -120,8 +120,8 @@ The SSE fan-out, factored out of `api/` so the delivery hot path ([#294](https:/
...
+- **types.go** — `EventMessage` struct (TableName, Scope, ReceivedTimestamp, Data) and `BufferConsumerName` constant, shared across API handlers and the ingest pipeline.
```
EricAndrechek: Same thing here, don't bother adding scope yet until we actually use it

docs/src/content/docs/deployment.md
```diff
@@ -81,15 +81,25 @@ Production images are published to GitHub Container Registry via GoReleaser:
 ghcr.io/wave-rf/wavehouse:<tag>
 ```
 
-Published images carry a signed [Sigstore](https://www.sigstore.dev/) build-provenance attestation (stored in the registry). Verify one before deploying:
+Today only `:dev` (rolling `main`-branch build) and `:dev-<full-commit-sha>` (immutable, pruned after 30 days) are published; `:vX.Y.Z` and `:latest` arrive with the first `v*` release tag. To pin (see the [alpha-stage caution](https://github.com/Wave-RF/WaveHouse#-project-status) in the README), use a `:dev-<full-commit-sha>` tag — the full 40-character commit SHA, not the short form — or an image digest.
```
EricAndrechek: I like this, it can be kept, but please:
- [x] update #475 to make sure it reminds us to change things in the docs here too, and
- [x] remove the language here saying today only `:dev` etc are published – write it so that it reads correctly assuming we do publish a `:latest` right after this is merged

docs/src/content/docs/ingest-pipeline.md
```diff
@@ -20,10 +20,12 @@ goroutine / channel / timer interplay is subtle.
 | `types.go` | `EventMessage` wire format and the `BufferConsumerName` constant |
 
 The pipeline is **insert-only**. The wire format carries
-`{table_name, received_timestamp, data}` and nothing else; the worker parses
+`{table_name, scope, received_timestamp, data}` — `scope` is an optional
+cache-invalidation scope that the ingest handler currently always publishes
+empty (see the TODO in `internal/api/ingest.go`). The worker parses
```
EricAndrechek: Just omit the scope stuff until it is properly wired in/through

docs/src/content/docs/reverse-proxy.mdx
```diff
@@ -197,6 +197,14 @@ WaveHouse does **not** derive a client IP from forwarded headers — it does no
...
+
+:::caution[Upgrading: fences written against the old paths]
+This surface used to be split across `/v1/admin/` plus the top-level `/v1/schema`, `/v1/schema/refresh`, and `/v1/dlq/stats`. No aliases were kept, so a proxy rule written against any of those paths — e.g. `location /v1/admin/ { deny all; }` — now matches nothing, and the surface it fenced is back within unauthenticated reach on the public vhost (the server-side `RequireAdmin` gate still denies non-admin callers). Re-point the rule at `/v1/ops/`.
+:::
+
```
EricAndrechek: Same thing here, don't think we need this since this is being changed pre released version so no one experiences these changes as breaking

internal/ingest/worker_test.go
EricAndrechek, line 1: What warranted/lead to the changes that were made in this file?

README.md
EricAndrechek, line 1: Undo all changes to this file

EricAndrechek (changes requested):
Good, relatively small PR/changes – super minor tweak to router which required tweaks in the SDK and lots of test cases which are all fine. Docs changes also were required as a result of the change, but many (likely due to the long cycling docs review pre push claude hooks) overstepped and added more to the docs than we want, so basically the entirety of these requested changes are just stepping back what was made.

clients/ts/README.md
taitelee: Removed this actually because npm builds published already predate the path rename so a plain install gets a client that 404s against a server built from main. Just had a warning here about that. But it is loud so removed. Not sure if it necessary right now and it was something Claude had added.

internal/ingest/worker_test.go
taitelee: This test went red on this PR's CI without the rename touching anything it covers. It proves table B flushes on its own size trigger instead of getting stranded behind table A waiting on the `maxWait` timer, and it demanded the whole publish, consume, and flush cycle finish within 1.75 seconds of wall clock to tell the two apart, which a busy runner can miss for reasons unrelated to the regression. The fix just widens the margins, `maxWait` goes to 30 seconds with a 10 second assertion window, so a size flush passes comfortably and a stranded batch still fails loudly. There is a changelog entry under Fixed with the same story. Do you want to take a look and see if you think it should be it's own separate PR or something because without the change I had failed a CI run once. Also not sure if GitHub's incident yesterday could have affected it but I don't think so.

docs/src/content/docs/deployment.md
taitelee: Kept the pinning guidance and dropped the publish status framing, so it now just says what each tag is and reads correctly once a latest exists. The compose health check example further down is back on latest too. And there is a note on #475 now flagging that this Registry section and the README quick start need a pass when the release path changes.

taitelee:
@coderabbitai review

internal/ingest/worker_test.go
EricAndrechek: Yeah, I think this is the only thing remaining I want tweaked – at the very least, this test needs to assert that they have uploaded BEFORE the (currently unconfigurable 5s) batch insert bound time to confirm that they are uploading faster than that when the batch size is hit, which this is supposed to do I believe. I agree that the 2 second value could be tight on a contested runner, but it should still be asserted to pass before the 5s batch time at the very least I think. Can we see if that still passes?

EricAndrechek (changes requested):

internal/ingest/worker_test.go
taitelee: ok the window is `defaultMaxWait` (the 5s constant from `worker.go`, same package so it's referenced directly and tracks the bound if it ever changes), so the test asserts the size-trigger flush beats the timer production actually runs. I kept the test's own `maxWait` time at 30s so a stranded batch still can't sneak in under the window.

EricAndrechek (approved):

taitelee:
@coderabbitai review

taitelee:
@coderabbitai review

PR #500, opened 2026-08-19
feat(settings): add settings-directory validation and validate CLI

taitelee:
## Summary

First piece of the file-based control plane: settings move to a directory of JSON documents (`roles.json`, `policies.json`, `pipes.json`, `config.json`), and `settings.Validate(dir)` is the single gate every consumer of that directory runs — the CLI now, boot and live reload when they land.

- New `internal/settings` package: pure validation (no network, no ClickHouse), collecting all findings in one pass — directory contents (exactly the four files, unknown `*.json` rejected), strict JSON decoding (unknown fields, duplicate keys, truncated files), per-file sha and cross-file role referential integrity. Warnings don't invalidate.
- New `wavehouse validate [dir]` subcommand: directory from the argument or `WH_SETTINGS_DIR`, prints findings, exits 0/1/2 (valid/invalid/usage) operators and CI can gate a config change before it reaches a running instance.
- `settings.dir` / `WH_SETTINGS_DIR` added to boot config as the reload machinery will follow.

## TODOs (follow-up PRs)

- In-memory snapshot: a loader that holds the current good document and swaps it atomically on successful validation.
- Boot integration: load from `WH_SETTINGS_DIR` at startup, refuse to start loudly on invalid settings, and repoint runtime consumers (roles, policy, pipes, tenant tunables) at the snapshot.
- Reload trigger: `/v1/ops/reload` and/or SIGHUP and/or fsnotify with debounce; a failed reload keeps serving the previous good document.
- Decide the fate of the existing `/v1/ops` settings write endpoints (reverses the #48-era runtime-settings direction).
- Docs for the settings directory and the `validate` command once loader lands.

NOTE: The e2e bullet reflects what the exclude actually says in .testcoverage.yml: the package is only reachable via the CLI today, the e2e suite is server-focused so it sat at 0% and pulled that suite's gate under its floor, unit coverage is thorough, the merged total is untouched, and the exclude is explicitly temporary until the server consumes the directory.

## Related Issues

Advances #48 (validation gate only — reload wiring lands separate


internal/settings/settings.go
```diff
@@ -0,0 +1,108 @@
...
+	FileRoles    = "roles.json"
+	FilePolicies = "policies.json"
+	FilePipes    = "pipes.json"
+	FileConfig   = "config.json"
+)
+
+// Files lists every expected file. All four must exist (an empty document is
+// written as {}), so an accidental deletion reads as an error, never as
+// "defaults".
+var Files = []string{FileRoles, FilePolicies, FilePipes, FileConfig}
```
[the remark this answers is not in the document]
taitelee: Files and `pipeParamTypes` are immutable lookup tables, the same idiom this codebase already uses throughout internal/ (`mutationVerbs`, `nonMutationVerbs`, `identEscaper`, `intBounds`, package-level regexes). The "no global state" guideline is about injected dependencies and singletons, not constant tables; replacing them with functions would allocate on every call without adding any real protection.

internal/settings/validate_test.go
```diff
@@ -0,0 +1,294 @@
...
+	t.Run("non-json entries ignored", func(t *testing.T) {
+		t.Parallel()
+		files := validFiles()
+		files["notes.txt"] = "scratch"
+		files[".roles.json.swp"] = "vim"
+		doc, findings := parse(writeDir(t, files))
+		assert.Empty(t, findings)
+		assert.NotNil(t, doc)
+	})
+}
```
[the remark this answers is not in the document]
taitelee: `TestValidate_Warnings` was uniform and is now table-driven, matching `TestValidate_FileSyntax` and `TestValidate_ContentRules`. The remaining blocks stay as literal subtests deliberately — their cases differ structurally (a nonexistent path vs. a file-as-directory vs. deleting/adding files; multi-assertion cases with a `NotContains`; per-case `t.Setenv`), so a table would need per-case closure fields, at which point the `t.Run` blocks are the table. Same pattern as `internal/api/health_test.go` and `router_test.go`.

cmd/wavehouse/main.go
```diff
@@ -37,12 +37,16 @@ var (
...
+	// Subcommand dispatch. `health` self-probes /livez for the distroless
+	// Dockerfile HEALTHCHECK; `validate` checks a settings directory without
+	// starting the server. If this grows further, swap to a real argv router.
+	if len(os.Args) > 1 {
+		switch os.Args[1] {
+		case "health":
+			os.Exit(runHealthCheck())
+		case "validate":
+			os.Exit(runValidate(os.Args[2:]))
+		}
```
EricAndrechek: Can we make this easier to maintain somehow for if/when we add additional subcommands and also so that it can print out help and helpful error messages and stuff somehow, or what arguments something like validate might take, etc? Not sure what the best way in go to do that is, so likely needs a bit of research.

taitelee:
@coderabbitai review

taitelee:
@coderabbitai review

cmd/wavehouse/main.go
```diff
@@ -95,12 +95,16 @@ func buildInfoFallback() {
...
```
EricAndrechek, line 124: Can we make this easier to maintain somehow for if/when we add additional subcommands and also so that it can print out help and helpful error messages and the version and stuff somehow, or what arguments something like validate might take, etc? Not sure what the best way in go to do that is, so likely needs a bit of research.

internal/settings/validate.go
```diff
@@ -0,0 +1,331 @@
...
+		v.errorf("", "", "read settings directory %q: %v", dir, err)
+		return nil, false
+	}
+
+	files := map[string][]byte{}
+	for _, e := range entries {
+		name := e.Name()
+		if e.IsDir() || strings.HasPrefix(name, ".") {
+			continue
+		}
```
EricAndrechek: So we do allow nested directories or files starting with `.` in the settings dir?

internal/settings/settings.go
EricAndrechek: I actually agree with CodeRabbit's note here, this should be `const` or something like that.

internal/settings/settings.go
```diff
@@ -0,0 +1,108 @@
...
+
+// EnvDir is the environment variable naming the settings directory.
+const EnvDir = "WH_SETTINGS_DIR"
```
EricAndrechek: This is already defined and used in `internal/config/config.go`, redefining it here is just dangerous in that they could drift, have it in one spot only.

internal/settings/validate.go
```diff
@@ -0,0 +1,331 @@
...
+		if !slices.Contains(Files, name) {
+			if strings.EqualFold(filepath.Ext(name), ".json") {
+				v.errorf(name, "", "unknown settings file — expected only %s", strings.Join(Files, ", "))
+			}
```
EricAndrechek: Or actually just any non-`.json` files are allowed...? If this and the last point are intentional it feels like there's a much simpler/more readable way to do this, but my thinking was that we didn't want to allow any files other than the 4 `.json` files at all, no? What do you think? Open to suggestions/ideas here

internal/settings/decode.go
```diff
@@ -0,0 +1,88 @@
...
+		for i := 0; dec.More(); i++ {
+			if err := scanValue(dec, file, fmt.Sprintf("%s[%d]", path, i), findings); err != nil {
+				return err
+			}
+		}
+		_, err = dec.Token() // consume ']'
+		return err
+	}
+	return nil
+}
```
EricAndrechek, line 95: Not saying this is wrong, just quite complicated… is there not an easier way to do this using? For instance on a quick google it looks like `json/v2` rejects duplicates and is case sensitive and all that out of the box for us.

internal/settings/validate.go
```diff
@@ -0,0 +1,331 @@
...
+		return nil
+	}
+	if err := policy.Validate(&p); err != nil {
+		v.errorf(FilePolicies, "", "%v", err)
+	}
+	return &p
+}
+
+// pipeParamTypes are the documented ParamDef.Type values ("" = unspecified).
+var pipeParamTypes = []string{"", "string", "number", "boolean", "array"}
```
EricAndrechek: Agree with coderabbit note here too, should be `const` or something

internal/settings/validate.go
```diff
@@ -0,0 +1,331 @@
...
+		v.errorf(name, "", "file is empty — not valid JSON; an empty document is {}")
+		return false
+	case "null":
+		// The one well-formed document decodeStrict can't catch: a top-level
+		// null decodes into the zero value without error, silently reading as
+		// "no settings". (Other non-object roots already fail the decode.)
+		v.errorf(name, "", "document is null — an empty document is {}")
+		return false
+	}
+	v.findings = append(v.findings, dupKeyFindings(name, data)...)
```
EricAndrechek, line 188: Is all this really necessary, does the JSON reading part not do this for us?

internal/settings/validate.go
```diff
@@ -0,0 +1,331 @@
...
+			}
+		}
+	}
+	return f.Pipes
+}
+
+func (v *validator) parseConfig(data []byte) TenantConfig {
+	var c TenantConfig
+	if !v.parseFile(FileConfig, data, &c) {
+		return TenantConfig{}
```
EricAndrechek, line 310: Is this not an error?

internal/settings/validate.go
```diff
@@ -0,0 +1,331 @@
...
+			if td.IDField == nil && td.RequireID == nil {
+				v.warnf(FileConfig, path, "override sets nothing — remove it, or set id_field or require_id")
+			}
+		}
+	}
+	if q := c.Query; q != nil {
+		if q.DefaultMaxRows != nil && *q.DefaultMaxRows < 0 {
+			v.errorf(FileConfig, "query.default_max_rows", "must be non-negative, got %d", *q.DefaultMaxRows)
+		}
+	}
```
EricAndrechek, line 334: This sort of thing feels like it is now more complicated to do than if we were to define structs with like json comment things that go's json marshalling stuff enforced for us, or some other mechanism? Basically just every field we add looks really repetitive here, we have an if its that struct and not null and then some checks, and these will just continue to grow and are really messy to try and read through. Wonder what other ways we can do this here?

EricAndrechek (commented):
I really like the warn and error appending helpers and how you did that so that we can show all the errors or warnings we can possible figure out without a premature exit, that is great DX.

I will say, though, that a lot of this feels excessively complicated like SURELY there's a better way than rolling our own stuff to this extent? Like maybe better using Go's native JSON (v2?) library and functionality more fully with all the options and settings that can take or something, and then just the custom assertions like some value is in a safe range etc would become basic lines?

I also mentioned it earlier, but I think it's worth looking into json schemas and zod – they basically would be what would allow your VSCode to tell you the JSON is invalid for our schema before/without even having to run our binary, and I'd imagine this is a solved problem via those mechanisms, and probably has a clean way to be enforced from go too.

Finally, I thought we were trying to split our config into two types: those that need to be set prior to running the binary by the operator (and thus stay as yaml, json, or env variables with the existing config mechanism/path) and do NOT need to be hot-reloadable, and those that DO get set mid-run and should be hot-reloadable, with the latter being moved into the `config.json` file this PR is adding – but I see none of them actually moved over in this PR. Is that intentional and will be a separate PR, or what is the plan there?

cmd/wavehouse/main.go
taitelee: I'm not sure how we could make the block you're referring to more simple than that since we are already simply checking different subcommands. I am adding additional subcommands for versioning and help.

internal/settings/validate.go
taitelee: From Claude after some looking into: Any entry that is not one of the four files is now an error, whether it is a json file, a text file, or a subdirectory. The one carve out is entries starting with a dot, and that is for the standalone hand editing story rather than anything exotic: vim keeps a `.pipes.json.swp` in the directory for as long as you have the file open, and macOS drops `.DS_Store` just from browsing the folder. If those were errors, validate would fail the whole directory any time an editor is open, and once reload lands every reload during an edit would reject the directory with a finding nobody can act on. Dot entries are never settings content and a typo never produces a leading dot, so skipping them costs the strict check nothing. It also happens to keep us compatible with how Kubernetes mounts config volumes later, but that is a bonus, not the reason.

internal/settings/settings.go
taitelee: Ok. One thing to note here is that a struct tag has to be a string literal, so the tag cannot reference the constant. I have a reflection test that pins the tag to the constant, so any drift fails the config tests. You can check it out on my next push.

internal/settings/settings.go
taitelee: So Go doesn't allow const slices or maps, so I'm going to make `Files` a function that returns a fresh slice each call, which makes the set immutable that way. You can take a look on my next push, but it's a more complicated workaround. Same treatment for the param type list, but it'll be a switch inside a helper.

internal/settings/validate.go
taitelee: It is an error. `parseFile` appends the finding before returning false, so nothing is suppressed. The zero value return just lets the sweep continue through the other files so the operator sees every problem in one run, and the parsed document is thrown away whenever any error finding exists.

internal/settings/validate.go
taitelee: Yes there's one thing I made clear now. Anything that is not one of the four files is an error, json or not, files and subdirectories alike. The only exception I think should be is dot prefixed entries, because from Claude made a comment on, vim keeps a `.pipes.json.swp` in the directory while you edit and macOS drops `.DS_Store` just from browsing, and erroring on those would fail validation any time an editor is open. A typo never starts with a dot so the strict check loses nothing. The loop got simpler too, skip dots, check membership, parse. You can take a look and confirm yourself and let me know what you think. I guess I could just keep it strict to the four files we have designated right now, but errors might come up in the near future.

internal/settings/validate.go
taitelee: Made a comment on this above.

internal/settings/decode.go
taitelee: Ok so looked more into this and Claude found some things for me. It seems that `encoding/json/v2` in Go 1.26 is still gated behind `GOEXPERIMENT=jsonv2` and does not even compile in a normal build, so adopting it today means turning on an experiment flag for the entire binary. There is also a behavioral gap that matters here. `v2` rejects duplicate keys but fails on the first one with a single error, while this scanner reports every duplicate with its full path, like`tables.clicks.select.a`, which keeps the show everything in one pass behavior you mentioned liking. So even once `v2` is stable the scanner keeps earning its keep for the path reporting, though `v2` could replace `decodeStrict` for the unknown field handling at that point. I should probably track this or mark a todo somehwere?

internal/settings/validate.go
taitelee: It does not do this for us, each check closes a specific hole. The stock decoder accepts a top level null without error and leaves the struct at its zero value, which would silently read as no settings. An empty or truncated file comes back as the bare error EOF, which tells an operator nothing. And duplicate keys are silently resolved as last one wins with no error at all.

internal/settings/validate.go
taitelee: Yes this will not scale, and I think the fix is the JSON Schema idea you have rather than struct tags. Something I would propose is publish a schema per file, which gets the editor experience you talk about for free since VSCode reads $schema and flags problems before our binary ever runs, then validate against those same schemas in Go with a some library (any) which reports paths we can map straight onto our findings. The hand written code then shrinks to what schemas cannot express, meaning the cross file role references and the warning tier. I think this should be a separate PR though what do you think?

internal/settings/validate.go
EricAndrechek: Ok yeah that makes sense to me, we allow dot-prefixed entries, that sounds good.

internal/settings/settings.go
EricAndrechek: Haven't looked at your new change yet, but why not just make it private for only this package to access then? And maybe fix a size? So like:
```go
var files = [4]string{FileRoles, FilePolicies, FilePipes, FileConfig}
```

internal/settings/decode.go
EricAndrechek: Ah ok good point, didn't think about it erroring out on the first occurrence vs not. As for JSON, Go 1.27 just dropped which I think brings `json/v2` into the normal runtime which is why I was suggesting that since I imagine we'll bump go versions sometime soonish anyway, but yeah doesn't look like there's a way to have it detect all occurrences vs just failing at the first one, so even if/when we start using `json/v2` we still will need this code. But yeah, `v2` could at least simplify `decodeStrict`, I agree.

internal/settings/validate.go
EricAndrechek: Yeah with this one being private to the package already, I think it's probably fine. We could potentially do like:
```go
var pipeParamTypes = [5]string{"", "string", "number", "boolean", "array"}
```
so it's at least fixed, but at that point we might as well make a struct with an iota or something…

internal/settings/validate.go
EricAndrechek: Is this something that would collapse/be fixed by adopting `json/v2`? We don't necessarily need to rush/force adoption of go 1.27 now in this PR and switch to using that, as it may break things elsewhere, but I have to imagine it would certainly make our live's easier here? I don't think an issue just to track this is warranted, but maybe a `` // TODO: remove with `json/v2` `` would be good?

internal/settings/validate.go
EricAndrechek: Yeah I really like this idea, good plan, let's do it

internal/settings/settings.go
```diff
@@ -16,22 +16,24 @@ import (
...
-
-// EnvDir is the environment variable naming the settings directory.
-const EnvDir = "WH_SETTINGS_DIR"
+// Files returns every expected file. All four must exist (an empty document
+// is written as {}), so an accidental deletion reads as an error, never as
+// "defaults". A function rather than a package variable because Go has no
+// const slices — this keeps the set genuinely immutable.
+func Files() []string {
+	return []string{FileRoles, FilePolicies, FilePipes, FileConfig}
+}
```
EricAndrechek, line 36: This also works for me, either way is fine by me up to you

internal/settings/validate.go
```diff
@@ -188,8 +233,16 @@ func (v *validator) parsePolicies(data []byte) *policy.Policy {
...
+// validParamType reports whether t is a documented ParamDef.Type ("" =
+// unspecified). A switch rather than a package-level slice so the set is
+// genuinely constant.
+func validParamType(t string) bool {
+	switch t {
+	case "", "string", "number", "boolean", "array":
+		return true
+	}
+	return false
+}
```
EricAndrechek, line 245: Oh yeah, seeing as we don't iterate over the param types or anything like we do for `Files`/`Files()`, this purely exposed as a helper function to clean up the old `if !slices.Contains(pipeParamTypes, param.Type) {` code with `if !validParamType(param.Type) {` is MUCHHHH cleaner than trying to make a getter or pseudo-const slice, good move

internal/settings/settings.go
EricAndrechek: Yeah current code version works fine too, resolving this one – will move any comments on approach to new thread

internal/settings/settings.go
EricAndrechek: Ah ok I see what you mean there, yeah it's going to more or less need to be in two places. I think a unit test to guard against them drifting from equal is perfect, like literally just asserting they are the same is perfectly fine and simple.

internal/settings/validate.go
EricAndrechek: Edit: saw the new code for this and like its approach much better, resolving

cmd/wavehouse/main.go
EricAndrechek: I looked into it a bit myself. I think that at a minimum we start using Go's stdlib flag (`flag.FlagSet`) package, so `cmd/main.go` would start looking something more like:
```go
func main() {
	if len(os.Args) > 1 {
		switch os.Args[1] {
		case "health":
			os.Exit(runHealthCheck(os.Args[2:]))
		case "validate":
			os.Exit(runValidate(os.Args[2:]))
		case "version", "--version", "-v":
			fmt.Printf("wavehouse %s (commit %s, built %s)\n", Version, GitCommit, BuildTime)
			os.Exit(0)
		case "help", "--help", "-h":
			printUsage(os.Stdout)
			os.Exit(0)
		default:
			fmt.Fprintf(os.Stderr, "wavehouse: unknown command %q\n\n", os.Args[1])
			printUsage(os.Stderr)
			os.Exit(2)
		}
	}
	os.Exit(run())
}
```
But that too will grow to be too complex to maintain if we add any/many more options or subcommands, etc. and would need changed in the near-term too. So, I think either do the above as a minimum change and no new dependencies, or look into switching us to using cobra (as the new code comment mentions) or kong, decision is yours to make.

EricAndrechek (changes requested):
Thanks for the explanations on things, I resolved a bunch of old threads and followed up on others. I think the main remaining points are:
- [x] when it was just run with no commands (and the config loader instead), and then one jerry-rigged `health` command for docker healthchecks it wasn't a huge deal, but as we introduce more actual subcommands we recommend people run, like `validate` with its own options and help output ideally, we need to actually use a proper cli tool, whether it be stdlib's `flag` package, or the industry standard `cobra` one that k8s and docker and stuff use, or something like `kong` – whatever it is, we should add something NOW before the problem continues cascading and growing.
- [ ] I don't know if you saw it in my last review comment as it was seemingly unaddressed or I just missed it, but re-pasting below:
> Finally, I thought we were trying to split our config into two types: those that need to be set prior to running the binary by the operator (and thus stay as yaml, json, or env variables with the existing config mechanism/path) and do NOT need to be hot-reloadable, and those that DO get set mid-run and should be hot-reloadable, with the latter being moved into the config.json file this PR is adding – but I see none of them actually moved over in this PR. Is that intentional and will be a separate PR, or what is the plan there?

Would love those two things addressed, and a new issue opened for PR work to begin on either a json schema or zod or something, and then this is ready to merge. Thanks!

internal/settings/settings.go
taitelee: Yep just noting that it is addressed in `TestEnvSettingsDir_MatchesStructTag` in `internal/config/config_test.go`.

internal/settings/decode.go
taitelee: Adding a TODO on `decodeStrict` (gated on the go directive reaching 1.27), with a note that the scanner survives v2 for the every-occurrence-with-path reporting.

internal/settings/validate.go
taitelee: Adding TODO on `decodeStrict` (mentioned this in previous comment) since that's the part v2 can do for us (unknown fields + trailing content via `RejectUnknownMembers`). The comment also notes what v2 can't do so nobody deletes something important later. The dup-key scanner we are keeping for finding and reporting all the duplicate key occurrences since v2 fails on the first dup. The null/empty-file checks also stay because v2 still decodes a top-level null to a zero value without error.

cmd/wavehouse/main.go
taitelee: Ok going with the minimum for now. The switch stays as the router (it was already basically what you had above), and each subcommand now owns a `flag.FlagSet`, so `validate -h` and `health -h` will print out real per-command help, unknown flags and stray `args` will `exit 2`, and `health` takes `os.Args[2:]` like the others. I think for now we don't have to use cobra since we have four commands and zero actual flags. The per-subcommand `FlagSet` shape is exactly what a cobra port would be useful for, so maybe we can look into that later?

internal/settings/validate.go
taitelee: Opened #506 to track this

taitelee:
Ok so my reasoning for not using cobra yet is that we have only four commands and zero real flags so far, and the FlagSet-per-subcommand is exactly what a cobra port would be useful for and use, so we can just easily get that in when that does happen. I think this is marked as a commet in main.go. But again I also don't have a strong opinion on whether or not we should just integrate cobra now. 

On the split you mention for the config stuff, `TenantConfig` in `settings.go` names exactly which keys move into hot-reloadable `config.json` (dedupe `id_field`/`require_id` + per-table overrides, `query.default_max_rows`, `schema.refresh_interval`, `cors.allowed_origins`), and the doc comments highlight what stays boot config (lifecycle things like `dedupe.enabled`, since only a restart can change Pebble's lifecycle, secrets and wiring, platform-infra toggle stuff like the SSE keepalives). The next PR will include boot loading + reload and removes the migrated keys from `config.yaml`.

Opened #506 for the JSON Schema work.

internal/settings/validate_unix_test.go
```diff
@@ -0,0 +1,50 @@
...
+	dataDir := filepath.Join(dir, "..data")
+	require.NoError(t, os.Mkdir(dataDir, 0o700))
+	for name, content := range validFiles() {
+		require.NoError(t, os.WriteFile(filepath.Join(dataDir, name), []byte(content), 0o600))
+		require.NoError(t, os.Symlink(filepath.Join(dataDir, name), filepath.Join(dir, name)))
+	}
+
+	doc, findings := parse(dir)
+	require.NotNil(t, doc, "findings: %s", findingStrings(findings))
+	assert.Empty(t, findings)
```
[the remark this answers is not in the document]
taitelee: From Claude: Partially applied: TestRunHealthCheck_Flags is now table-driven and gained the unknown-flag case — good catch on the uncovered fs.Parse error path. The other two sites stay as literal subtests for the reason already given on the earlier table-driven thread in this PR: their cases differ structurally (per-case t.Setenv and directory setup in TestRunValidate; opposite setups and assertions in the FIFO vs. symlink tests), so a table would need per-case closure fields for both setup and asserts — at which point the t.Run blocks are the table. Same pattern as internal/api/health_test.go and router_test.go.

cmd/wavehouse/main.go
EricAndrechek: Works for me

EricAndrechek (approved):
Looks good, and gotcha I see so we will move the things out of the current config and into this new file config in the next PR then, not all at once that makes sense.

My only remaining note would be that I think we can probably move many more settings away and into this new `dir/config.json` approach, but I think it will be easier to audit which can get moved or not in the PR where you actually do that so we can say what remains.

taitelee:
@coderabbitai review

taitelee:
@coderabbitai review

Issue #461, opened 2026-08-12
security(policy): policies loaded from KV skip validation, so validation fixes never reach existing deployments


EricAndrechek:
**Area:** policy — security (a shipped validation fix does not reach existing deployments) · found via PR #457 review

**Expected:** every path that loads a policy into the in-memory cache validates it first, so `Evaluate` can never be handed a policy `Validate` would reject — and so a validation rule added later actually takes effect on deployments that are already running.

**Actual:** only `Store.Put` (`internal/policy/store.go:95`) calls `Validate`. The two paths that populate the same cache from KV do not:

- `Store.load` — `internal/policy/store.go:184-188`
- `Store.Watch` — `internal/policy/store.go:152-159`

Both unmarshal the KV value and cache it directly.

## Why this is no longer just defense-in-depth

This was originally filed as low-impact — reachable only by writing the KV bucket out-of-band. #457 changed that. It added a config-load rejection for malformed claim templates (a hyphenated or namespaced-OIDC path like `{{ jwt.https://app.example.com/tenant_id }}`, which the resolver would otherwise bind as literal `{{…}}` text — a read leak on `_neq`/`_lt` and silent row corruption on `check`). That guard runs in `validateRolePerms`, so it only fires where `Validate` is called.

Trace through `NewStore` (`internal/policy/store.go:44-84`):

| Deployment state | Path | Malformed template outcome |
|---|---|---|
| Fresh deploy, `policy.file_path` set, KV empty | `loadPolicyFile` → `Put` → `Validate` | **fatal — server refuses to start** ✅ |
| Runtime update | `POST /v1/admin/policy` → `Validate` (`internal/api/policy.go:79`) | 400, rejected ✅ |
| **Existing deploy, policy already in KV** | `s.load(ctx)` → cache → `return s, nil`, **no `Validate`** | **loads unvalidated — the fail-open persists silently** ❌ |

The KV-populated branch is the first thing `NewStore` tries, and it is documented as intentional ("KV is authoritative when populated" — the bootstrap file is a seed, #229). That is the right call for *configuration precedence*, but it means a security rule added to `validateRolePerms` never applies to any deployment that already has a policy — which is precisely the population most likely to be carrying the bad policy already. They stay exposed until someone happens to re-`PUT`.

This generalizes: **every future rule added to `validateRolePerms` inherits the same blind spot**, including the ones proposed in #460.

## Scope

- Validate on the `load` and `Watch` paths, not just `Put`.
[...]

Issue #460, opened 2026-08-12
security(policy): mistyped or unhonored policy rules load silently and disable row security


EricAndrechek:
**Area:** policy — security (fail-open) · found via PR #457 review

**Expected:** a policy rule that the engine will not honor — because it was mistyped, or because it has no meaning for that operation — is rejected at config load, loudly. Row security is never silently disabled.

**Actual:** two independent gaps let a malformed rule load cleanly and produce **no predicate at all**, which on the row-`filter` path means the role reads the whole table.

## 1. A typo'd operator key silently disables row security

Every decoder is non-strict — `internal/policy/store.go:153,185,210,214` and `internal/api/policy.go:46,71` — so an unknown key is dropped rather than rejected. A `Filter` with all five operator pointers nil then contributes zero clauses (`resolveFilters` has one `if` per operator and no `else`), `Evaluate` leaves `WhereClause` empty (`internal/policy/policy.go:192-196`), and `InjectPermissionFilters` no-ops on an empty clause.

`validateRolePerms` (`internal/policy/policy.go:517-523`) checks only `chsql.BindUnsafe(col)` for filter columns — it never requires that an operator be set.

Verified end-to-end:

```yaml
select:
  user:
    filter: { tenant_id: { eq: "{{ jwt.tenant_id }}" } }   # "eq", not "_eq"
```

```
→ Filter{Eq:nil, Neq:nil, Gt:nil, Lt:nil, In:nil}
→ Allowed=true, WhereClause=""
→ SELECT * FROM `clicks` LIMIT 10000        # entire table, every tenant
→ Validate() == nil, and POST /v1/admin/policy/validate returns {"valid":true}
```

A one-character typo disables RLS, and the validate endpoint actively confirms the policy is fine.

Note the asymmetry: Go's `encoding/json` matches field names case-insensitively, so `_EQ` and `_Eq` **do** work — but `eq` does not, and YAML is exact-match only. The format is lenient in a direction nobody needs and silent in the one that bites.

## 2. `filter:` under an `insert:` role entry is resolved, then ignored

`internal/policy/policy.go:186-197` resolves filters for every operation, but `internal/api/ingest.go` reads only `CheckClauses` and the column rules. Nothing rejects the policy — in pointed contrast to the loud config-load rejection of `_neq`/`_gt`/`_lt` on `check` at `internal/policy/policy.go:531`. Same accept-but-ignore family as #224.

## Scope

[...]

Issue #363, opened 2026-07-01
bug/feat(pipes): WH_PIPES_DIR bootstrap can't declare allowed_roles/params (lands admin-only); inline empty default `{{name:}}` treated as required


EricAndrechek:
## Summary
Two related rough edges in the `WH_PIPES_DIR` file-bootstrap path for named pipes.

**1. Bootstrapped pipes are stuck admin-only.** A `.sql` file loaded from `WH_PIPES_DIR` becomes a pipe with **empty `allowed_roles` → admin-only** (`internal/pipes/pipes.go`: _"empty = admin role only (fails closed)"_), and the file can't declare `allowed_roles` or parameter definitions. So a file-bootstrapped *public* pipe can't actually be reached by a non-admin role — you must re-register it via `PUT /v1/admin/pipes/{name}` with roles+params. In practice `WH_PIPES_DIR` is only usable for admin-only, param-less pipes.

**2. Inline empty default `{{name:}}` is treated as REQUIRED (bug).** A pipe body with `... {{division:}} ...` rejects a call that omits `division` (HTTP 400 "missing required param") instead of defaulting to empty string. Given `{{name:default}}` semantics, `{{name:}}` reads as "default = empty string", not "required". (e.g. `scoreboard_public?day=1` 400s because `division` has an empty inline default.)

## Impact
`WH_PIPES_DIR` looks like a convenient GitOps way to ship a pipe library, but every pipe then needs a follow-up admin-API call to set roles + dodge the empty-default trap — so the bootstrap is effectively unusable for the common case (public pipes with optional params).

## Suggested
- Let the file-bootstrap declare `allowed_roles` + param defs — e.g. a sidecar `<name>.json`/`.yaml`, SQL frontmatter, or a `pipes.manifest.json` the loader reads.
- Treat `{{name:}}` as an **empty-string default**, not a required param.

(Related to the tracked "bring named pipes under the main access-control policy" work. Found building the SCC 2026 tracker pilot; we ship a `pipes.manifest.json` + register via the admin API to work around both.)

Issue #214, opened 2026-06-03
Per-table policy storage with per-table authorization scopes (unblocks multi-tenant + e2e parallelization)


EricAndrechek:
## Problem

The access-control policy is stored and served as a **single global document**: `internal/policy` keeps one NATS KV entry, and `GET`/`PUT /v1/admin/policy` read and **replace the whole thing**. Two structural limits fall out of this:

1. **No atomic / concurrent per-table edits.** `PUT /v1/admin/policy` is a full-document replace (`policy.Store.Put`) with **no optimistic concurrency** — no revision / `If-Match`. Two callers doing read-modify-write race: last writer wins and the other's table edit is silently lost. (The KV layer *has* revisions — `internal/policy/store.go` logs `entry.Revision()` on watch — but the HTTP API never exposes a compare-and-swap.)

2. **No per-table authorization scope for managing policy itself.** Policy editing is all-or-nothing today: any `admin_role` caller can read **and** rewrite the **entire** policy — every table, and (in a shared deployment) every tenant. There is no notion of "role X may manage the policy for table Y." Policy *management* needs its own per-table (or per-namespace/tenant) read/write scopes — a meta-policy over who can edit which tables' rules — distinct from the data-plane `admin_role`.

## Why it matters (multi-tenant)

A single WaveHouse instance fronting ClickHouse for multiple tenants (each owning a set of tables) currently shares **one** policy document with **no isolation of edit rights**. Tenant A's admin can read and modify tenant B's table policies, and concurrent edits from different tenants clobber each other. Safe multi-tenancy needs per-table policy storage plus scoped management authority.

## Proposed direction (sketch — design TBD)

- **Store policy per table** (one KV entry per table, e.g. `WAVEHOUSE_POLICY.<table>`) rather than one monolithic document, so updates are naturally atomic per table and can't clobber unrelated tables.
- **Per-table policy endpoints**: `GET`/`PUT`/`DELETE /v1/admin/policy/tables/{table}` with **optimistic concurrency** (return a revision; require `If-Match` on write). Keep `GET /v1/admin/policy` as a read-only roll-up for compatibility.
- **Per-table policy-management scopes**: a capability controlling who may read vs. write each table's policy (per-table / per-namespace / per-tenant), separate from the data-plane `admin_role`. Define how it composes with the existing single `admin_role` (e.g. `admin_role` = superuser; scoped roles manage only their own tables). This is the meta-policy layer that makes multi-tenant policy administration safe.
[...]

PR #508, opened 2026-08-21
feat(settings): hot-reload the settings directory at boot and runtime

### LINKED ISSUE #229: Policy/pipes bootstrap file is a seed, not a synced source of truth (document + optional reconcile-on-boot)
**Area:** policy · ops — footgun (silent) + docs · found via WaveHouse-Stats dogfooding

**Expected:** editing the bootstrap policy file and restarting updates the live policy.

**Actual:** it doesn't. Once NATS KV holds a policy, the file is ignored forever; changes must go via `PUT /v1/admin/policy` (`internal/policy/store.go:33-43`, `:55-84`). The `store.go` docstring is explicit, but the *behavior* (edit file, restart, nothing changes) is a silent operational surprise and isn't in the consumer-facing docs.

**Impact (Stats):** confirms our GitOps model — the git file is source-of-truth, but CI MUST apply it via the admin API on change; the file only seeds a truly empty KV.

**Scope:** document the seed-vs-source-of-truth distinction prominently; optionally an opt-in "reconcile from file on boot" mode for file-as-SoT deployments.

---

taitelee:
## Summary

The server now consumes the settings directory that #500 only validated. `settings.dir` / `WH_SETTINGS_DIR` is required: boot validates and adopts it (missing or invalid refuses to start, same contract as `policy.file_path`), and a running instance re-validates and re-adopts on a directory watch (fsnotify with debounce; survives a delete-and-recreate), `SIGHUP`, or admin-gated `POST /v1/ops/settings/reload`, all through one serialized reload path. A rejected reload, a deleted file, or a vanished directory keeps the previous good snapshot.

- Tenant tunables move out of boot config into `config.json`: the ClickHouse wiring (`clickhouse.addr` / `http_port` / `http_scheme` / `database` / `username` / `query_timeout` — `chconn.Manager` swaps the connection behind one `driver.Conn` on reload), the auth verifier wiring (`auth.jwks_url` / `auth.role_claim` — `auth.Authenticator` swaps a whole verifier atomically; `auth.Middleware` is gone), `dedupe.enabled` (the new `dedupe.Managed` opens or closes the Pebble store from an `AfterAdopt` hook) with `dedupe.id_field` / `dedupe.require_id` and per-table overrides (#222), `dlq.enabled` with per-table overrides, `query.default_max_rows` / `query.timestamp_bucket_seconds`, `schema.refresh_interval`, `stream.keepalive_interval` / `keepalive_buckets` (`Heartbeater.Reconfigure` rebuilds the wheel in place) / `gap_window_minutes`, and `cors.allowed_origins`. The YAML/env keys for all of these are removed, and boot config is now strict: an undeclared key refuses to boot, naming it. What stays in boot config is only what can't change under a running process — resource sizing, listeners, observability exporters — and the secrets (`clickhouse.password`, `auth.jwt_secret`, `auth.operator_key`), never in a tracked JSON file.
- Every `config.json` key is required and the binary carries no compiled defaults, so the adopted snapshot is what the files say. Defaults live in one checked-in seed (`internal/settings/seed/`, `go:embed`ded): the new `wavehouse bootstrap [dir]` writes it (refusing a non-empty directory; resolves `dir` exactly as `validate` does — argument, else `WH_SETTINGS_DIR`), the compose stack bind-mounts a checked-in copy at `deployments/compose/settings/` so the quickstart stays `up -d`, `make dev` seeds a gitignored `./settings`, and the e2e fixture ships its own. The container images ship no settings directory: `WH_SETTINGS_DIR=/app/settings` is preset and the operator mounts one there — a missing mount refuses to boot rather than running on defaults nobody chose. The seed ships no policy (fail-closed, warned).
- Consumers take functions instead of values (`IngestHandler.DedupeSettings`, the structured-query `defaultMaxRows` / `bucketSecs` getters, the ingest worker's `dlqEnabled`, the sweeper's `gapWindow`, `corsMiddleware`'s origins getter, `SchemaRegistry`'s database and refresh-interval sources) so a reload applies without a restart and `internal/api` stays testable without a directory.

- `roles.json`, `policies.json`, and `pipes.json` are the runtime authority for access control and named pipes, read per request off the same adopted snapshot (`policy.Source` / `pipes.Source`; `settings.Store.Policy()` / `Pipe()` / `Pipes()`), so a reload applies to the next request. Every adoption runs the full validation (strict JSON, cross-file role references), so there is no stored copy that skips it. Files are the only write path — edited on the host standalone, written by the control plane in cloud — because without NATS KV there is no bundled node sync and a node-local write would let distributed nodes drift. Removed: `PUT /v1/ops/policy`, `PUT`/`DELETE /v1/ops/pipes/{name}`, the `WAVEHOUSE_POLICY` / `WAVEHOUSE_PIPES` KV buckets and KV watch, `policy.file_path` / `WH_POLICY_FILE_PATH`, `pipes.dir` / `WH_PIPES_DIR` and the `.sql` bootstrap, `deployments/compose/dev-policy.yaml` (the trial policy now ships in `deployments/compose/settings/`, which `make dev` also seeds), `/app/pipes` in the images, and the SDK `wh.policy.set` / `wh.pipes.set` / `wh.pipes.delete` (replaced by `wh.settings.reload()`). The reads stay: `GET /v1/ops/policy`, `POST /v1/ops/policy/validate`, `GET /v1/ops/pipes[/{name}]`. The operator key remains the break-glass for a locked-out deployment: it can inspect the adopted policy and trigger a reload after the files are fixed.
- `mq.max_bytes_gb` is a `config.json` key too: a reload updates the ingest and DLQ stream limits in place (`DiscardNew`, so shrinking backpressures until the worker drains — nothing buffered is dropped).

## Related Issues

Closes #229, #33, #461, #514, #460, #363. Advances #48 (reload wiring; reverses the #48-era runtime-settings direction), #214, and #222 (per-table `id_field`; the keyspace fix stays open there). Follow-up surfaced in review: #510 (distroless digest pin).


deployments/Dockerfile
```diff
@@ -30,13 +30,20 @@ RUN --mount=type=cache,target=/go/pkg/mod \
...
+#
+# /app/settings is the baked seed settings directory (every key at its
+# default, the same files `wavehouse init-settings` writes) so a bare
+# `docker run` boots; production bind-mounts its own directory over it.
+# Deliberately NOT a VOLUME (unlike /app/data): it's config, not state, and an
+# anonymous volume would hide a forgotten mount.
+RUN mkdir -p /app/data /app/pipes && cp -r /src/internal/settings/seed /app/settings && chown -R 65532:65532 /app
 
 # Stage 2: Final minimal image
 FROM gcr.io/distroless/static-debian12
```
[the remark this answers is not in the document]
taitelee: Pre-existing: both FROM lines are unchanged from main; this PR only adds the seed COPY and ENV WH_SETTINGS_DIR to these files. Agreed it's worth doing, and the two Dockerfiles have to move together (the goreleaser variant is documented as lockstep with the source build), plus confirming Dependabot bumps the digest form. Tracked in #510 so it gets its own change rather than riding in a settings PR.

internal/api/ingest_test.go
```diff
@@ -163,7 +163,7 @@ func TestIngest_Dedup_FirstTime(t *testing.T) {
 	dedup := testutil.NewMockDeduplicator()
 	h := NewIngestHandler(testRegistry(t), pub, testutil.NopLogger())
 	h.Dedup = dedup
-	h.IDField = "event_id"
+	h.DedupeSettings = func(string) (string, bool) { return "event_id", false }
```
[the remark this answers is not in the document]
taitelee: The table argument is already exercised where the resolution lives: TestStore_DedupeFor_Cascade in internal/settings covers table override, inherited field, and unlisted table. These handler tests pin ingest's use of the resolved pair (missing-field warn vs. reject paths), so a constant callback is the right stub for them. The two-table "same id, both publish" assertion you describe can't pass today regardless of what the callback does — the Pebble key has no table namespace (see the thread on ingest.go:447) — so that test belongs with the keyspace fix in #222, where it would actually be a regression test rather than a known-failing one.

internal/api/ingest.go
```diff
@@ -422,19 +425,23 @@ func (h *IngestHandler) processRecord(
...
 					Status:  http.StatusBadRequest,
-					Message: fmt.Sprintf("missing dedupe id field %q", h.IDField),
+					Message: fmt.Sprintf("missing dedupe id field %q", idField),
 				}, nil
 			}
-			h.logger.WarnContext(ctx, "dedupe id_field missing; publishing without idempotency", "id_field", h.IDField, "table", table)
+			h.logger.WarnContext(ctx, "dedupe id_field missing; publishing without idempotency", "id_field", idField, "table", table)
 		} else {
 			eventID := fmt.Sprint(idVal)
 			dup, err := h.Dedup.CheckAndMark(ctx, eventID)
```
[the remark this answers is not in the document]
taitelee: Real, but pre-existing: main has the same CheckAndMark(ctx, eventID) with a raw-id key, and it's already written up as problem 2 in #222 (the per-table id_field in this PR is problem 1 of the same issue). Not taking the suggested diff here on purpose: changing the Pebble key format means every previously seen id is accepted once more after upgrade — fine pre-1.0, but that needs its own changelog line and review, not a silent inclusion in a settings PR. I've left a status comment on #222 noting that this PR advances it and the keyspace fix stays open there.

internal/settings/validate.go
```diff
@@ -309,8 +315,16 @@ func (v *validator) parseConfig(data []byte) TenantConfig {
...
+	if d := c.Dedupe; d == nil {
+		v.required("dedupe")
+	} else {
+		if d.IDField == nil {
+			v.required("dedupe.id_field")
+		}
+		if d.RequireID == nil {
+			v.required("dedupe.require_id")
+		}
+		v.checkIDField("dedupe.id_field", d.IDField)
```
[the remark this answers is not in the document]
taitelee: Intentional. tables is the one non-scalar key: a map of optional per-table overrides, where nil and {} mean the same thing — no overrides. The required-keys rule exists so no tunable's effective value can depend on something the files don't state; an absent overrides map doesn't create that gap, because the global id_field / require_id are required and always apply. The docs table marks it optional (seed writes {} so the shape is discoverable). Rejecting nil would add a failure mode with no semantic payoff. If #506's JSON Schemas want tables non-nullable for editor tooling, that's the place to revisit.

internal/api/router_test.go
```diff
@@ -120,9 +121,43 @@ func TestRequireAdmin_OperatorBypassesNilPolicy(t *testing.T) {
...
+	assert.Empty(t, get("https://new.example.com").Header().Get("Access-Control-Allow-Origin"), "not yet allowed")
+
+	// "Reload": swap the list the getter returns.
+	mu.Lock()
+	origins = []string{"https://new.example.com"}
+	mu.Unlock()
+
+	assert.Equal(t, "https://new.example.com", get("https://new.example.com").Header().Get("Access-Control-Allow-Origin"), "new allowlist applies on the next request")
+	assert.Empty(t, get("https://old.example.com").Header().Get("Access-Control-Allow-Origin"), "old origin no longer allowed")
+}
```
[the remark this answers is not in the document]
taitelee: Leaving this one as a single sequential test on purpose. It isn't a set of independent cases — it's one scenario with an ordered narrative: request under the old allowlist, swap the getter's list (the "reload"), request again. Splitting that into t.Run subtests would make them order-dependent and non-parallel, which is the opposite of what the table-driven pattern is for, and each subtest would be asserting a step of the same story rather than a case. The independent CORS cases in this file (TestCORSMiddleware_AllowedOrigins and friends) already use the table form.

taitelee:
@coderabbitai review

taitelee:
@coderabbitai review

deployments/compose/standalone.yaml
```diff
@@ -39,19 +39,23 @@ services:
       # quickstart works tokenless. Tune it before production — see that file.
       WH_POLICY_FILE_PATH: /app/policy.yaml
       WH_DEDUPE_ENABLED: "false"
-      # With dedupe on, reject rows missing the id field instead of publishing
-      # them un-deduped (default false → logged + counted, not rejected).
-      WH_DEDUPE_REQUIRE_ID: "false"
-      WH_SCHEMA_REFRESH_INTERVAL: "60"
       WH_DLQ_ENABLED: "true"
```
EricAndrechek: Why are these other config things not moved over too?

deployments/Dockerfile.goreleaser
```diff
@@ -17,6 +17,10 @@
 # once natively and COPY'ing into each target image is correct AND
 # removes the binfmt/QEMU dependency entirely.
 FROM --platform=$BUILDPLATFORM alpine:3 AS layout
+# The seed settings directory arrives via `extra_files` in .goreleaser.yaml
+# (the goreleaser context holds only the dist tree). Baked at /app/settings so
+# a bare `docker run` boots; not a VOLUME — it's config, not state.
+COPY internal/settings/seed /app/settings
 RUN mkdir -p /app/data /app/pipes && chown -R 65532:65532 /app
```
EricAndrechek: Why'd you change it differently in this file compared to the plain `Dockerfile`? Why not change it so both are changed the same way to better keep inline/in sync with each other?

docs/src/content/docs/configuration.mdx
```diff
@@ -129,9 +128,13 @@ WaveHouse's per-role caps are sent as per-query `SETTINGS` on its connection, so
 
 | YAML Key | Env Var | Default | Description |
 | --- | --- | ------- | ----------- |
-| `dedupe.enabled` | `WH_DEDUPE_ENABLED` | `false` | Enable event deduplication. When enabled, the ingest handler checks for duplicates using the configured ID field. |
-| `dedupe.id_field` | `WH_DEDUPE_ID_FIELD` | `event_id` | JSON field name in the ingest body used as the dedup key. |
-| `dedupe.require_id` | `WH_DEDUPE_REQUIRE_ID` | `false` | With dedupe enabled, controls what happens to a row missing `id_field` (which can't be deduped, so idempotency wouldn't apply to it). Such a row is always logged at `WARN` and counted by `wavehouse_ingest_dedupe_missing_id_total`, in both modes. Default (`false`): it is then published un-deduped. Set `true` to reject it instead (`400` for a single insert; a per-record failure in a batch) — a server-side tripwire for producers that must guarantee the id. |
+| `dedupe.enabled` | `WH_DEDUPE_ENABLED` | `false` | Enable event deduplication. Boot config because it owns the embedded Pebble store's lifecycle, which only a restart can change. |
```
EricAndrechek: I'd like pebble and dedupe settings to be able to become hot-reloadable too

internal/settings/seed.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+// seedFS holds the starter settings directory: every file present, every
+// key set to its default. It is the ONE place defaults live — the binary
+// has no compiled fallbacks — and it ships three ways: `wavehouse
+// init-settings` writes it out, the container images bake it at
+// /app/settings, and the compose quickstart serves from that copy.
+//
+//go:embed seed/*.json
+var seedFS embed.FS
```
EricAndrechek, line 19: I don't understand this, why do it so many different ways? Like if we are aiming to embed it in the go binary, then why have the docker image builder do it a different way?

docs/src/content/docs/configuration.mdx
```diff
@@ -182,11 +185,55 @@ See [API — Authentication](/api#authentication).
...
+
+The server validates and adopts the directory at boot — a missing or invalid directory refuses to start, the same fail-loud contract as `policy.file_path` (run `wavehouse validate` to reproduce the findings, `wavehouse init-settings` to create a starter directory).
+
+A running instance re-validates and re-adopts on any of three triggers, all funneling through one serialized reload path:
+
+- **File change** — the directory is watched; a burst of writes (an editor save, a Kubernetes ConfigMap republish) coalesces into one reload, and a directory that is deleted and recreated (or renamed over) is picked up again.
+- **`SIGHUP`** — `kill -HUP <pid>`.
+- **`POST /v1/ops/settings/reload`** (admin-only) — returns `{"adopted": bool, "findings": [...]}`; `200` when adopted, `422` when rejected.
+
+A reload that fails validation is logged (and reported by the endpoint) and the previous good settings stay in effect — an operator mid-edit, a deleted file, even a vanished directory degrades to a log line, never to a broken server. Warnings don't block adoption, matching `wavehouse validate`.
```
EricAndrechek: I think we talked about this and my understanding was that it would fail to boot/start if the validation failed – otherwise with this approach you describe now, how does a server restart continue operating with the old previous good settings even if the files were mid-edit and bad or whatever? Like this would require a hidden internal copy of the files, no?

docs/src/content/docs/configuration.mdx
EricAndrechek, line 1: Ok so this file has me thinking, and I'd love your thoughts on this too, what do you think about splitting this file into two now? One for the operator configs that aren't hot reloadable, and one for the things that are? I think it would help make it more organized and make a lot more sense, no?

docs/src/content/docs/configuration.mdx
```diff
@@ -287,9 +325,10 @@ pipes:
   dir: ""                # empty = skip bootstrap; set + read-only mount to seed pipes
 
 settings:
-  dir: ""                # settings directory (roles/policies/pipes/config .json);
-                         # `wavehouse validate` reads its [dir] argument or
-                         # WH_SETTINGS_DIR (the env var), not this YAML key
+  dir: ""                # REQUIRED: settings directory (roles/policies/pipes/
```
EricAndrechek: This block is introduced as "Every key, with its default. Save the YAML as `config.yaml` next to the binary" — but with `dir: ""` it won't boot. `internal/config/config.go` now rejects an empty `settings.dir`, so following these instructions literally gets you:

```
{"level":"ERROR","msg":"load config","error":"validate config: settings.dir (WH_SETTINGS_DIR) is required: point it at a settings directory, or create one with `wavehouse init-settings <dir>`"}
```

The comment right next to it even says REQUIRED, which makes the empty value read like an oversight rather than a default.

I'd treat this as one item with three sites rather than three separate fixes — it's the same stale copy each time. `config.yaml` itself got a real value on this branch (`dir: internal/settings/seed`), but nothing downstream of it did:

1. **This example block**, plus `WH_SETTINGS_DIR=` in the env tab at :392.
2. **`.config.local.yaml`** — the Makefile only copies `config.yaml` over when the file is absent, so every existing checkout keeps `dir: ""` and `make dev` now exits 1 at boot with nothing pointing at the cause. `.air.toml` doesn't set the env either, so there's no fallback. At minimum this needs a "delete `.config.local.yaml` to re-seed" line in `development.md`; better would be having the recipe warn when `config.yaml` is newer than the local copy instead of silently skipping.
3. **README install path C (`go install`)**, which still says you only need `WH_CH_ADDR` reachable — a bare `wavehouse` in an empty directory now dies at config validation before it ever touches ClickHouse.

Related: §Loading Order says "all values are read from environment variables (with defaults)", which isn't true anymore now that one key has no default.

internal/api/router.go
```diff
@@ -307,6 +307,24 @@ func corsMiddleware(allowedOrigins []string) func(http.Handler) http.Handler {
...
 			}
 
+			// Resolved per request (not captured at router build) so a settings
+			// reload changes the allowlist without a restart. The lists are a
+			// handful of origins, so a linear scan beats rebuilding a set.
+			var allowedOrigins []string
+			if origins != nil {
+				allowedOrigins = origins()
+			}
+			allowAll := len(allowedOrigins) == 0
```
EricAndrechek, line 315: Flagging a pre-existing footgun rather than a regression — to be clear, this isn't something you introduced. `main` has the identical `allowAll := len(allowedOrigins) == 0` at the top of `corsMiddleware`; this PR only moves the resolution per-request so reloads take effect. But I think this change makes it materially easier to step on, so I'd like it closed here rather than left for later.

The behavior: an empty list doesn't mean "no cross-origin access", it means allow every origin. Driving this middleware directly with `Origin: https://evil.example`:

```
empty slice []      -> Access-Control-Allow-Origin: "*"
nil slice           -> Access-Control-Allow-Origin: "*"
["*"]               -> Access-Control-Allow-Origin: "*"
explicit allowlist  -> Access-Control-Allow-Origin: ""   (rejected)
```

What changes here is reachability. On main the key carried `env-default:"*"` and `config.yaml` shipped `- "*"` explicitly, so omitting it gave you a deliberate-looking allow-all. Now `cors.allowed_origins` is a **required** key in a hand-edited JSON file, so an operator has to type something — and `[]` is the natural spelling for "no origins allowed". `validate.go` only rejects `null` (nil → required error, `[]` → passes clean), so that misconfiguration validates green. Then, because this is now hot-reloadable, it takes effect immediately with no restart to catch it.

Two suggestions:

- A warning in `validate.go` next to the existing ones — you already warn on much lower-stakes things ("override sets nothing", "admin is an unconditional bypass"), so something like `cors.allowed_origins: empty list allows every origin — write ["*"] to say so explicitly` is consistent with the surrounding validator.
- The `cors.allowed_origins` row in `configuration.mdx` documents `"*"` and says to "tighten to your frontend's exact origin(s)", but never says that `[]` is the wrong way to tighten. Worth stating that `[]` and `["*"]` are equivalent.

Also worth noting `router_test.go` has no coverage of the empty list in either direction, so nothing currently pins this behavior.

EricAndrechek (changes requested):

deployments/compose/standalone.yaml
taitelee: I made a split on this based on ownership and lifecycle rather than what simply looks like configuration. Everything left in the environment variables controls resources that run for the entire process. The dedupe variable manages the Pebble store, DLQ variable creates the NATS stream, and the data directory and CH variables define where state lives. These need a full restart to change, so I kept them here The policy file path and pipes directory are the only exceptions. They are legacy seed paths that I'll remove in the next PR.

deployments/Dockerfile.goreleaser
taitelee: Looking at this again I don't really think there's a reason. The plain `Dockerfile` already had the source at `/src` from `COPY . . ` so I just used `cp`, and the goreleaser one only gets `extra_files` so it had to be a `COPY`. Changed the plain `Dockerfile` to use the same `COPY internal/settings/seed /app/settings`. ALs omade note that they're the same.

internal/settings/seed.go
taitelee: There is only one source for the seed. It uses the files under `internal/settings/seed`. Two things use this folder. The binary embeds the files directly. This allows the `init-settings` command to write them anywhere without needing the source tree. The container images just copy the exact same directory. I could not run `init-settings` during the image build. The build platform creates the goreleaser image but the target platform compiles the binary. Running it during the build would require QEMU. I wrote the Dockerfile specifically to avoid that. The only real inconsistency was how the two Dockerfiles copied the files. I fixed that (comment somewhere above). They both use the exact same copy line now. I also updated the comment to explain this setup instead of saying it ships three ways (on next push).

docs/src/content/docs/configuration.mdx
taitelee: Working on now.

docs/src/content/docs/configuration.mdx
taitelee: Yes I agree on this. Will do on next push.

docs/src/content/docs/configuration.mdx
taitelee: Fixing these in next push. The example block and the env tab will use `./settings` with the `init-settings` command next to it, since the seed path only makes sense inside a repo checkout. For `.config.local.yaml`, the `Makefile` rule was already re-running whenever `config.yaml` was newer, it just didn't do anything in that case, so it will print a warning telling you to port the change over or delete the local copy to re-seed it. I have `development.md` updates calling out `settings.dir` as the key that will stop boot if it's missing. The `README` go install path will have the `init-settings` and `WH_SETTINGS_DIR` step before the ClickHouse note. Loading Order step 3 will note every key has a default except `settings.dir`.

internal/api/router.go
taitelee: Did both. The validator warns on an empty list (empty list allows every origin — write ["*"] to say so explicitly, or list the origins to allow) with a test alongside the other warning cases, and the docs row says [] is the same as ["*"]. Made edits to `TestCORSMiddleware_EmptyOrigins_AllowAll` to cover a nil getter, a nil slice, and an empty slice. I didn't go as far as making `[]` mean deny-all. That would also change what a nil getter does on main, and because it is security relevant I think it should have it's seaprate PR.

docs/src/content/docs/configuration.mdx
taitelee: `dedupe.enabled` now moved into `config.json` with the rest of the dedupe block and hot reloads like the others. A small `dedupe.Managed` wrapper owns the Pebble store, and the settings store got an `AfterAdopt` hook that opens or closes it after each successful reload. Seen ids survive an off/on cycle since it's the same directory on disk. Two edge cases worth knowing about: at boot a failed open still refuses to start, same as every other store, and if the open fails during a reload it's logged and ingest returns 500 until the next reload. The files asked for dedupe, so quietly publishing un-deduped didn't seem like an acceptable fallback. `WH_DEDUPE_ENABLED` is gone from boot config. There are no other Pebble knobs today. If we add any they'd go in the same block.

internal/api/router.go
EricAndrechek: Ok I agree, can you make an issue to track this so we can get it done in another PR?

taitelee: Yes #515

docs/src/content/docs/configuration.mdx
```diff
@@ -37,7 +39,8 @@ export WH_CONFIG=/etc/wavehouse/config.yaml
 | --- | --- | ------- | ----------- |
 | `server.port` | `WH_SERVER_PORT` | `8080` | HTTP server listen port. |
 | `server.shutdown_timeout` | `WH_SERVER_SHUTDOWN_TIMEOUT` | `10` | Graceful shutdown timeout in seconds. |
-| `server.cors_allowed_origins` | `WH_SERVER_CORS_ALLOWED_ORIGINS` | `*` | Comma-separated list of allowed CORS origins. `*` allows any browser origin. WaveHouse is a Bearer-token API — `Access-Control-Allow-Credentials` is intentionally never sent, so this allowlist controls *which origins can read responses*, not cookie scope. Tighten to your frontend's exact origin(s) in production (e.g. `https://dashboard.example.com,http://localhost:3000`). |
+
+The CORS allowlist is a [settings-directory](/settings-directory) tunable (`cors.allowed_origins` in `config.json`), so it reloads without a restart.
 
 The server speaks **plain HTTP** — there is no inbound-TLS setting. Terminate TLS at a [reverse proxy](/reverse-proxy#tls) for internet-facing deployments. (The `clickhouse.http_scheme` option below is the *outbound* WaveHouse → ClickHouse hop, unrelated to your clients' TLS.)
```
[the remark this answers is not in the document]
taitelee: Reworded in `settings-directory.mdx` (that row moved when the docs split): 'to restrict browser origins, list only the ones you allow.'

taitelee:
@coderabbitai review

internal/settings/watch.go
```diff
@@ -0,0 +1,96 @@
...
+// errors are logged and the loop continues — SIGHUP and the ops reload
+// endpoint remain as triggers even if the watcher degrades.
+func (s *Store) Watch(ctx context.Context) error {
+	w, err := fsnotify.NewWatcher()
+	if err != nil {
+		return fmt.Errorf("settings watcher: %w", err)
+	}
+	defer func() { _ = w.Close() }()
+	dir := filepath.Clean(s.dir)
+	if err := w.Add(dir); err != nil {
```
EricAndrechek, line 47: We only start listening here, but we read the files way earlier in `settings.Open`. Anything edited in between fires no event, so it just never gets picked up — no log, nothing. That gap has the whole ClickHouse dial in it.

The case that worries me is a ConfigMap update landing during a rolling restart. One burst, nothing touches the dir again, so those pods sit on stale settings until someone bounces them, and they all look healthy the entire time.

Can we just `TriggerReload("watch")` right after the `Add`? `watch_test.go` already retries the write in a loop to dodge this same race, so we kind of already know about it.

cmd/wavehouse/main.go
```diff
@@ -310,17 +353,33 @@ func run() int {
...
 		config.WarnIfFreshDataDir(logger, "pebble", pebbleDir)
-		dedup, err = dedupe.NewEmbedded(pebbleDir)
-		if err != nil {
+		if err := dedup.Apply(true); err != nil {
 			config.LogStorageInitError(logger, "dedupe", pebbleDir, err)
 			return 1
 		}
-		defer func() { _ = dedup.Close() }()
 	}
+	settingsStore.AfterAdopt(func() {
```
EricAndrechek: Can this move above the boot apply? The read at 365 isn't under `s.mu`, so if a reload lands between there and here there's no hook registered yet and we end up with the settings saying dedupe is on while Managed is still off.

Different problem from the watch.go one I commented on — there the settings never make it into memory, here they do and we just don't act on them.

Window is tiny so I'm not holding the PR on it, but worth flagging that `ed5c9ce` made it quieter: that state used to be a wall of 500s, now it takes the new ErrDisabled path and silently publishes un-deduped instead. `Apply` is idempotent so hoisting this and having boot call the same closure should be fine, no?

internal/api/ingest.go
```diff
@@ -422,29 +425,41 @@ func (h *IngestHandler) processRecord(
...
+			} else {
+				eventID := fmt.Sprint(idVal)
+				dup, err := h.Dedup.CheckAndMark(ctx, eventID)
+				switch {
+				case errors.Is(err, dedupe.ErrDisabled):
+					// A reload flipped dedupe.enabled between the snapshot
+					// read above and this call (the two transition at
+					// different instants). Publish un-deduped, as a record
+					// under the other setting would have been.
+					h.logger.WarnContext(ctx, "dedupe switched off mid-reload; publishing without idempotency", "event_id", eventID, "table", table)
```
EricAndrechek: Can we get a counter next to this like the missing-id case does at 437? WARN exports at 100%, so a busy ingest crossing a reload is a pile of lines with nothing aggregating them.

The bigger reason though — if that main.go ordering race ever latches, this stops being transient and goes unbounded, every record forever, and a counter is the only way you'd ever notice. Probably drop the log line to Debug once the counter is carrying it.

EricAndrechek (changes requested):

internal/settings/watch.go
taitelee: Yes. This actually fixes the race condition that would occur in `main.go` since the watcher used to start before the dedupe hook was registered. Moving the reload trigger to start after every `AfterAdopt` hook is in place.

cmd/wavehouse/main.go
taitelee: Yes. There is only one `reconcileDedupe` closure now. I registered it with `AfterAdopt` before the boot apply calls it. Either the hook sees a reload in the middle or the boot apply reads it. There is no race window either way. I moved the SIGHUP and watcher goroutines below this block for the same reason as the previous comment.

internal/api/ingest.go
taitelee: Adding `wavehouse_ingest_dedupe_disabled_total{table}` next to the missing-id counter and dropping the line to Debug

taitelee:
The configuration file is only for boot settings now. I removed the intro line about tunables and the CORS sentence. I also deleted the Query, Schema Discovery, and Deduplication sections. The settings directory row is the only mention of settings left in the document. I moved all the details from those old sections into the settings directory file, includes the notes on result shaping, resource limits, refresh cycles, and the on demand endpoint.

taitelee:
@coderabbitai review

internal/api/ingest.go
```diff
@@ -60,6 +63,15 @@ var dedupeMissingIDCounter, _ = otel.Meter("wavehouse-ingest").Int64Counter(
...
 )
 
+// dedupeDisabledCounter counts records published un-deduped because the
+// settings snapshot said dedupe was on while the store was switched off —
+// transient across a reload; a climbing rate means the store and the
+// settings have come apart.
+var dedupeDisabledCounter, _ = otel.Meter("wavehouse-ingest").Int64Counter(
+	"wavehouse_ingest_dedupe_disabled_total",
+	metric.WithDescription("Ingested records published without dedupe because the store was switched off while settings said enabled (reload window)"),
+)
```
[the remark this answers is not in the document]
taitelee: Keeping as-is: this matches the existing convention for OTel instruments here — `dedupeMissingIDCounter` directly above (on main) and `auth.operatorKeyFailures` are package-level for the same reason. OTel instruments are process-global by design; the injected seam is the meter provider set at boot. Moving all of them behind a constructor dependency would be its own refactor.

README.md
EricAndrechek, line 1: So something I realized while trying to deploy stuff on WaveHouse-Stats is that this new PR is going to break our documented quickstart stuff… We have it here in the README at :39-:39 and :101-:102, but also in lots of places in the docs I think… Some I notice are fixed I think, like in `getting-started` and `deployment`, but not sure how many other files or places may need updating, like maybe `index` which is the landing page and I think shows some getting started stuff too?

You already have the fix documented at those two doc places, with:
```bash
docker compose -f deployments/compose/standalone.yaml run --rm wavehouse init-settings /app/settings
```
But I'd love for you to make sure it's documented everywhere, and, crucially, that that actually WORKS – like run it yourself, manually, not telling claude to do it, so that you can see what you get hung up on if anything to make sure its well documented and easy to run as a quickstart should be.

deployments/compose/standalone.yaml
EricAndrechek: Can the DLQ setting (enabled or not) not move into the hot-reload piece too? Like it would presumably be on per-tenant/table, no?

taitelee:
@coderabbitai review

docs/src/content/docs/configuration.mdx
```diff
@@ -349,15 +308,13 @@ WH_AUTH_JWKS_URL=
...
-WH_SCHEMA_REFRESH_INTERVAL=60
-
 WH_DLQ_ENABLED=true
 
 WH_POLICY_FILE_PATH=
 
 WH_PIPES_DIR=
 
-WH_SETTINGS_DIR=
+WH_SETTINGS_DIR=./settings  # required, no default (`wavehouse init-settings ./settings`); the container images preset /app/settings; you mount a directory there
```
EricAndrechek: I could be mistaken, but I don't _think_ that docker likes comments inline on an env file, so if someone tried to copy-paste this and run it exactly with this example it would fail. I think.

docs/src/content/docs/deployment.md
EricAndrechek: Do we need to add anything about the settings directory/volume stuff in here?

docs/src/content/docs/development.md
```diff
@@ -228,12 +228,16 @@ curl -s -X POST http://localhost:8080/v1/ops/query \
...
+`make dev` points `settings.dir` at the checked-in seed, which ships with `dedupe.enabled: false`. Write your own settings directory and flip it there (the seed is committed — don't edit it in place):
 
 ```bash
-WH_DEDUPE_ENABLED=true WH_DEDUPE_ID_FIELD=event_id make dev
+./tmp/wavehouse init-settings ./settings   # or: go run ./cmd/wavehouse init-settings ./settings
+# set "enabled": true under "dedupe" in ./settings/config.json,
+# and settings.dir: ./settings in .config.local.yaml
 ```
 
+The key hot-reloads, so once the server is running you can toggle it by editing `config.json` — no restart. Records dedupe on their `event_id` field by default; the same file overrides the field globally or per table (see [Configuration — Deduplication](/settings-directory#deduplication)).
```
EricAndrechek: You have the new URL here for settings directory stuff, but the link still says `[Configuration – Deduplication]` – is tat intended?

docs/src/content/docs/getting-started.md
```diff
@@ -113,4 +115,4 @@ The handful of things that most often trip up a first session — each is expect
 ## Going further
 
 - **Validate JWTs**: set `WH_AUTH_JWT_SECRET=<secret>` (the middleware always runs; without a secret every request is the policy `default_role`) and replace the shipped trial policy (`deployments/compose/dev-policy.yaml`) with a least-privilege one — see [API Reference — Authentication](/api#authentication) and [Access Control](/access-control).
-- **Enable deduplication**: set `WH_DEDUPE_ENABLED=true` and `WH_DEDUPE_ID_FIELD=event_id` — see [Configuration — Deduplication](/configuration#deduplication).
+- **Enable deduplication**: set `dedupe.enabled` to `true` in the settings directory's `config.json` (it hot-reloads, no restart) — records dedupe on their `event_id` field by default; pick a different field (globally or per table) in the same file — see [Configuration — Deduplication](/settings-directory#deduplication).
```
EricAndrechek: Same thing here, you updated the link URL but left the link title stale

docs/src/content/docs/settings-directory.mdx
```diff
@@ -0,0 +1,73 @@
...
+
+Check a directory with `wavehouse validate [dir]` — the argument falls back to `WH_SETTINGS_DIR`. It validates without starting the server (JSON syntax including unknown fields and duplicate keys, per-file shape rules including the required keys, and cross-file role references), prints every finding in one pass, and exits `0` for valid (warnings allowed), `1` for invalid, `2` for usage — so operators and CI can gate a settings change before it reaches a running instance.
```
EricAndrechek: In some places, like a few lines above, you say "Create one with `wavehouse init-settings <dir>`" but here you say "Check a directory with `wavehouse validate [dir]`".
In my mind, validate and bootstrap commands should be synonymous, and so you should be able to replace `init-settings` with `validate` and have the same dir passed to both, but it's weird that we vary between `<dir>` and `[dir]` – pick one and stick with it consistently EVERYWHERE – likely will need changes in more places than just this file here where I noticed it.

EricAndrechek: Similarly, they should both use the same env variables to write a bootstrap settings dir/files or validate a settings dir/files if specified, and have the same default behaviors if no path or env variable is defined.

internal/config/config.go
```diff
@@ -117,10 +104,11 @@ type OTelLogs struct {
 	SampleRate float64 `yaml:"sample_rate" env:"WH_OTEL_LOGS_SAMPLE_RATE" env-default:"1.0"`
 }
```
EricAndrechek, line 99: Open to debate on this one: do you think o11y config and prometheus and stuff should be a boot-time operator config, or a hot-reloadable config? My thinking is that we want to make everything we possibly can a hot-reloadable config, as it's a lot easier to track and commit and validate then, but that there is also possible justification to say that this o11y stuff is genuinely harder to fit in as hot reloadable. On the flip side, though, it may genuinely be the case that tenants want to be able to control their own metrics endpoints and logs and sample rates, etc., making it justify the effort. What do you think?

internal/config/config.go
EricAndrechek: I think streaming should go in hot reloadable too. It is something that very likely may need to be tweaked during operation to see how frequently keep alive needs to happen, or may need to change the bucket number to address growing load during operation, although I can see how the latter may be difficult to do hot-reloadable, so open to feedback there. But I think that it is likely that different tenants have things behind different proxies and whatnot that may have different requirements on keep alive.

EricAndrechek, line 116: I think this one should definitely be hot-reloadable, what do you think? It is entirely likely a ClickHouse instance goes down or you need to migrate to another one, while still accepting ingest traffic durably. Not to mention different tenants WILL have different clickhouse configs or databases or username/passwords, etc.

internal/config/config.go
```diff
@@ -150,12 +138,6 @@ type MQ struct {
 	MaxBytesGB       int `yaml:"max_bytes_gb" env:"WH_MQ_MAX_BYTES_GB" env-default:"50"`
 }
```
EricAndrechek: Can this not be hot-reloadable too? How would that work?

internal/config/config.go
EricAndrechek, line 123: Why can't this be hot-reloadable and in settings config?

EricAndrechek: I think this may be one of if not the most frequently changed settings in here, especially for new deployments setting up and trialing auth, and should absolutely be hot-reloadable and able to be validated, etc, no? I will note that we may want a new issue to track how to store the secrets here, namely `jwt_secret` and `operator_key`, which presumably you wouldn't want in plaintext in a tracked file...

EricAndrechek: Can this not be hot-reloadable too?

EricAndrechek, line 234: Not sure how or best way to do it, but when we read in the config (probably can't/shouldn't do it for env variables read in, since there will always be ones unrelated to us set), it would be nice if there was a way we could loudly assert/fail when unused values were set (so like if someone still was setting dedupe settings in a yaml/json this read in), or potentially if required ones were unset, etc, but may be more work given we combine with env too, etc – likely a later PR but may worth an issue to track as well, so we can have validate working loudly on the non-hot-reloadable startup ops config too.

EricAndrechek (changes requested):
My required changes from you that aren't in per-file reviews:
- [ ] Need to update/add the new settings package you've added for `architecture.md` and to `AGENTS.md` probably too, etc
- [ ] I definitely should have mentioned this sooner, and for that I'm very sorry, but I don't know if I like the command being `init-settings` – having a dash in there means it splits as two words when you try and terminal option-backspace and stuff, I'd rather it be one word cleanly. What do other popular projects that ship with an option like this use? I am immediately thinking of options like `init`, `bootstrap`, `defaults`, etc but open to pushback or other ideas etc here.
- [ ]  I keep asking but not really hearing about it, so I finally added them all as individual comments… but basically why were the things you had in config all you had? I know pipes and policy are going to come in a follow-up PR, but there are so many other settings things that I think should be there and not the config stuff too that I'd like to at least discuss. Happy to take offline and discuss them all in-person if easier.

Some other/minor nits claude caught for me, verbatim:
- [ ] ` architecture.md untouched: AGENTS.md §Documentation Sync requires it for "add a package". AGENTS.md:36 was updated for dedupe.Managed; the page it points at wasn't. Stale: :54-67 package tree omits settings/; :111-114 lists only dedupe.go/embedded.go, missing managed.go; :74 omits /v1/ops/settings/reload; :109 still says "All settings use WH_ prefixed env vars".`
- [ ] `PR body is stale — and it becomes the squash commit message: Still says the seed is baked into "both container images at /app/settings". 1593fe5 reversed exactly that (Dockerfile:33-41 now creates a deliberately empty mount point), and the e2e fixture uses its own copy. Merging as-is writes a commit message contradicting its own code.`
- [ ] `The dev settings-dir story dirties the tree and voids the make ci marker: config.yaml:93 points the live, watched dir at the checked-in go:embed seed (internal/settings/seed) — make dev inherits it, so any local experiment edits a tracked file and silently changes what init-settings writes for everyone. The documented escape (development.md:231-236, init-settings ./settings) isn't in .gitignore either, so it leaves four untracked files that change the tree SHA scripts/ci-marker.sh:19-27 computes via git add -A. → Point the dev default at ./tmp/settings (already ignored) or add /settings/ to .gitignore, and have make dev seed via init-settings.`
- [ ] `deployment.md:185, :212 — /app/settings missing from the mount contract: docs "Pre-create /app/data and /app/pipes" is now incomplete (both Dockerfiles also pre-create /app/settings, and the copy-up mechanic at :212 is what makes the named volume land with UID 65532). The whole "Persistent Storage (REQUIRED for containers)" section never mentions it — despite a missing mount there now being the #1 startup failure.`
- [ ] `settings-directory.mdx:70 — "or the reverse" over-claims the counter. Managed.CheckAndMark returns ErrDisabled only when the store's switch is off (managed.go:60-65), the only path that increments (ingest.go:455-462). In the reverse case the handler skips the dedupe block entirely (ingest.go:441) — un-deduped but not counted.`

README.md
taitelee: I did run this. There's no more seed step. `standalone.yaml` mounts `deployments/compose/settings/` (the seed with clickhouse.addr already pointed at the clickhouse service), so the quickstart is just up -d again. This should be documented everywhere you can check on your next review. 

Now run: 
```bash
docker compose -f deployments/compose/standalone.yaml run --rm -v wavehouse-settings:/app/settings wavehouse bootstrap /app/settings
```

Making note here that the command is not `init-settings` and is now `bootstrap`.

Running the above should output something like: `roles.json, policies.json, pipes.json, config.json to /app/settings`. The `-v` points `/app/settings` at an empty named volume for that one run. Without it, `/app/settings` is the checked in bind mount, which already has the four files, and bootstrap refuses with `/app/settings` is not empty and will refuse to overwrite an existing settings directory.

docs/src/content/docs/configuration.mdx
taitelee: Fixed.

docs/src/content/docs/deployment.md
taitelee: Added.

docs/src/content/docs/development.md
taitelee: That was outdated. Fixed now.

docs/src/content/docs/getting-started.md
taitelee: Fixed.

docs/src/content/docs/settings-directory.mdx
taitelee: Both things should be fixed. Both `bootstrap` and `validate` should now use `[dir]`. There are no mentions of `<dir>` anymore at all in the codebase. The same env variable should be used now: `WH_SETTINGS_DIR`

internal/config/config.go
taitelee: Yes, and fixed. Let me know if there's anything I missed on this one.

internal/config/config.go
taitelee: Keeping this as is (boot time) for now. The exporter endpoints already come from the OTel SDK's own `OTEL_EXPORTER_OTLP_*` env vars, not our config, and the enabled flags and the Prometheus port all mean rebuilding the provider chain or rebinding a listener, which is the same class of thing as changing `server.port`. The two sample rates are the exception since they're read per span or log record, so having those in `config.json` isn't breaking really. I could bring this up in a follow up or try tackling this in this PR.

internal/config/config.go
taitelee: Made changes based on this. Needs re-review.

internal/config/config.go
taitelee: Moved it. `mq.max_bytes_gb` is now in `config.json` and a reload updates both the ingest and DLQ stream limits in place. After each confirmed change (after validate runs) a hook calls JetStream's `UpdateStream` with the new `MaxBytes`, which NATS applies to the running stream without touching what's buffered.

internal/config/config.go
taitelee: I moved `timestamp_bucket_seconds` and it's now `query.timestamp_bucket_seconds` in `config.json`, read per query. Didn't move `l1_max_cost` since Ristretto can't resize in place, so a live change would mean dropping the whole cache and rebuilding it, and it's process memory sizing for the host rather than something a tenant should be setting. Same category as `data_dir`.

taitelee: Yep. Moved everything that isn't a secret: `jwks_url` and `role_claim` (`jwks_url` must be an absolute http(s) URL, `role_claim` a non-empty claim path), validate and a reload swaps the whole verifier atomically, so trialing auth on a new deployment is a file edit. `jwt_secret` and `operator_key` stay. Yes on the issue, I'll open one for how secrets get stored and rotated (env only vs `*_file` references vs something else) and mention it here

taitelee: I removed them. But please look over. These were the paths to the policy YAML and `.sql` pipes directory that seeded NATS KV on first boot. I made `policies.json` and `pipes.json` runtime priority, so the keys, the KV buckets, and the `PUT/DELETE` endpoints are all removed.

taitelee: Did the half that's cheap here: Load re-reads the YAML against the struct tags and refuses to boot naming every undeclared key, so a leftover dedupe:, mq:, or policy: block can't be silently ignored. Env can't be checked that way. The rest: a validate-style dry run for boot config and required-key checks like `settings.dir`. I can create an issue.

taitelee: #530

taitelee: #529

deployments/compose/standalone.yaml
```diff
@@ -33,29 +35,31 @@ services:
...
+      # (clicks, events) and nothing else (no admin, no raw SQL). Tune it
+      # before production. All four files reload live (file watch / SIGHUP /
+      # POST /v1/ops/settings/reload) — edit them on the host and the running
+      # container picks them up; the files are the only write path.
     volumes:
       - wavehouse-data:/app/data
-      - ./dev-policy.yaml:/app/policy.yaml:ro    # permissive trial policy (public role, demo tables) — tune for prod
-      # - ./my-pipes:/app/pipes:ro    # uncomment alongside WH_PIPES_DIR
+      - ./settings:/app/settings    # checked-in seed + trial policy, clickhouse.addr=clickhouse:9000; hot-reloads on edit
+      # - wavehouse-settings:/app/settings    # or a named volume: swap this line for the
```
EricAndrechek: I tried following this instruction and it doesn't work. Two separate problems.

**1. Compose won't start at all.** If you swap the bind mount for `wavehouse-settings:/app/settings`, Compose also needs that name declared in the top-level `volumes:` list at the bottom of this file. That list currently only has `clickhouse-data` and `wavehouse-data`, so you get:

```
service "wavehouse" refers to undefined volume wavehouse-settings: invalid compose project
```

**2. Even after fixing that, you'd seed the wrong volume.** The seeding command we point people at (`deployment.md:198`) is:

```
docker compose -f deployments/compose/standalone.yaml run --rm -v wavehouse-settings:/app/settings wavehouse bootstrap
```

Compose prefixes the volumes it manages with the project name, so the volume the service actually mounts ends up called `compose_wavehouse-settings`. But a `-v` flag on the command line is passed straight through to Docker, which creates a literal volume called `wavehouse-settings`. Those are two different volumes — so `bootstrap` writes the four JSON files into one, the server mounts the other (empty) one, and it exits with four "missing" errors.

The fix is to add `wavehouse-settings:` to the top-level `volumes:` list, and drop the `-v` from the seed command so it runs *after* the mount swap and writes into the volume the service actually uses.

That said — is this option worth keeping at all? The image is distroless, so there's no shell to go edit `clickhouse.addr` inside the volume afterwards, which is what the next line tells people to do. The bind mount doesn't have that problem. Might be cleaner to just delete the named-volume path.

docs/src/content/docs/settings-directory.mdx
```diff
@@ -0,0 +1,196 @@
...
+
+## `roles.json`
+
+The role registry — every role a policy grant or a pipe allowlist may name:
+
+```json
+{ "roles": ["admin", "analyst", "public"] }
+```
+
+Rules: each name must be non-empty, without surrounding whitespace, and unique. Cross-file, every role referenced elsewhere must be declared here: `policies.json`'s `default_role` and `admin_role`, every role under a table's `select` / `insert`, and every entry in a pipe's `allowed_roles`. An undeclared reference is an error, so a typo in a grant fails validation instead of silently matching nobody. The admin role is exempt only where naming it is redundant (see below).
```
EricAndrechek: This doesn't answer the question someone actually has when writing their first `roles.json`, which is just "do I need to put `admin` in this list?" The "(see below)" also points at two different later sections, so it's not obvious where to look.

The rule in the code (`validate.go:511-513`, `:526-529`, `:546-548`) is:

- If you set `admin_role` in `policies.json`, you have to declare that role here.
- If you leave it at the default, you don't.
- Naming the admin role inside a table grant or a pipe's `allowed_roles` never needs a declaration — that only warns.

Our own `deployments/compose/settings/roles.json` is just `["public"]` and validates fine, precisely because `policies.json` doesn't set `admin_role`. Worth stating the rule outright rather than making people infer it from the exemption.

internal/config/config.go
EricAndrechek: Fair enough makes sense for the time being.

internal/config/config.go
EricAndrechek: Agreed that makes sense to me, I think that split is good.

EricAndrechek (changes requested):
Good work, getting close to mergeable. I think for our future PRs though we NEED to start using stacked PRs because this has been horrid to review and is so massive it's just impossible…

My findings:
- [ ] broken named-volume path
- [ ] dev command fails on a fresh clone
- [ ] some migration prose

**`development.md:281`** — not inline, that line isn't in the diff. The "Running Modes at a Glance" row `make build && ./bin/wavehouse` doesn't work on a fresh clone: `config.yaml:72` sets `settings.dir: settings`, `/settings/` is gitignored, and it's created only by the `settings/config.json` target, which only `dev` depends on. You get `settings directory "settings" does not exist ... refusing to start`. Suggest `make settings/config.json && make build && ./bin/wavehouse`, plus a note at `deployment.md:54`. Same applies to `go run ./cmd/wavehouse` at :263.

deployments/compose/standalone.yaml
taitelee: I agree and will just remove the alternative option.

docs/src/content/docs/settings-directory.mdx
taitelee: Ok adding a comment to say that you only declare an `admin_role` if it is declared in `policies.json`. Also making note that grants/pipe `allowlists` naming admin only warn, with a link to the admin-bypass section.

docs/src/content/docs/configuration.mdx
```diff
@@ -110,83 +91,49 @@ Set the backstop on the profile of the ClickHouse user WaveHouse connects as (`c
...
 | --- | --- | ------- | ----------- |
-| `settings.dir` | `WH_SETTINGS_DIR` | *(empty)* | Path to the settings directory. Its one consumer is the `wavehouse validate` CLI, which — like `wavehouse health` — resolves its input directly from the `[dir]` argument or `WH_SETTINGS_DIR` without loading this file, so a value set only in the YAML key is not picked up. No default, same reasoning as `policy.file_path`: a baked-in path would turn a missing mount into silent misconfiguration instead of an explicit operator choice. |
+| `settings.dir` | `WH_SETTINGS_DIR` | *(required)* | Path to the settings directory. The server loads it at boot; the `wavehouse validate` and `wavehouse bootstrap` CLIs — like `wavehouse health` — resolve their input directly from the `[dir]` argument or `WH_SETTINGS_DIR` without loading this file, so a value set only in the YAML key validates on boot but not via the bare CLI. No default: a baked-in path would turn a missing mount into silent misconfiguration instead of an explicit operator choice. The container images preset it to `/app/settings` but ship no files there — mount a directory (or seed a volume with `bootstrap`). |
```
EricAndrechek: One leftover from dropping the named-volume path: "(or seed a volume with `bootstrap`)" here now contradicts `settings-directory.mdx:14` and `deployment.md:199`, which both say to use a bind mount and specifically not a named volume.

The env-tab copy further down this same page (`:244-245`) already has the corrected wording. Suggest matching it — something like "bind-mount a directory you wrote with `bootstrap`."

docs/src/content/docs/development.md
```diff
@@ -138,16 +134,16 @@ dev: deps-up $(AIR)
...
 - ClickHouse on `http://localhost:8123` (HTTP) and `localhost:9000` (native protocol), Compose project name `wavehouse-dev` so containers/volumes are namespaced.
-- Hot reload: editing any `.go` file under `cmd/` or `internal/` triggers a debounced rebuild + restart. Config isn't hot-reloaded — `make dev` loads `.config.local.yaml` (a gitignored copy seeded once from `config.yaml`), so edit `.config.local.yaml` and restart to apply config changes. Air's stdout/stderr stream live so you see compile errors and server logs in the same terminal.
+- Hot reload: editing any `.go` file under `cmd/` or `internal/` triggers a debounced rebuild + restart. Boot config isn't hot-reloaded (settings-directory files are) — `make dev` loads `.config.local.yaml` (a gitignored copy seeded once from `config.yaml`), so edit `.config.local.yaml` and restart to apply config changes. Air's stdout/stderr stream live so you see compile errors and server logs in the same terminal.
 
 ### Dev convenience targets
```
EricAndrechek, line 148: Small one, in the paragraph just below (`:150`): "useful directly when you want to run WaveHouse outside of air (e.g. `make build && ./bin/wavehouse`)".

This is the last bare-binary invocation in the docs that doesn't seed first, and it's the one a contributor on a fresh clone is most likely to hit — `config.yaml:72` points `settings.dir` at the gitignored `settings/`, which only `make settings/config.json` creates. It also now disagrees with the Running Modes table at `:282`, which this PR just corrected.

Same three-part command here would do it.

EricAndrechek (dismissed):
Ok three small missing doc things, 2 in git diff and one not so I put it here:

- [ ] `deployments/compose/standalone.yaml:3` is still `clickhouse/clickhouse-server:latest`, but you pinned `26.6.3.62` in the two `ci.yml` prefetches, `dependencies.yaml`, `scripts/orchestrator/main.go` and `tests/integration/setup_test.go`.

Going to approve though so that you can fix them and just merge and be done without another PR review

EricAndrechek (dismissed):

taitelee:
@coderabbitai review

EricAndrechek (approved):

PR #541, opened 2026-09-01
fix(policy): reject no-op rules, drop the policy HTTP surface

### LINKED ISSUE #460: security(policy): mistyped or unhonored policy rules load silently and disable row security

### LINKED ISSUE #514: security(policy): strict-decode policy config so unknown keys are rejected, not dropped
**Area:** policy · config — hardening (fail-open surface) · found via PR #457 review, split out of #460

**Expected:** a policy or role config carrying a key the engine does not recognise is rejected at load, so a misspelling cannot silently become "no rule."

**Actual:** every decoder is non-strict — `internal/policy/store.go:153,185,210,214` and `internal/api/policy.go:46,71` — so an unknown key is dropped rather than rejected. That is the mechanism behind #460: `filter: { tenant_id: { eq: … } }` (should be `_eq`) decodes to a `Filter` with all five operator pointers nil, contributes zero clauses, and disables row security.

**Impact:** the general case of #460's specific one. #460 fixes the symptom that bites hardest (an all-nil `Filter` is now rejected, and `/validate` stops certifying it); this closes the class, so the *next* misspelled key doesn't find a new way through.

## Why this is split out rather than fixed in #460

Strict decoding is a **breaking change to config parsing** with a real migration cost, and #460 flags it as such: any stored policy or bootstrap file carrying a stray or misspelled key currently loads fine and would start hard-failing. `Validate` also runs on KV-loaded policies once #461 lands, so this can take a running deployment down at boot.

Keeping it inside #460 forces the urgent half — a one-character typo silently disabling RLS while `/validate` returns `{"valid":true}` — to move at this half's pace. They are separable: #460's rejection of an all-nil `Filter` does not need strict decoding to work.

## Scope

- `yaml.Decoder.KnownFields(true)` and `DisallowUnknownFields` at the six decode sites above.
- **Warn-then-reject over one release**, per #460's migration note — decide explicitly whether an already-stored bad policy fails boot or loads with a loud warning naming the offending key.
- Sequence after #461, so the validation actually reaches deployments loading from KV.

Related: #460 (the urgent half this was split from), #461 (validation must reach KV-loaded policies first), #243 (offline `policy validate` would surface this at authoring time), #224.

---
_Split out of #460 by the pm-triage routine on 2026-08-25, on Eric's instruction; decode sites verified against `b2eee9d`._

---

## Update (2026-09-01)

Overtaken by events, then finished by PR #541:

- **Five of the six decode sites are gone or strict** via #508: the `internal/policy/store.go` sites were deleted with the KV store, and the settings directory's loader strict-decodes every file (unknown keys, duplicate keys, trailing content, BOM, top-level `null` are all errors).
- **The sixth** — the lenient decoder behind `POST /v1/ops/policy/validate` — is **removed with the endpoint itself** in PR #541, which drops the policy's whole HTTP surface (`GET /v1/ops/policy` and the SDK `wh.policy` namespace go with it): with files as the only write path, the policy is read, edited, and validated where it lives (`wavehouse validate`), and the endpoint's lenient decode had it certifying `{"valid": true}` for documents a reload refuses.
- **Warn-then-reject is moot**: the settings directory is new and unreleased, so there is no installed base of stored policies to hard-fail; #508 shipped hard-reject directly.
- **The #461 sequencing is moot**: NATS KV no longer exists; every adoption runs the full validation.

Note: #508's body said "Closes … #514 …" but GitHub binds one issue per closing keyword, so only #229 auto-closed — this issue stayed open by that accident, which happened to be justified until #541.

taitelee:
## Summary

The policy's entire HTTP surface is removed: `GET /v1/ops/policy`, `POST /v1/ops/policy/validate`, and the SDK's `wh.policy` namespace. Both endpoints were born alongside `PUT /v1/ops/policy` and outlived it when #508 deleted the policy write API — with files as the only write path, the policy is read, edited, and validated where it lives. The dry run had also kept its original lenient decoder while adoption became strict, so it certified `{"valid": true}` for documents a reload would refuse — including #460's `"eq"`-for-`"_eq"` typo, which silently decodes to a filter that disables row security; deleting it removes the last non-strict policy decode site (#514's other five were deleted or made strict by #508). Its replacement is `wavehouse validate`, which enforces strictly more (the cross-file role references a single-document dry run could never see). The break-glass story narrows to match: the operator key still triggers `POST /v1/ops/settings/reload`, whose findings report exactly why a rejected directory was refused; a bad edit still never breaks a running server. The e2e read-modify-write pattern moves from `wh.policy.get()` to reading the harness-owned `policies.json` (`readPolicyFile()` — the file is the adopted policy there, since `setPolicy` fails unless the reload reports adoption). The SDK keeps the policy document types (`Policy`, `TablePolicy`, `RolePermissions`) — they describe `policies.json`; `GET /v1/ops/pipes[/{name}]` is untouched.

`validateRolePerms` gains three rejections for well-formed rules the engine would silently not honor, the fail-open shapes strict decoding can't see (#460): a `filter` entry with no operator (`"tenant_id": {}` — zero predicates, row security off), an operator-less `check` entry, and `filter:` under an `insert:` grant. Every adoption — boot, watch, `SIGHUP`, reload, `wavehouse validate` — funnels through the one `policy.Validate` path, so the checks land everywhere at once. No shipped seed, compose, or fixture policy carries any rejected shape. Docs are synced across the API, access-control, settings-directory, configuration, development, architecture, reverse-proxy, and SDK pages, plus `AGENTS.md` and `CHANGELOG.md`.

## Related Issues

Closes #514
Closes #460


docs/src/content/docs/access-control.mdx
```diff
@@ -329,7 +331,7 @@ Two admin endpoints (both require `admin_role`) read and dry-run it:
...
+Validation decodes the document strictly — an unknown or misspelled key (`eq` for `_eq`) or a duplicate key is rejected, never silently dropped as "no rule" — and rejects every rule the engine would not honor: empty role names, negative limits, [malformed claim templates](#jwt-claim-templating), a `filter` or `check` entry that sets no operator (it would constrain nothing), `filter` under an `insert` grant (inserts are constrained by `check`), and unsupported `check` operators — `_neq`/`_gt`/`_lt`, or `_eq` and `_in` together on one column (but it intentionally allows `default_role == admin_role`, warning instead). The dry-run endpoint runs the identical per-document checks, so it never certifies a document a reload would refuse. `wavehouse validate` runs the same checks on the whole directory — including the cross-file role references the dry-run endpoint can't see — so gate a policy change with it before it reaches a running instance.
```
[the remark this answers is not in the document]
taitelee: Removed POST /v1/ops/policy/validate entirely: with files as the only write path, a draft is checked with wavehouse validate, which does enforce the cross-file role references a single-document dry run couldn't see. The flagged sentence is gone with the endpoint. 

@coderabbitai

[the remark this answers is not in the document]
taitelee: @coderabbitai

taitelee:
@coderabbitai review

docs/src/content/docs/configuration.mdx
```diff
@@ -117,7 +117,7 @@ WaveHouse accepts only the signing algorithms matching the active verifier — `
...
 
 **Operator key (break-glass).** `auth.operator_key`, when set, is a non-JWT credential for the person running the deployment. A request presenting it — in the standard `Authorization: Operator <key>` header (forwarded verbatim by proxies, no collision with Bearer JWTs), or the `X-Operator-Key` alias — is authorized as a full-access platform operator (the whole data plane plus `/v1/ops/*`) without minting a JWT, and it is honored even when no policy is adopted, so an operator locked out by a wiped or broken `policies.json` can still inspect the adopted policy and trigger `POST /v1/ops/settings/reload` after fixing the file. It is matched with a constant-time comparison and is independent of `jwt_secret`/`jwks_url`; it takes precedence over any Bearer token on the same request. Because it is effectively an admin secret, load it from a secret store, serve it only over TLS, and leave it empty on deployments that do not need it.
```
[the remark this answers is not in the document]
taitelee: Fixed. Both flagged passages (and a third in `architecture.md`'s request-flow notes) now say the operator key triggers the reload; policy inspection is file-based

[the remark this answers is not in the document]
taitelee: @coderabbitai

internal/policy/policy.go
```diff
@@ -649,6 +657,10 @@ func validateRolePerms(table, op, role string, perms RolePermissions) error {
 	if perms.MaxMemoryUsage < 0 {
 		return fmt.Errorf("table %q, op %q, role %q: max_memory_usage must be non-negative", table, op, role)
 	}
+
+	if op == "insert" && len(perms.Filter) > 0 {
+		return fmt.Errorf("table %q, op %q, role %q: filter has no effect on insert — use check to constrain inserted values", table, op, role)
+	}
```
EricAndrechek, line 663: **[SHOULD] The symmetric no-op is left open, and it's the fail-open direction.**

This guard rejects `filter` under an `insert` grant. Its mirror image is still accepted and silently ignored: a **`check` entry under a `select` grant**.

`CheckClauses` is consumed only by `internal/api/ingest.go:386`, and `Evaluate` is called with `"insert"` only from the ingest path — so this adopts cleanly and applies **no row restriction on reads at all**:

```json
{ "select": { "viewer": { "check": { "tenant_id": { "_eq": "{{ jwt.tenant_id }}" } } } } }
```

The author believes they've scoped reads to a tenant; every viewer sees every row, with no warning.

That's the same fail-open class as the operator-less `filter` this PR rejects, and arguably the likelier authoring mistake: `check` and `filter` share the `Filter` Go type and the `_eq`/`_in` operators, so confusing them is a one-word slip that silently disables RLS. Note the asymmetry in severity, too — the case this PR fixes fails *closed* (server won't boot), while this one fails *open*.

The fix mirrors the guard above exactly, right next to it:

```go
if op == "select" && len(perms.Check) > 0 {
    return fmt.Errorf("table %q, op %q, role %q: check has no effect on select — use filter to constrain rows read", table, op, role)
}
```

Fair caveat: this pre-exists on `main` and is outside #460's literal Scope list, so a follow-up issue is a defensible alternative — but I couldn't find one tracking it (#460/#463/#514 don't), and the PR's own framing ("the fail-open rule shapes strict decoding can't see") makes this the natural place. Either fix it here or file it before this merges, so it doesn't get lost.

tests/e2e/sdk/admin.test.ts
```diff
@@ -107,11 +99,10 @@ describe("Admin", () => {
...
-      expect(getResult.data).toBeDefined();
-      expect(getResult.data).toHaveProperty("tables");
-      expect(getResult.data).toMatchObject({
+      // setPolicy throws unless the reload reported adopted, so reaching this
+      // line means the server took the document; the file on disk IS the
+      // adopted policy (there is no HTTP read of it).
+      expect(readPolicyFile()).toMatchObject({
         default_role: "viewer",
         admin_role: "admin",
       });
```
EricAndrechek: **[SHOULD] This assertion can no longer fail.**

`setPolicy()` writes `policies.json`; this then reads that same file back with `readPolicyFile()` and asserts on the `default_role` / `admin_role` it just wrote. That exercises `writeFileSync` / `readFileSync` — the server is never consulted. As the comment above concedes, "there is no HTTP read of it", which is exactly why the read-back no longer proves anything.

Contrast the sibling test at line 122, which still proves adoption via a server read-back (`wh.pipes.get(pipeName)`).

Two ways out, either fine:

1. Assert an observable *effect* of adoption — e.g. a tokenless request now resolving as `viewer` against `T.clicks`.
2. Drop the `expect` entirely and let `setPolicy`'s own adoption check (it throws unless the reload reported adopted) be the test.

As written it reads like coverage that isn't there, which is worse than no assertion.

docs/src/content/docs/access-control.mdx
```diff
@@ -322,28 +324,20 @@ The resource limits (`max_rows`, `max_execution_time`, `max_rows_to_read`, `max_
...
-# Dry-run a policy document (admin token required) without adopting it.
-curl -X POST http://localhost:8080/v1/ops/policy/validate \
-  -H "Authorization: Bearer $ADMIN_TOKEN" \
-  -H "Content-Type: application/json" \
-  -d @policies.json
+# Check the edited settings directory before the server sees it:
+wavehouse validate ./settings
 
-# Then write it to the settings directory; the watcher adopts it, or force it:
+# Then let the watcher adopt the change, or force it:
```
EricAndrechek: **[SHOULD] These comments describe an ordering the server can't give you.**

"Check the edited settings directory **before the server sees it**", followed by "**Then** let the watcher adopt the change" — but the directory watch is unconditional (`internal/settings/watch.go`: fsnotify on the directory, 250 ms debounce; started at `cmd/wavehouse/main.go:557`, degrading only if watcher setup fails). A running instance adopts — or rejects — the edit on save, before you get a chance to run `wavehouse validate ./settings` against that same mounted directory. The "then" is backwards.

The prose directly above at line 329 already has the right framing — "gate a policy change with it **before the edit reaches a running instance**" — matching `settings-directory.mdx:16` ("so operators and CI can gate a settings change before it reaches a running instance"). The code comments just need to agree with it:

```bash
# Gate the change in CI, or on the copy you're about to deploy:
wavehouse validate ./settings

# On the running instance: the watcher picks the edit up, or force it:
curl -X POST http://localhost:8080/v1/ops/settings/reload \
  -H "Authorization: Bearer $ADMIN_TOKEN"
```

docs/src/content/docs/access-control.mdx
EricAndrechek: We now tell the reader to write a policy that won't boot.

> A role with no entry under the operation it is attempting is denied. The set of permission fields below applies to both `select` and `insert` entries, though some (row `filter`, aggregations, limits) **only take effect on the read path** and some (`check`) only on the write path…

That was true before this PR. It isn't now — `internal/policy/policy.go:662`, added here, rejects `filter` under an `insert` grant outright:

```go
if op == "insert" && len(perms.Filter) > 0 {
    return fmt.Errorf("... filter has no effect on insert — use check to constrain inserted values", ...)
}
```

So `filter` inside an `insert` grant is no longer accepted-and-ignored: `policy.Validate` fails, the server **refuses to boot** (`cmd/wavehouse/main.go:171`), and a reload is rejected (pinned by `internal/settings/validate_test.go:249`). This is the framing paragraph for the whole "Anatomy of a policy" section — the one place a reader learns which fields may appear where — so following it produces a server that won't start.

The rest of the clause is still accurate: `allowed_aggregations` / `denied_aggregations` / `max_*` under `insert` *are* still accepted and ignored (`validateRolePerms` applies no op gate to them). Only `filter` needs carving out, e.g.:

> …though aggregations and limits only take effect on the read path and `check` only on the write path — and row `filter` is *rejected* outright under an `insert` grant (see [Row-level security](#row-level-security)) — the [enforcement table](#where-each-rule-is-enforced) maps which is which.

EricAndrechek (changes requested):

docs/src/content/docs/access-control.mdx
taitelee: Rewritten.

docs/src/content/docs/access-control.mdx
taitelee: Rewritten to split the fields: filter/check are operation-specific and rejected at validation under the wrong operation, while aggregations and the `max_*` limits are accepted under both but only take effect on reads.

internal/policy/policy.go
taitelee: Fixed so `validateRolePerms` rejects check under a select grant, placed directly beside the insert-filter guard with the matching error shape ("check has no effect on select — use filter to constrain visible rows")

tests/e2e/sdk/admin.test.ts
taitelee: Removed e2e read-back expect; the test now relies on `setPolicy` throwing unless the reload reports adoption.

EricAndrechek (approved):

taitelee:
@coderabbitai review

PR #546, opened 2026-09-01
fix(cors): make cors.allowed_origins [] deny every browser origin

### LINKED ISSUE #515: CORS: make `cors.allowed_origins: []` mean deny-all instead of allow-all
## Problem

`corsMiddleware` (`internal/api/router.go`) treats an empty or nil allowlist as allow-all:

```
empty slice []      -> Access-Control-Allow-Origin: "*"
nil slice           -> Access-Control-Allow-Origin: "*"
["*"]               -> Access-Control-Allow-Origin: "*"
explicit allowlist  -> Access-Control-Allow-Origin: ""   (rejected)
```

This predates #508 (`main` has the same `allowAll := len(allowedOrigins) == 0`), but #508 makes it easier to step on: `cors.allowed_origins` is now a required key in a hand-edited `config.json`, so `[]` is the natural spelling for "no origins allowed", and the key is hot-reloadable, so the misconfiguration takes effect immediately with no restart to catch it.

#508 mitigates with a validator warning (`empty list allows every origin — write ["*"] to say so explicitly, or list the origins to allow`), a docs note in `settings-directory.mdx` that `[]` is equivalent to `["*"]`, and `TestCORSMiddleware_EmptyOrigins_AllowAll` pinning nil getter / nil slice / empty slice. The behavior change itself was deferred because it is security-relevant and also changes what a nil getter does.

## Scope

- `[]` → no `Access-Control-Allow-Origin` header (deny all browser origins); `["*"]` stays the only allow-all spelling.
- Decide what a nil getter / nil slice does (no settings source wired) — probably deny as well.
- Replace the validator warning with whatever still applies, update the `cors.allowed_origins` row in `settings-directory.mdx`, and flip `TestCORSMiddleware_EmptyOrigins_AllowAll` to pin the new behavior.
- CHANGELOG entry.

Origin: review thread on `internal/api/router.go` in #508.

taitelee:
## Summary

`corsMiddleware` treated an empty or nil `cors.allowed_origins` as `["*"]`, so the natural spelling for "no origins" in a hand-edited, hot-reloadable `config.json` silently opened the API to every origin. #508 could only warn about it.

- `[]` is now an empty allowlist: no CORS headers to any origin, preflights get a bare `204`, `Vary: Origin` still emitted so a shared cache can't replay the reject after a later reload allows an origin. A nil getter or nil slice (no settings source wired) fails closed the same way. `["*"]` is the only allow-all spelling.
- The `[]` validator warning is removed, the `settings-directory.mdx` row and the store comment are updated, and `TestCORSMiddleware_EmptyOrigins_AllowAll` is flipped to `_DenyAll` covering GET and OPTIONS.
- The seed and compose settings still ship `["*"]`, so `wavehouse bootstrap` output is unchanged.

## Related Issues

Closes #515


taitelee:
@coderabbitai review

internal/api/router.go
```diff
@@ -272,11 +272,14 @@ func RequireAdmin(store policy.Source, logger *slog.Logger) func(http.Handler) h
...
-//     API because no credentials are required to read the response.
+//   - allowedOrigins contains "*": any origin is accepted. The response
+//     echoes "*" (no Vary), which is correct for our Bearer-auth API because
+//     no credentials are required to read the response.
 //   - Otherwise: only origins in the allowlist receive Access-Control-Allow-Origin
-//     (echoed back, with Vary: Origin so caches key on it).
+//     (echoed back, with Vary: Origin so caches key on it). An empty or nil
+//     list — including a nil getter, i.e. no settings source wired — is an
+//     empty allowlist: every browser origin is denied. "*" is the only
+//     allow-all spelling, so [] can't hot-reload into allow-all by accident.
```
EricAndrechek, line 283: This block is now correct, but its sibling didn't get the same treatment: the doc comment on the exported `Dependencies.CORSOrigins` field at `internal/api/router.go:41` still reads

```go
// Nil func, an empty list, or ["*"] all mean allow-all.
```

which is precisely the semantics this PR removes. That's the comment a caller wiring up `Dependencies` actually reads, and it tells them leaving `CORSOrigins` nil is allow-all — the exact trap #515 exists to close. No runtime effect, but it's a stale statement of a security invariant sitting on the exported API.

Suggested replacement:

```go
// production). An empty or nil list — including a nil func — denies every
// browser origin; ["*"] is the only allow-all spelling.
```

`grep -rn -i "allow-all" --include="*.go"` says this is the only remaining contradiction in Go source; `Store.CORSOrigins` in `internal/settings/store.go` was updated correctly.

internal/api/router_test.go
```diff
@@ -591,19 +592,29 @@ func TestCORSMiddleware_EmptyOrigins_AllowAll(t *testing.T) {
 		{"empty list", func() []string { return []string{} }},
```
EricAndrechek, line 592: Nice coverage here — three empty spellings × GET/OPTIONS is the right matrix, and asserting 204 on `OPTIONS` is what actually pins the short-circuit.

One leftover a few hundred lines up: the `TestCORSMiddleware_NoCredentialsHeader` table at `internal/api/router_test.go:251` still has

```go
{"empty-allowlist-is-wildcard", nil, "https://anything.example.com"},
```

The case still passes — it only asserts that `Access-Control-Allow-Credentials` is absent, which holds on the deny path too — so it's purely the name that's now wrong, asserting the belief this test disproves. Worth keeping the case (it pins that the deny path doesn't emit credentials either) and renaming it to something like `empty-allowlist-denies` or `nil-allowlist`.

taitelee: Renamed to `empty-allowlist-denies` and kept the case, since it's the one that tests the deny path never emitting Access-Control-Allow-Credentials.

internal/api/router.go
taitelee: Reworded.

EricAndrechek (approved):

taitelee:
@coderabbitai review

PR #550, opened 2026-09-02
feat(discovery): widen schema discovery with DDL, position, version

EricAndrechek:
## Summary

Additive metadata the upcoming native type layer needs, captured on the
same refresh as the columns so they can never describe different servers:

- Column gains DefaultExpression and Position, both scanned from the
  widened system.columns select. TableSchema.Columns was already ordered
  by position, so declaration order needed no new structure.
- TableSchema gains DDL from system.tables.create_table_query. It is
  json:"-": the schema endpoint marshals TableSchema straight to the
  client and an external-engine table (S3, MySQL, Kafka) renders its wiring
  in that statement unconditionally — endpoint, bucket or host, database,
  username, S3 access key id. (ClickHouse masks the password itself as
  `[HIDDEN]` from ~23.9, verified on 26.7.3, so what is withheld here is the
  topology rather than the secret.) A table listed in system.tables with no
  system.columns rows is skipped, never published column-less.
- SchemaRegistry gains ServerVersion(), from a SELECT version() probe
  next to the existing SELECT timezone().

Both new queries fail the refresh on error, matching timezone() and
system.columns: callers keep the prior cache and retry.

## Stacked PR

This is **part 1 of 7** in a stack that replaces #540. Each PR is based on the one above it, so review this PR's own diff against its base — GitHub shows only this layer's changes.

| # | Branch | Base | |
| - | ------ | ---- | - |
| 0 | `stack/0-classify-paths` | `main` |  |
| 1 | `stack/1-discovery` | `stack/0-classify-paths` | **→ this PR** |
| 2 | `stack/2-policy` | `stack/1-discovery` |  |
| 3 | `stack/3-content-type` | `stack/2-policy` |  |
| 4 | `stack/4-seams` | `stack/3-content-type` |  |
| 5 | `stack/5-positional-wire` | `stack/4-seams` |  |
| 6 | `stack/6-computed-columns` | `stack/5-positional-wire` |  |

Merge in order, top to bottom. Rebasing or squashing out of order will make the later PRs' diffs unreadable.

## Test plan

- [x] `make ci` green on this branch's exact tree (verify, unit, integration against live ClickHouse, e2e, all coverage gates)
- [x] The branch descends from its base and carries only this layer's change (plus any follow-up commits answering review)
- [x] `go.mod` / `go.sum` untouched; no new dependencies

## Review

Both pre-push reviewers gate the tip of the stack (#555), whose delta against `main` is the union of all seven branches. This branch carries a logged skip (`scripts/skip-pre-push-review.sh`) on that basis, recorded in `tmp/review-skips-*.log`.

https://claude.ai/code/session_018Epn88jTEw4ZkXrvTKzZXQ


EricAndrechek:
@coderabbitai review

Auto-review is off for PRs whose base isn't the default branch, and this one sits in a stack (base `stack/0-classify-paths`). Requesting a review explicitly.

Note for context: this PR's diff against its base is one layer of a seven-PR stack replacing #540. Reviewing it in isolation is the intent — the layers below it are already in its history.

EricAndrechek:
@coderabbitai review

internal/discovery/discovery.go
```diff
@@ -134,18 +167,67 @@ func (sr *SchemaRegistry) Refresh(ctx context.Context) error {
...
+// scan didn't return, and one created between the two queries — is skipped rather
+// than added: a TableSchema with no columns is not a schema, and the two queries
+// are not a snapshot.
+func (sr *SchemaRegistry) attachDDL(ctx context.Context, tables map[string]*TableSchema) error {
+	rows, err := sr.conn.Query(ctx,
+		`SELECT name, create_table_query
+		 FROM system.tables
+		 WHERE database = ?
+		   AND name NOT LIKE '.%'`,
+		sr.database(),
```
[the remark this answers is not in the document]
EricAndrechek: Good catch — this is a real race and it is mine (`attachDDL` is new in this PR, and it added the second read).

`database` is a live getter (`func() string`), and its own comment says a ClickHouse reconfigure is honored "by the next refresh" — which only holds if a refresh reads it once. Reading it per-query meant a reconfigure landing between the `system.columns` and `system.tables` scans would attach DDL from the new database to same-named schemas discovered from the old one. Silent, and it would survive into `/v1/ops/schema`.

Fixed: `Refresh` now snapshots `database := sr.database()` before the first query and threads it to both; `attachDDL` takes it as a parameter rather than re-reading.

Added the test you asked for — `TestRefresh_DatabaseSnapshottedForWholeRefresh`. `fakeConn` gained an `onQueryArgs` hook so it can observe each query's bound args, and the getter flips `old` → `new` after the first call. **Verified it fails without the fix**: reverting just the `attachDDL` argument gives

```
--- FAIL: TestRefresh_DatabaseSnapshottedForWholeRefresh
    both queries in one refresh must use the same database; a reconfigure applies to the next refresh
```

🤖 Claude Opus 5 via [Claude Code](https://claude.com/claude-code)

@coderabbitai

internal/discovery/discovery_test.go
```diff
@@ -77,9 +88,59 @@ func TestRefresh_PopulatesAndLookups(t *testing.T) {
...
+
+// TestRefresh_ServerVersionQueryFails: version() is probed on the same
+// connection as timezone(), so its failure fails the refresh rather than
+// publishing a registry with no idea what server produced it.
+func TestRefresh_ServerVersionQueryFails(t *testing.T) {
+	t.Parallel()
+	conn := &fakeConn{versionErr: errors.New("code: 497, not enough privileges")}
+	sr := NewSchemaRegistry(conn, func() string { return "test" }, func() time.Duration { return time.Hour }, discardLogger())
+	require.ErrorContains(t, sr.Refresh(context.Background()), "query server version")
+	assert.Empty(t, sr.ServerVersion())
```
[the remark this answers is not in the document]
EricAndrechek: Accepted — and the sharpest part is that `TestRefresh_TablesQueryFails`' own comment already claimed "keeps the prior cache instead of publishing DDL-less schemas", while the test only ever asserted `Nil` after a single failed refresh. That assertion is satisfied by the exact regression it was meant to prevent: a failure that wipes a good registry.

Both now seed a successful refresh first:

- `TestRefresh_ServerVersionQueryFails` — refreshes with `version: "25.3.2.2"` and a `clicks` schema, asserts both are live, *then* sets `versionErr` and refreshes again, asserting the version **and** the schemas survive.
- `TestRefresh_TablesQueryFails` — refreshes with a table carrying DDL, asserts it landed, then sets `tablesErr` and asserts the prior schema *and its DDL* are still there.

That makes them tests of the documented behaviour ("callers keep the prior cache and retry", which the CHANGELOG entry for this PR also promises) rather than of emptiness.

🤖 Claude Opus 5 via [Claude Code](https://claude.com/claude-code)

@coderabbitai

EricAndrechek:
@coderabbitai review

The two findings from your last pass are fixed and pushed (`f41f0d7c`): the database is now snapshotted once per refresh, and both refresh-failure tests seed a successful refresh before asserting the prior cache survives. Requesting a re-review so this layer can go to human review — it is the first PR in the stack after #549, which is already approved.

taitelee:
@coderabbitai review

internal/api/boot_chain_test.go
```diff
@@ -31,16 +32,23 @@ type errsThenSuccessConn struct {
 	calls atomic.Int32
 }
 
-func (c *errsThenSuccessConn) Query(_ context.Context, _ string, _ ...any) (driver.Rows, error) {
+// Query serves both of Refresh's result sets. Only the system.columns call
+// advances the error sequence and the counter, so `calls` still counts refresh
+// ATTEMPTS rather than statements.
+func (c *errsThenSuccessConn) Query(_ context.Context, q string, _ ...any) (driver.Rows, error) {
+	if strings.Contains(q, "system.tables") {
+		return &chainEmptyRows{}, nil
```
taitelee, line 46: This check here `if strings.Contains(q, "system.tables")`. I get why this is here because a `Refresh` now issues a second `Query` for `system.tables`, and without this check that call would bump calls (`c.calls.Add(1)`) and I think somewhere there will be a scripted error and the attempt count would be wrong. But on a greater scope, anything that isn't `system.tables` is treated as a refresh attempt, so the next query someone adds to `Refresh` should also fail no? I guess no one would know unless they do add and run a specific assertion that fails. But I wouldn't want to keep building the boot chain test like this. Maybe we could change this so that anything other than `system.columns` will just return `chainEmptyRows` and make this specifically a query for `system.columns` so that one attempt is one columns scan.

internal/api/boot_chain_test.go
taitelee: I think with the `system.tables` Refresh there are now 5 Query calls. You can either change it to say 5 Query calls or say there are four refresh attempts or four `system.columns` queries. This is related to my previous comment above in this file.

taitelee (dismissed):
Looks great overall and straightforward. I can start to see the scaffolding for Chtypes.

internal/api/boot_chain_test.go
EricAndrechek: You're right, and the framing is the part worth keeping: the count was a symptom, the deny-list was the defect. Fixed in `9681da1c`.

Written as "skip `system.tables`, count everything else", a query added to `Refresh` later doesn't just inflate the count — it consumes `c.errs[n-1]`, an error scripted for a columns scan, and shifts the whole retry sequence. The test would go wrong *quietly* rather than fail, which is the bad direction for a fake that exists to script failures.

Inverted to an allow-list on `system.columns`, exactly as you suggested, so one attempt is one columns scan by construction and anything this fake doesn't recognize is inert. The comment now says the match is an allow-list on purpose and why, so the next person doesn't "simplify" it back.

internal/api/boot_chain_test.go
EricAndrechek: Confirmed, and 5 is exactly right — I instrumented the fake rather than counting by eye:

```
PROBE total Query=5   system.columns=4   system.tables=1
```

`Refresh` returns as soon as the `system.columns` scan fails, so the three failed attempts never reach `system.tables`; only the successful fourth issues both. That's why the counter reads 4 while the real call count is 5.

Went with your "four refresh attempts" wording in `9681da1c` — it names the thing the assertion actually protects (the retry loop's off-by-one) instead of a statement count that will drift again the next time `Refresh` grows a query. The comment now also states the 5-vs-4 reason, and the failure message reads `expected 4 refresh attempts (system.columns scans)`.

taitelee (approved):

Issue #526, opened 2026-08-27
chore(pm-triage): rubric.md still anchors P0-P3 on the pre-launch flip, which shipped


EricAndrechek:
**Area:** infra · agent tooling — gap · found via the pm-triage routine's own runs

**Expected:** `.claude/skills/pm-triage/references/rubric.md` carries the priority anchor the daily triage routine rates every issue against, and instructs "Re-anchor when the milestone moves."

**Actual:** the file still anchors on a milestone that shipped eight days ago:

> **Anchor:** the imminent public launch — `v0.1.0-alpha.1` cut **and** the repo public-flip, which happen together (tracked in **#149**).
> **P0 — blocks the flip.** Must be fixed before the repo goes public / the tag is cut.

`v0.1.0` was tagged 2026-08-19, the repo is public, and **#149 is closed**. Nothing can "block the flip" any more, so the written P0 definition selects the empty set.

**Impact:** the routine has been rating against a post-launch anchor that lives only in a comment on #501 and in the routine's local `pending.md` —

> P0 — live in a shipped, public release and severe enough to cut a patch for. P1 — wanted in the next release. P2 — blocks broader adoption. P3 — backlog.

— i.e. not in the file the next run reads. That working anchor has already priced real issues (#514/#515/#516/#517/#518, and #523/#524 today), and both current P0s (#460, #474) are shipped-release defects, a category the written rubric has no slot for at all. Two consecutive runs recorded that nothing had been mis-rated yet and called that luck; the ratings above are what the luck was covering.

Scope: re-anchor `rubric.md`'s P0–P3 definitions and examples on post-`v0.1.0` reality — one file, through the normal gate.

Related: #501 (where the working anchor is recorded), #149 (closed), #522 (last change to a pm-triage reference file)

---
_From the pm-triage `all` routine, escalated across seven runs; validated by code-read against `7b3c25e` on 2026-08-27._

PR #569, opened 2026-09-08
feat(config): refuse unbound WH_* env and unusable data_dir at boot

### LINKED ISSUE #530: config: loud validation for boot config (dry-run validate, required keys, unbound WH_* env)
## Context

`wavehouse validate [dir]` checks the settings directory loudly without starting the server. Boot config (`config.yaml` + `WH_*` env) had no equivalent.

#508 did the cheap half: `config.Load` re-reads the YAML against the `Config` struct tags and refuses to boot naming every undeclared key, so a `dedupe:`, `mq:`, or `policy:` block left over from before the settings-directory move can't be read, ignored, and believed (`internal/config/strict.go`).

Raised by @EricAndrechek in review of #508: it should be possible to have validation "working loudly on the non-hot-reloadable startup ops config too".

## Decision

**Boot itself is the validator for the non-hot-reloadable half. There is no dry-run subcommand.** Boot config only takes effect through a restart, so the restart is where it is checked; a refused boot with the offending key, variable, or path named in the error is the loud signal. `wavehouse validate` stays the settings-directory dry run only — that half is edited under a running server, which is why it needs one.

The earlier draft of this issue proposed a `validate`-style dry run for boot config with the same 0/1/2 exit contract and a two-command docs story. Dropped: it would duplicate the loader for no new signal.

## Remaining work

- **Unbound `WH_*` environment variables refuse boot.** `config.Load` errors, naming every `WH_*` name in the environment that no `Config` field binds and the binary doesn't otherwise read (`WH_CONFIG`, `WH_LOG_LEVEL`) — the environment half of the strict YAML loader. Only the `WH_` prefix is checked, since the environment always carries unrelated names. This is a refusal rather than a warning: a deployment still exporting a pre-#508 variable (`WH_DEDUPE_ENABLED`, `WH_CH_ADDR`, …) fails on upgrade instead of silently ignoring it.
- **`data_dir` is probed before anything dials out.** Right after `Load`, boot checks that `data_dir` is a writable directory, or absent under a writable parent so it can be created, so a root-owned bind mount refuses boot in the first second of the log with the UID-65532 remediation attached, rather than after ClickHouse discovery.
- **Docs** state that boot is the validator (`configuration.mdx` loading order, `settings-directory.mdx` cross-reference, `architecture.md`, `AGENTS.md`).

Not in scope: an auth-secret requirement. No secret plus no `jwks_url` is a documented, valid public posture, and the example `config.yaml` ships the placeholder secret, so refusing either would break `make dev`. The existing boot warnings for that posture stay.

## Related

#514 covered strict decoding for the policy file and is closed by #508 (policies.json goes through `settings.Validate`). This issue is boot config only.

taitelee:
## Summary

Boot is now the validator for the non-hot-reloadable half of configuration; there is no dry-run subcommand (decided on #530). `config.Load` refuses to start on a `WH_*` environment variable that no `Config` field binds and the binary doesn't otherwise read, naming every offender — the environment half of the strict YAML loader, so `WH_DEDUPE_ENABLED=true` left over from before the settings-directory move can no longer be set, ignored, and believed. Only the `WH_` prefix is checked, since the environment always carries unrelated names.

Right after `Load`, before the settings directory or ClickHouse are touched, `config.CheckDataDir` probes that `data_dir` is a writable directory (or absent under a writable parent, so boot can create it), so a root-owned bind mount refuses boot in the first second of the log with the UID-65532 remediation attached. Docs state that boot is the validator and that `wavehouse validate` stays the settings-directory dry run only.

## Related Issues

Closes #530


internal/config/check.go
```diff
@@ -0,0 +1,113 @@
...
+	switch {
+	case errors.Is(err, fs.ErrNotExist):
+		// Boot creates it; check the ancestor below.
+	case err != nil:
+		return fmt.Errorf("data_dir %s: %w", dir, err)
+	case !info.IsDir():
+		return fmt.Errorf("data_dir %s is not a directory", dir)
+	}
+
+	target := dir
```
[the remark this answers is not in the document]
taitelee: Confirmed reachable: `WH_DATA_DIR=` reaches Load as an empty string and the probe fell back to the working directory. `CheckDataDir` now refuses a blank value outright, with a test. It lives in the probe rather than `Config.Validate` because the existing `Validate` tests build literals without a data directory and the probe is what boot runs right after Load either way. 

@coderabbitai

taitelee:
@coderabbitai review

taitelee:
@coderabbitai review

docs/src/content/docs/deployment.md
```diff
@@ -206,12 +206,13 @@ volumes:
   - /srv/wavehouse:/app/data
 ```
 
-Bind mounts do **not** copy-up — Docker exposes the host directory as-is, and the image's pre-created dir is masked entirely. If `/srv/wavehouse` is owned by `root:root` on the host (the default for a freshly `mkdir`'d directory), the binary fails at startup with a permission error from NATS:
+Bind mounts do **not** copy-up — Docker exposes the host directory as-is, and the image's pre-created dir is masked entirely. If `/srv/wavehouse` is owned by `root:root` on the host (the default for a freshly `mkdir`'d directory), the binary refuses to start before it touches the settings directory or ClickHouse — `data_dir` is probed for writability right after the config loads:
```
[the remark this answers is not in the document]
taitelee: Agreed, the probe checks write access as the running user. The shared hint now says the directory must be writable by UID 65532 with chown as the usual fix, and deployment.md states the same condition and notes that a root-owned directory with permissive mode or an ACL passes.

docs/src/content/docs/configuration.mdx
```diff
@@ -16,7 +16,10 @@ WaveHouse is configured via a YAML file with environment variable overrides. All
...
+4. Both sources are **strict**. A YAML key this page doesn't list — a typo, or a tunable that has moved to the settings directory (`dlq.enabled`, `clickhouse.addr`, `stream.*`, a leftover `policy:` or `pipes:` block, …) — refuses to boot and names every offending key, so nothing is read, ignored, and believed. A `WH_*` environment variable that binds to no key on this page (`WH_DEDUPE_ENABLED`, `WH_CH_ADDR`, a misspelling) refuses to boot the same way; only the `WH_` prefix is checked, since the environment always carries names that aren't WaveHouse's. One outside source does share the prefix. Kubernetes injects `{SERVICE}_SERVICE_HOST`, `{SERVICE}_PORT`, and similar link variables into every pod, for each Service that existed before the pod started. A Service named `wh` or `wh-*` therefore produces `WH_SERVICE_HOST`, `WH_PORT`, and more, and the pod refuses to boot on its next restart. Set `enableServiceLinks: false` on the pod spec, or name the Service something else. The error says so.
```
EricAndrechek: This rule forbids `WH_CONFIG`, which line 24 tells the reader to export. It binds to no key on this page or anywhere else. `WH_LOG_LEVEL` is the same — Env Var column, no YAML key.

Both are exempt in `processEnv`. Say so here; CHANGELOG and architecture.md already do.

internal/config/check.go
```diff
@@ -0,0 +1,133 @@
...
+func CheckDataDir(dir string) error {
+	if strings.TrimSpace(dir) == "" {
+		return errors.New("data_dir (WH_DATA_DIR) is required: an empty value would scatter NATS and Pebble state under the working directory")
+	}
+	info, err := os.Stat(dir)
+	switch {
+	case errors.Is(err, fs.ErrNotExist):
+		// Boot creates it; check the ancestor below.
+	case err != nil:
+		return fmt.Errorf("data_dir %s: %w", dir, err)
```
EricAndrechek, line 102: Permission denial lands on this arm, not the `CreateTemp` path below, so it returns without `permissionHint` — the one case the hint exists for. `/srv/data` at `root:root 0700` with `WH_DATA_DIR=/srv/data/wavehouse`: `os.Stat` returns EACCES, not `ErrNotExist`, the walk never runs, and the operator gets bare `permission denied`.

Add the `fs.ErrPermission` branch here. `check_test.go` has no unsearchable-parent case, which is why this arm went unnoticed.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
...
...-link variables for a Service named `wh` or `wh-*` — is named in the error with the `enableServiceLinks: false` remediation. Right after `Load`, before ClickHouse or the settings directory are touched, `config.CheckDataDir` probes that `data_dir` is a writable directory (or absent under a writable parent, so boot can create it) by creating and removing one temp file, so a `data_dir` the process cannot write to refuses boot before schema discovery rather than after it; a permission denial carries the UID-65532 remediation (a bind mount owned by root is the typical cause), and that hint string is now shared with `LogStorageInitError`. An empty or blank `data_dir` — reachable through `WH_DATA_DIR=` — is refused by the probe outright, since the ancestor walk would otherwise fall back to the working directory and NATS and Pebble state would land under it. `EnvConfig` and `EnvLogLevel` join `EnvSettingsDir` as the exported names for the process-level variables. Boot is the validator for the non-hot-reloadable half — there is no dry-run subcommand, by decision on #530: boot config only takes effect through a restart, so the restart is where it is checked, and the docs say so. Closes #530.
```
EricAndrechek: Wrong section. This stops existing deployments from booting, so it's Changed + BREAKING, not Added — same as policy format v2, ingest `Content-Type`, and the envelope column this release. Filed here, anyone scanning for breaking changes misses it.

Also doesn't mention the not-a-directory or dangling-symlink refusals.

docs/src/content/docs/configuration.mdx
```diff
@@ -16,7 +16,10 @@ WaveHouse is configured via a YAML file with environment variable overrides. All
...
+5. Before anything dials out, `data_dir` is probed: it must be a writable directory, or absent with a writable parent so it can be created. A `data_dir` the process cannot write to refuses to boot before ClickHouse discovery rather than after it, and a permission denial carries the UID-65532 remediation, since a bind mount owned by root is the typical cause. An empty `data_dir` is refused outright.
```
EricAndrechek: Incomplete. `CheckDataDir` also refuses a path that exists and isn't a directory, and a dangling symlink at `data_dir` or any component above it. The symlink case is new in this PR and is exactly the failed-mount scenario — as written this bullet says it passes.

List all of them. architecture.md:111 needs the same.

docs/src/content/docs/architecture.md
```diff
@@ -107,7 +107,8 @@ The SSE fan-out, factored out of `api/` so the delivery hot path ([#294](https:/
...
+- **config.go** — Loads *boot* configuration from a YAML file with environment variable overrides (using [cleanenv](https://github.com/ilyakaznacheev/cleanenv)); every key has a `WH_`-prefixed env var. Boot config is only what can't change under a running process — resource sizing, listeners, observability exporters, the settings-directory path, and the secrets (`clickhouse.password`, `auth.jwt_secret`, `auth.operator_key`). Everything tenant-tunable lives in the settings directory (`settings/`). Both sources are strict: `rejectUnknownKeys` refuses to boot naming every YAML key the struct doesn't declare, and `rejectUnboundEnv` does the same for every `WH_*` environment variable no field binds, so a tunable that moved to the settings directory can't be read, ignored, and believed. Boot is the validator for this half — there is no dry-run command. See [Configuration Reference](/configuration).
```
EricAndrechek: repo root `config.yaml`'s header comment contradicts this claim and still describes strictness as YAML-only

docs/src/content/docs/deployment.md
```diff
@@ -206,12 +206,13 @@ volumes:
...
+Bind mounts do **not** copy-up — Docker exposes the host directory as-is, and the image's pre-created dir is masked entirely. The condition boot enforces is that UID 65532 can **write** to the directory — a freshly `mkdir`'d `root:root` directory at the default mode cannot be, which is the common case, though a root-owned directory with permissive mode bits or an ACL passes. If `/srv/wavehouse` is not writable, the binary refuses to start before it touches the settings directory or ClickHouse — `data_dir` is probed for writability right after the config loads:
 
 ```text wrap=false
-ERROR  mq init failed  error="..."  path=/app/data/nats
-       hint="if running in a container with a host bind mount, the host
-       directory must be owned by UID 65532..."
+ERROR  check data_dir  error="data_dir /app/data is not writable; if running in a
+       container with a host bind mount, the host directory must be writable by
+       UID 65532 (the `nonroot` user in the distroless image); the usual fix is
+       `sudo chown -R 65532:65532 /your/host/path`. ...: permission denied"
```
EricAndrechek, line 215: §Upgrading below is the upgrade runbook and doesn't mention auditing the environment. This release dropped ~20 boot env vars and this PR turns every leftover into a boot failure, so following those steps as written hands you a pod that won't start. Needs an audit step.

internal/config/check.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+// envPrefix is what every WaveHouse environment variable starts with, and the
+// only filter that makes an unbound-name check meaningful: the environment
+// always carries names that aren't ours.
+const envPrefix = "WH_"
+
+// processEnv lists the WH_* names the binary reads outside the Config struct
+// (cmd/wavehouse), so UnboundEnv doesn't flag them.
+var processEnv = []string{EnvConfig, EnvLogLevel}
```
EricAndrechek, line 21: Missing `WH_SKIP_DIAGRAM_PNG` — `docs/src/integrations/diagram-png.mjs` reads it. Anyone with it exported can't run `make dev` or `make test-e2e`, and the error blames a typo or the settings-directory move.

Add it here, or, ideally, rename the docs knob out of `WH_`.

internal/config/check.go
```diff
@@ -0,0 +1,133 @@
...
+	return fmt.Errorf("unbound environment variable(s): %s — a typo, or a key that moved to the settings directory (%s); unset it, or rename it to a key the Config struct declares. On Kubernetes, a Service named wh or wh-* injects WH_SERVICE_HOST, WH_PORT, … into every pod started after it: set enableServiceLinks: false on the pod spec, or rename the Service", strings.Join(unbound, ", "), EnvSettingsDir)
+}
+
+// UnboundEnv returns, sorted, every WH_* name in environ (os.Environ() form,
+// "KEY=value") that no Config field's env tag binds and the binary doesn't
+// read otherwise. Such a name is almost always a typo or a key that moved to
+// the settings directory — `WH_DEDUPE_ENABLED=true` left in a compose file
+// would otherwise be set, ignored, and believed. Only the WH_ prefix is
+// checked, since the environment is shared with whatever launched the process.
+func UnboundEnv(environ []string) []string {
```
EricAndrechek: Why is this exported (capitalized)? No caller outside the package — tests are in-package, and the dry-run subcommand that might have used it was ruled out on #530. Unexport it, or say what's going to call it. `rejectUnknownKeys` is unexported for the YAML equivalent.

docs/src/content/docs/deployment.md
```diff
@@ -186,7 +186,7 @@ On a first-ever run this is expected. On every subsequent run it should be silen
 
 ### Distroless Permission Traps (named volume vs bind mount)
 
-WaveHouse images run as the distroless `nonroot` user (UID 65532). Bind mounts and named volumes interact with this differently, and the distroless image has no shell to `chown` things at runtime — so getting the host side wrong produces a hard-to-read permission error from NATS or Pebble at startup.
+WaveHouse images run as the distroless `nonroot` user (UID 65532). Bind mounts and named volumes interact with this differently, and the distroless image has no shell to `chown` things at runtime — so getting the host side wrong refuses boot in the first lines of the log with a named `data_dir` error and the `chown` remediation attached.
```
EricAndrechek: Wrong for half this section. The `/app/settings` mount fails with a settings-validation error from `settings.Open`, not a `data_dir` one. Scope the sentence to `/app/data`.

docs/src/content/docs/deployment.md
```diff
@@ -206,12 +206,13 @@ volumes:
   - /srv/wavehouse:/app/data
 ```
 
-Bind mounts do **not** copy-up — Docker exposes the host directory as-is, and the image's pre-created dir is masked entirely. If `/srv/wavehouse` is owned by `root:root` on the host (the default for a freshly `mkdir`'d directory), the binary fails at startup with a permission error from NATS:
+Bind mounts do **not** copy-up — Docker exposes the host directory as-is, and the image's pre-created dir is masked entirely. The condition boot enforces is that UID 65532 can **write** to the directory — a freshly `mkdir`'d `root:root` directory at the default mode cannot be, which is the common case, though a root-owned directory with permissive mode bits or an ACL passes. If `/srv/wavehouse` is not writable, the binary refuses to start before it touches the settings directory or ClickHouse — `data_dir` is probed for writability right after the config loads:
```
EricAndrechek, line 209: `deployments/Dockerfile` and `deployments/Dockerfile.goreleaser` still say the host directory must be *owned* by UID 65532. This line is the correct version — writable. Update both; `Dockerfile.goreleaser` asks for them to be kept in step.

docs/src/content/docs/configuration.mdx
EricAndrechek: Service link variables are only injected for Services in the pod's own namespace. "into every pod" reads as though a `wh` Service anywhere in the cluster breaks every WaveHouse pod, which sends people hunting in the wrong namespaces.

docs/src/content/docs/deployment.md
EricAndrechek: Separate point for the Kubernetes section further down: it says nothing about service links, and that's the likeliest way an operator meets the new env check. A Service named `wh` or `wh-*` injects `WH_SERVICE_HOST`, `WH_PORT` and friends, and the pod stops booting on its next restart.

The error names the remediation so they do recover, but the only copy of the warning is a sentence buried in configuration.mdx's loading-order list, and AGENTS.md routes deployment questions here.

EricAndrechek (changes requested):
Good stuff. My only real find is that the permission-denied path returns without the permission hint which is what it exists for and there's no test that would have caught that I don't think.

A couple tiny other nitpick things I left in `check.go`, and then everything else is just stale comment/doc things, as per usual.

docs/src/content/docs/configuration.mdx
taitelee: Fixed to say both are exempt.

internal/config/check.go
taitelee: Added the `fs.ErrPermission` branch on the stat now that attaches the hint. Also added test for it.

CHANGELOG.md
taitelee: Moved to breaking. Also now names the not-a-directory and dangling-symlink refusals, says outright that a deployment still exporting a moved variable stops booting, and points at the audit step in the upgrade runbook.

docs/src/content/docs/configuration.mdx
taitelee: Listing all of them and in `architecture.md` as well.

docs/src/content/docs/architecture.md
taitelee: Header now says the environment is strict too, whether the offender is a YAML key or a `WH_*` variable no key binds.

docs/src/content/docs/deployment.md
taitelee: Added a third pre-drain audit: every `WH_*` the binary no longer reads refuses boot, with the moved key groups named, a pointer to the `config.json` keys, and the service-links case.

internal/config/check.go
taitelee: Renamed the docs knob to `DOCS_SKIP_DIAGRAM_PNG` rather than exempting it. It had one reader in diagram-`png.mjs`, and the prefix is meant to be the server's.

internal/config/check.go
taitelee: No caller. Unexported to `unboundEnv`, matching `rejectUnknownKeys`; the doc references followed.

docs/src/content/docs/deployment.md
taitelee: Scoped to `/app/data`.

docs/src/content/docs/deployment.md
taitelee: Both Dockerfiles now say writable by UID 65532, not owned by it.

taitelee:
@coderabbitai review

EricAndrechek:
@coderabbitai review

docs/src/content/docs/configuration.mdx
EricAndrechek: Reopening. The exemption is named now, but the reason given is wrong.

`WH_LOG_LEVEL` is read at `main.go:208`, ~50 lines *after* `config.Load` at `:157` — the bootstrap logger at `:147` is built with nil options and only rebuilt at the configured level once config is loaded. So neither variable is exempt "because the process reads them before the file".

They're exempt because neither is a `Config` field, which is what `config.go:35-38` already says correctly. Fix the clause.

docs/src/content/docs/configuration.mdx
```diff
@@ -16,7 +16,10 @@ WaveHouse is configured via a YAML file with environment variable overrides. All
...
+4. Both sources are **strict**. A YAML key this page doesn't list — a typo, or a tunable that has moved to the settings directory (`dlq.enabled`, `clickhouse.addr`, `stream.*`, a leftover `policy:` or `pipes:` block, …) — refuses to boot and names every offending key, so nothing is read, ignored, and believed. A `WH_*` environment variable that binds to no key on this page (`WH_DEDUPE_ENABLED`, `WH_CH_ADDR`, a misspelling) refuses to boot the same way. Two variables have no YAML key and are exempt, because the process reads them before the file: `WH_CONFIG` (below), which locates the file, and `WH_LOG_LEVEL`. Only the `WH_` prefix is checked, since the environment always carries names that aren't WaveHouse's. One outside source does share the prefix. Kubernetes injects `{SERVICE}_SERVICE_HOST`, `{SERVICE}_PORT`, and similar link variables into every pod, for each Service that existed before the pod started. A Service named `wh` or `wh-*` therefore produces `WH_SERVICE_HOST`, `WH_PORT`, and more, and the pod refuses to boot on its next restart. Set `enableServiceLinks: false` on the pod spec, or name the Service something else. The error says so.
```
EricAndrechek: Wrong twice. Link variables are injected only into pods in the Service's **own namespace**, and Services with no ClusterIP are skipped entirely, so a headless `wh` Service injects nothing at all.

As written an operator hunts for a `wh-*` Service cluster-wide, or clears a WaveHouse pod because the Service sits in another namespace. Same imprecision in the error string at `check.go:30`, which this bullet's "The error says so" points at — the two should end up matching.

docs/src/content/docs/deployment.md
```diff
@@ -339,10 +340,11 @@ This affects the streaming surface too, and more quietly. SSE gap-fill (`?since=
...
+- **Every `WH_*` variable the binary no longer reads refuses boot.** The tunables that moved to the settings directory this release — the ClickHouse wiring other than the password, `dedupe.*`, `dlq.*`, `stream.*`, `mq.*`, `query.*`, `schema.*`, `cors.*`, `auth.role_claim`, and the old policy-file and pipes-directory paths — used to be `WH_*` variables. The old binary ignored a leftover; the new one names every variable it does not bind and exits before it opens the queue, so a pod spec or compose file that still carries one comes back from the upgrade as a container that will not start. Diff the environment against the [Configuration Reference](/configuration) first, and move each leftover to its [`config.json` key](/settings-directory#configjson-keys) or drop it. A Kubernetes Service named `wh` or `wh-*` counts too: it injects link variables under the `WH_` prefix (`WH_SERVICE_HOST` and `WH_PORT` for `wh`, `WH_FOO_SERVICE_HOST` and `WH_FOO_PORT` for `wh-foo`), so set `enableServiceLinks: false` on the pod spec.
```
EricAndrechek: This list reads as exhaustive and misses two that `CHANGELOG.md:15` names as removed this release: `WH_AUTH_JWKS_URL` and `WH_CACHE_TIMESTAMP_BUCKET_SECONDS`.

JWKS is the one that matters. It's the standard production auth posture, so the variable most likely to be exported by a real 0.1.0 deployment is the one the audit doesn't name. The cache omission misleads the other way: no `cache.*` appears here while `WH_CACHE_L1_MAX_COST` is still bound at `config.go:132`, so a reader concludes theirs survived the move. Name it under its old name, since that's what they're diffing against.

EricAndrechek: "move each leftover to its `config.json` key" can't be followed for two items this same sentence lists. `WH_POLICY_FILE_PATH` and `WH_PIPES_DIR` have no `config.json` key — they became `policies.json` and `pipes.json`, separate files in the settings directory. A reader scans the key table, finds nothing, and stalls.

docs/src/content/docs/deployment.md
```diff
@@ -339,10 +340,11 @@ This affects the streaming surface too, and more quietly. SSE gap-fill (`?since=
 
 On the worker side the outcome depends on the DLQ. **With the DLQ enabled for the table**, the message is parked on `dlq.{table}` with `X-DLQ-*` headers and is recoverable by hand — but re-ingest each parked envelope's inner `data` object as a fresh `POST /v1/ingest`; republishing the envelope as-is onto `ingest.{table}` fails the same `format` check and simply re-parks it. **With the DLQ switched off for the table, it is permanently lost**: acked and dropped with an `ERROR` log and a `wavehouse_ingest_poison_total` increment carrying `disposition="dropped"`, unrecoverable from either the ingest stream or the DLQ, because a message that can never insert must not redeliver forever. Draining first is cheaper than a manual replay, and it is the only option at all where the DLQ is off.
 
-Two audits belong **before** the drain, because neither announces itself afterwards:
+Three audits belong **before** the drain, because none of them announces itself afterwards:
```
EricAndrechek: "none of them announces itself afterwards" is false for the third bullet you just added. The first two are silent-corruption cases. An unbound `WH_*` is the loudest failure the system has — the bullet itself says boot refuses and names every offender.

internal/config/check.go
```diff
@@ -0,0 +1,139 @@
...
+var processEnv = []string{EnvConfig, EnvLogLevel}
+
+// rejectUnboundEnv is the environment half of rejectUnknownKeys: an error
+// naming every WH_* variable in environ that nothing reads.
+func rejectUnboundEnv(environ []string) error {
+	unbound := unboundEnv(environ)
+	if len(unbound) == 0 {
+		return nil
+	}
+	return fmt.Errorf("unbound environment variable(s): %s — a typo, or a key that moved to the settings directory (%s); unset it, or rename it to a key the Config struct declares. On Kubernetes, a Service named wh or wh-* injects WH_SERVICE_HOST, WH_PORT, … into every pod started after it: set enableServiceLinks: false on the pod spec, or rename the Service", strings.Join(unbound, ", "), EnvSettingsDir)
```
EricAndrechek: This string is wrong twice. A Service named `wh-foo` produces `WH_FOO_SERVICE_HOST`, never `WH_SERVICE_HOST` — kubelet uppercases the Service name and maps `-` to `_`. And link variables reach only pods in the Service's namespace, not "every pod".

`deployment.md:347` got the correct naming in 26d90767; this and `configuration.mdx:19` didn't get the same edit. This is the copy that matters most: an operator reads it, greps the pod env for `WH_SERVICE_HOST`, finds nothing, and rules out service links — while `WH_FOO_SERVICE_HOST` is sitting in the offender list printed right above it.

docs/src/content/docs/deployment.md
EricAndrechek: `enableServiceLinks: false` exists only here, in the upgrade section. The hazard isn't upgrade-specific: any new deployment with a Service named `wh` or `wh-*` gets the same refusal, and someone installing fresh never reads §Upgrading.

It belongs on the PodSpec snippet at §Kubernetes, with this bullet cross-referencing it rather than carrying the only copy.

docs/src/content/docs/architecture.md
EricAndrechek: This section maps functions to files and this pair is mis-mapped. `rejectUnboundEnv` is in `check.go:24`, not config.go, so it reads as living apart from `unboundEnv` on the next line. `rejectUnknownKeys` is in `strict.go:19` — pre-existing, worth the same pass. `strict.go` and `persistence.go` have no bullet at all.

docs/src/integrations/diagram-png.mjs
```diff
@@ -109,8 +109,8 @@ export function diagramPng(options = {}) {
     name: "wh-diagram-png",
     hooks: {
       "astro:build:done": async ({ dir, pages, logger }) => {
-        if (process.env.WH_SKIP_DIAGRAM_PNG === "1") {
-          logger.info("skipped (WH_SKIP_DIAGRAM_PNG=1)");
+        if (process.env.DOCS_SKIP_DIAGRAM_PNG === "1") {
```
EricAndrechek: Rename is clean in code, but `DOCS_SKIP_DIAGRAM_PNG` is now documented nowhere. `development.md:32` already documents the sibling knob `DOCS_WATCH_STRICT=1` next to this integration; it belongs there.

Same file, §"Using an .env File" at :255: a stale `WH_*` in a contributor's `.env` is a boot refusal now rather than a silent no-op. One clause covers it.

internal/config/check.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+// envPrefix is what every WaveHouse environment variable starts with, and the
+// only filter that makes an unbound-name check meaningful: the environment
+// always carries names that aren't ours.
+const envPrefix = "WH_"
+
+// processEnv lists the WH_* names the binary reads outside the Config struct
+// (cmd/wavehouse), so unboundEnv doesn't flag them.
+var processEnv = []string{EnvConfig, EnvLogLevel}
```
EricAndrechek: Not this PR, but the argument that renamed the diagram knob applies to nine more names. `.claude/skills/pm-triage/scripts/board.sh` and `state.sh` read `WH_REPO`, `WH_OWNER`, `WH_PROJECT_ID`, `WH_PROJECT_NUM`, `WH_PRIORITY_FIELD`, `WH_STATUS_FIELD`, `WH_PM_STATE_{BRANCH,AUTHOR,EMAIL}`, and `SKILL.md:93` documents them as the supported override mechanism.

Export one to point the routine at a different board and every `make dev` in that shell dies at boot blaming the settings-directory move. Lower odds than the docs knob, so a follow-up issue is fine — but it shouldn't just sit.

docs/src/content/docs/configuration.mdx
taitelee: Fixed please take a look again.

EricAndrechek (dismissed):
Ok approving it cause the rest is all docs. I don't know enough about k8s to understand what/why its getting nit-picky at comments and docs so I had it try to help explain it to me, which I thought you may find helpful too. It also flagged our pm-triage skill which I left out as I already know I need to fix that skill up and _think_ I have an issue to track it's general fixes already. Here's what it said:

## Kubernetes service links

Worth spelling out, because we describe this in three places and no copy is
quite right.

kubelet injects a set of environment variables into every container for each
Service it can see. It's a Docker-links compatibility feature, on by default
through `enableServiceLinks: true` on the PodSpec. Three rules decide what a
pod actually receives:

- **Namespace.** Only Services in the pod's own namespace, plus the
  `kubernetes` API Service from `default`. A `wh` Service in some other
  namespace is harmless.
- **Timing.** Only Services that already existed when the pod started. A
  Service created afterwards injects nothing until the pod restarts.
- **ClusterIP.** Services with `clusterIP: None` are skipped, so a headless
  `wh` Service injects nothing at all.

Names are the Service name uppercased with `-` mapped to `_`, then a fixed set
per Service (port numbers here are illustrative — they follow the Service's
own ports):

    Service "wh"     → WH_SERVICE_HOST, WH_SERVICE_PORT, WH_PORT,
                       WH_PORT_8080_TCP, WH_PORT_8080_TCP_ADDR, …
    Service "wh-foo" → WH_FOO_SERVICE_HOST, WH_FOO_SERVICE_PORT, WH_FOO_PORT,
                       WH_FOO_PORT_8080_TCP, …

Every one of those starts with `WH_`, so every one trips the new check. That
part of the feature is working as intended. What's off is how we describe it:

- `configuration.mdx:19` says "into every pod". It's every pod *in the
  Service's namespace*, and never for a headless Service.
- `check.go:30` says a Service named `wh` **or `wh-*`** injects
  `WH_SERVICE_HOST, WH_PORT`. Only an exact `wh` produces those — `wh-foo`
  produces `WH_FOO_SERVICE_HOST` and `WH_FOO_PORT`. This is the copy that
  matters most: it's what an operator reads when the pod won't start, and it
  sends them grepping for a variable that isn't there while the real offender
  is already named in the list printed directly above it. No inline comment on
  it, but it wants the same edit as the doc.
- `deployment.md:347` is already correct as of 26d90767.

Two consequences worth putting in the docs, because they're the shapes this
will actually take in the field:

- **The failure is a restart, not a deploy.** The Service almost always
  pre-exists, so a cluster runs fine until something recycles the pod — a
  rollout, a node drain, an OOM kill — and then it doesn't come back. Bad thing
  to learn at 3am.
- **The likeliest instance is self-inflicted.** If WaveHouse's own Service is
  named `wh`, WaveHouse's own pods get `WH_SERVICE_HOST` and refuse to boot.
  Our deployment docs should probably say don't name it that.

I'm reasoning from kubelet's behaviour rather than from a cluster in front of
me, so before rewriting anything: `kubectl exec <pod> -- env | grep ^WH_` on a
pod in a namespace that has a `wh-*` Service. If that comes back empty, I've
got something wrong and I'd rather find out first.

## Upgrade audit

Two comments on the same bullet at `deployment.md:347`. It's presented as the
complete list of what to strip before upgrading, so gaps in it are the
difference between a clean upgrade and a pod that won't start — `WH_AUTH_JWKS_URL`
in particular, since JWKS is the standard production auth posture.

## pm-triage — follow-up, not this PR

Renaming the diagram knob was the right call, and the same argument covers nine
more names we own: `.claude/skills/pm-triage/scripts/board.sh` and `state.sh`
read `WH_REPO`, `WH_OWNER`, `WH_PROJECT_ID`, `WH_PROJECT_NUM`,
`WH_PRIORITY_FIELD`, `WH_STATUS_FIELD`, and `WH_PM_STATE_{BRANCH,AUTHOR,EMAIL}`,
and `SKILL.md:93` documents them as the supported override mechanism.

`export WH_PROJECT_ID=…` to point the routine at a different board and every
`make dev` in that shell then dies at boot blaming the settings-directory move.
Lower odds than the docs knob — it's an occasional local routine — so this is a
follow-up issue rather than something to hold #530 for. Happy to file it.

docs/src/content/docs/configuration.mdx
taitelee: Matched now.

docs/src/content/docs/deployment.md
taitelee: Dropped the list instead of completing it. The bullet now says diff against the Configuration Reference and anything not in its tables is unbound. Pre-1.0, I'd rather not maintain a catalogue of dead names in the docs; the changelog keeps that detail.

taitelee: Same as above. It no longer points at a `config.json` key, just says whatever the variable used to configure now lives in the settings directory or is gone.

docs/src/content/docs/architecture.md
taitelee: `config.go` now says Load runs both checks and names the files. `rejectUnboundEnv` sits on the `check.go` bullet, and `strict.go` and `persistence.go` each have a bullet now.

taitelee:
Added a short note in the k8s section of deployment.md not to name our own Service `wh`, since that's the case people will actually hit and it only surfaces on a restart. 

pm-triage is tracked in #582, since #526 only covers the rubric.

@coderabbitai review

internal/config/check.go
```diff
@@ -0,0 +1,139 @@
...
+
+// rejectUnboundEnv is the environment half of rejectUnknownKeys: an error
+// naming every WH_* variable in environ that nothing reads.
+func rejectUnboundEnv(environ []string) error {
+	unbound := unboundEnv(environ)
+	if len(unbound) == 0 {
+		return nil
+	}
+	return fmt.Errorf("unbound environment variable(s): %s — a typo, or a key that moved to the settings directory (%s); unset it, or rename it to a key the Config struct declares. On Kubernetes, a Service in the pod's namespace named wh or wh-* injects link variables under the WH_ prefix (WH_SERVICE_HOST and WH_PORT for wh, WH_FOO_SERVICE_HOST and WH_FOO_PORT for wh-foo, …) into every pod started after it: set enableServiceLinks: false on the pod spec, or rename the Service", strings.Join(unbound, ", "), EnvSettingsDir)
```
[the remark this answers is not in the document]
taitelee: The error now says to rename the variable to one the binary reads, either a Config struct key or `WH_CONFIG` / `WH_LOG_LEVEL`, using the exported constants so the names cannot drift.

taitelee:
@coderabbitai review

EricAndrechek:
@coderabbitai review

EricAndrechek (approved):

PR #585, opened 2026-09-15
refactor(app): extract the process wiring from main.go into internal/app

taitelee:
## Summary

Pulls the ~550 lines of hand wiring and settings-reload hooks out of `cmd/wavehouse/main.go` into a new `internal/app` package. `app.New` builds every component from the boot config and the settings directory, `Run` drives the long-lived ones (ingest worker, sweeper, hub bridge, keepalive wheel, schema refresh, SIGHUP + directory watcher, API server, Prometheus sidecar) under one erxt is cancelled or a component fails, and `Close` releases what`New` opened in reverse order. Each component is wired in one place, with the settings store handed to its wiring function whole, so the per-tenant store registry (#583, story 2) lands there rather than in `main`.

- `main.go` shrinks to argv dispatch, logger, `config.Load`, `CheckDataDir`, `app.New`, `app.Run`; `run(ctx)` takes the signal context and is unit-tested, so
the `.testcoverage.yml` exclude for it is removed.
- `tests/integration` boots through the same `app.New` against its testcontainer instead of a hand-built handler subset that had drifted from the binary.
- Two exits made stricter: a graceful stop now waits for the ingerain within `server.shutdown_timeout` (open connectionsforce-closed at the deadline), and the `mq.max_bytes_gb` reload hook bounds its JetStream calls to 10s. The e2e orchestrator's kill budget is raised to 20s to
match.
- Docs, CHANGELOG, labeler (`area/app`), and stale `main.go` code comments synced.

## Test plan

- [ ] New unit tests: degraded boot without ClickHouse, dedupe on/off, Prometheus inline vs sidecar, Run stops cleanly on cancel, listen failure names the component, settings reload drives the relocated hooks, `run(ctx)` boot/refusal paths

## Related Issues

Closes #140. Part of #583 (story 0).

Issue #583, opened 2026-09-10
epic(settings): multi-tenant settings directory and tenant-scoped runtime


taitelee:
**Area:** settings / api / streaming — multi-tenant preparation epic · decided with @EricAndrechek on 2026-09-10

WaveHouse runs one settings snapshot per process. WaveHouse Cloud needs one process to serve many tenants, each with its own roles, policies, pipes, and config, delivered by the cloud fan-out as files on disk exactly like standalone. This epic makes the settings directory, the in-memory snapshots, and every request path tenant-aware without changing anything for a standalone operator. Split into the stories below as they are promoted to Ready (the #194 pattern).

## Decisions

- **Two directory shapes, no hybrid.** *Flat*: the four files (`config.json`, `policies.json`, `pipes.json`, `roles.json`) sit in the root and are tenant `0`. *Nested*: the root holds only subfolders, each folder name is a tenant id, each folder holds the four files. Loose files and subfolders in the same root is an error at boot and at reload.
- **Tenant id is a validated string.** Letters, digits, `_`, capped length — exactly the `query.SafeEncodeNATS` pass-through set, so the id appears verbatim in MQ subjects; folder name is the id. `0` is the reserved default. Not an integer: 19-digit ids already round when treated as numbers (`internal/auth/auth.go`).
- **Requests resolve to a tenant before auth.** The id comes from a request header; absent means `0`. An id with no snapshot is `404`. In flat mode tenant `0` always exists. In nested mode it exists only if a `0` folder does, and operators are advised not to create one. Putting the header on every request is the client's or proxy's job.
- **Tenant-exempt routes.** `/livez`, `/readyz`, metrics, version, and the ops reload endpoint are deployment-wide and skip tenant resolution. Reload reloads the whole tree.
- **Reload swaps per tenant.** Each folder validates independently; a rejected folder keeps that tenant's previous snapshot. The tree-wide shape check is the exception: a hybrid root rejects the whole reload. Boot stays all-or-nothing (#530 decision: boot fails loudly).
- **Every folder carries a full `config.json`.** ClickHouse, auth, dedupe, DLQ, query, schema, stream, MQ, CORS are all per tenant. This is the isolation model for cloud and means the runtime resources behind those blocks become per tenant too (see stories 6–10).
[...]

Issue #235, opened 2026-06-04
Clarify/implement scope (currently always empty)


EricAndrechek:
**Area:** ingest — confusion / FYI · found via WaveHouse-Stats dogfooding

**Expected:** know whether `scope` (in `EventMessage`) provides any isolation today.

**Actual:** it's always empty — the multi-tenant scope concept exists in the wire format but isn't populated or used (`internal/api/ingest.go:150-151`, with matching `// TODO: impl scope` in `structured_query.go:121` and `stream.go:54`).

**Impact (Stats):** none now (single-tenant); flagging so nobody assumes `scope` isolates anything if tenants are added.

**Scope:** document that `scope` is currently inert, and/or implement population as part of multi-tenant work.

Related: per-table policy storage + per-table authorization scopes #214 (the policy side of multi-tenant; `scope` is the runtime-population half).

---
_From `WAVEHOUSE-FEEDBACK.md` (WaveHouse-Stats dogfooding); validated by code-read against `0f8826c` on 2026-06-04._

EricAndrechek:
**Additional scope-TODO sites (from a codebase TODO sweep, 2026-06-04)** — beyond the ingest/query/stream `// TODO: impl scope` markers already noted, scope is also unimplemented in the pipes + DLQ paths:
- `internal/api/pipes.go:140` — `// TODO: scope impl`
- `internal/api/dlq.go:51` — `// TODO: do we need to break out scopes here?`

Worth covering these when scope is implemented so it's consistent across ingest / query / stream / pipes / DLQ.

Issue #20, opened 2026-04-13
Bridge Bento Processing into Distributed Traces


taitelee:
## Problem

Bento operates as a "middleware" between NATS and ClickHouse. If Bento transformations are slow or failing, those errors appear as isolated events. We need these transformations to show up as "Child Spans" directly under the original API ingestion trace.

## Proposed Solution

Configure the Bento OTel component to respect incoming tracing headers.
- Ensure Bento is initialized with the same TracerProvider as the Go app.
- Use the tracing processor in Bento to ensure that as events move through "Pipes," they maintain the original Trace ID.

## Alternatives Considered

None

## Additional Context

This allows a developer to see: API Request -> NATS Publish -> Bento Pipe Transformation -> ClickHouse Insert all in one single Gantt chart in SigNoz.

PR #586, opened 2026-09-15
refactor(mq): seal the MQ boundary behind an intent-level broker API

taitelee:
## Summary

`internal/mq` is now the only package that imports NATS/JetStream **and** the only one that knows how the broker works. The import half is enforced by a `depguard` rule in `.golangci.yml` (Key Design Decision #20). The semantic half: nothing outside `internal/mq` builds a subject, names a stream, or reasons in sequences, so story 5's tenant token lands in `mq.Topic` and `internal/mq/subject.go` with no call-site edits.

- **Addressing.** Events are addressed by `mq.Topic{Table, Scope}` with raw names. The `ingest.`/`dlq.` prefixes, `>` wildcards, stream names, and the subject-token encoder (formerly `query.SafeEncodeNATS`/`SafeDecodeNATS`) are private to `internal/mq/subject.go`. A delivered `Message` exposes `TopicKey()` (free; the delivered form) and `Topic()` (decodes on demand), so the per-message path decodes nothing.
- **Intent-level interfaces.** `Publisher` (`ErrQueueFull` is the backpressure signal; the API no longer matches broker error text), `Subscriber`, `ConsumerManager`/`Consumer`, `DeadLetterer` and `DeadLetterStats`, `Purger.PurgeAcked`, `Replayer.ReplaySince`, composed into `mq.Broker`. `internal/app` holds a `Broker`; `mq.NewEmbedded` is the one place the implementation is chosen.
- **Moved into mq, logic unchanged.** The Active Sweeper's ack-floor/binary-search arithmetic (`internal/ingest/sweeper.go` → `internal/mq/purge.go`; the sweeper keeps the schedule and the gap window), and the `mq.max_bytes_gb` reload from `internal/app` (`SetMaxBytes`: ingest stream and the DLQ at a tenth of it resized as a pair, with the rollback and time bounds from #585; `NewEmbedded` creates both streams, replacing `api.EnsureDLQStream` and `Resize`).
- **Raw types gone.** The accessors `JetStream()`, `NatsConn()`, `GetServer()`, the unread `api.Dependencies.JS`, and the never-wired cache connection parameter are removed. `mq.Headers` + `WithHeader` replace `*nats.Msg` options; `observability.InjectHeaders`/`ExtractHeaders` work over a plain header map; `RegisterSystemMetrics` takes a `func() (MQStats, error)`.
- **Cache.** Keeps its namespace-token encoder as the broker-neutral `query.SafeEncodeToken` (same output, cache keys unchanged).
- **Test mocks.** `MockMessage`, `MockPurger`, `MockDeadLetterStats`; `MockPublisher` records the topic and headers of every publish and dead-letter parking.

Wire behavior is preserved: subjects, stream names, consumer settings, ack semantics, and the `X-DLQ-*` headers are byte-for-byte what they were; parking on the DLQ is a prefix swap on the delivered subject. Merged with main (#585), porting `internal/app` to the mq-owned types and keeping its shutdown-cancelled gap-fill (`ReplaySince` checks its context between pulls).

Deliberate behavior differences, none on the wire:

- The sweeper warns only when the buffer consumer genuinely does not exist yet, and errors on any other lookup failure.
- The sweeper's binary search no longer discards the lower half of the stream when its midpoint holds no message, which could place the purge bound inside the replay window. A sequence with no message is kept as a candidate bound (purging less, never more); a lookup that fails outright aborts that sweep.
- `GET /v1/ops/dlq/stats` returns the documented `500` when the broker cannot be read; only a genuinely absent queue reads as empty.
- The ingest worker's delivery handoff watches its context, so a delivery in flight at stop returns instead of blocking on a full channel.
- SSE gap-fill logs a pull that fails outright at `WARN` instead of ending the replay silently; a cancelled replay is not logged as a failure.
- DLQ publishes go through the same header injection as every publish (trace headers when the context carries a span; the worker's flush context never does).

Deferred: terminal consumer errors on `mq.Consumer` (#587); breaking scope out of DLQ counts (#235). The `integration`-tagged tests sit outside lint's build context, so the boundary there rests on convention (documented).

## Test plan

- [x] CI green on `1b656c12` (lint incl. the `depguard` boundary rule, unit, integration, e2e, coverage, docs build)
- [x] `internal/mq`: subject round-trip and injectivity, dead-letter parking and counts (incl. a foreign subject parking under the same tail), `ErrQueueFull`, `PurgeAcked` end to end, `SetMaxBytes` pair resize / rollback / cancelled context, replay incl. cancellation, consumer config read-back, headers, stats
- [x] Purge arithmetic: the 16 sweeper cases ported, plus a hole after / straddling / before the boundary and failed lookups
- [x] Sweeper asserts the consumer name and cutoff it asks for; worker tests run against `MockPublisher`/`MockMessage` and the real embedded server; app asserts a reload reaches the broker's byte budget
- [x] E2E SDK suite covers ingest → SSE delivery through the real broker (the hub-key change)

## Related Issues

Part of #583 (story 4, plus the broker abstraction story 5 builds on). Follow-ups: #587, #235.


taitelee:
@coderabbitai review

taitelee:
@coderabbitai review

internal/mq/embedded.go
```diff
@@ -183,13 +218,135 @@ func (e *EmbeddedNATS) Subscribe(ctx context.Context, subject, consumerName stri
...
+}
+
+func (c *jsConsumer) Consume(handler func(msg *Message), prefetch int) (func(), error) {
+	var opts []jetstream.PullConsumeOpt
+	if prefetch > 0 {
+		opts = append(opts, jetstream.PullMaxMessages(prefetch))
+	}
+	cctx, err := c.cons.Consume(func(m jetstream.Msg) {
+		handler(wrapMsg(c.ctx, m))
+	}, opts...)
```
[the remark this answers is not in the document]
taitelee: Agreed this is a real gap, but it predates the PR (the old worker did not pass ConsumeErrHandler either) and it is a design change, not a refactor: it needs an error path on mq.Consumer and a restart-or-terminate policy in the dispatch loop. This PR is the story 4 boundary move and is meant to be behavior-preserving, so I'd rather land that as its own change. Worth noting the practical exposure today is narrow: the broker is in-process with DontListen, so there is no connection to drop and no external client to delete the durable consumer. It becomes real once the boundary is backed by a remote broker. Tracking in #587.

internal/mq/purge.go
```diff
@@ -0,0 +1,129 @@
...
+	lo, hi := first, last
+	result := last + 1 // Default: everything is older than the cutoff.
+
+	for lo <= hi {
+		mid := lo + (hi-lo)/2
+		ts, err := s.messageTime(ctx, mid)
+		if err != nil {
+			// Sequence may have been purged; scan forward.
+			lo = mid + 1
+			continue
```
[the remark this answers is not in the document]
taitelee: Confirmed, and fixed in 1b656c12. Your counter-example reproduces: with the boundary at 40 and 45–55 holding no message, the old search answered 56, which would have purged 40–44 inside the replay window.

This predates the PR (the search moved here verbatim from `internal/ingest/sweeper.go`), but it decides what the sweeper deletes, so it is fixed here rather than deferred. A sequence with no message is now kept as a candidate bound and the search continues below it, so the result can only move toward purging less; the next sweep starts past the hole. Separately, `messageTime` now distinguishes "no message at this sequence" (`jetstream.ErrMsgNotFound`) from a failed lookup, and a failed lookup aborts the sweep instead of being read as a missing sequence.

With this stream's configuration (limits policy, `DiscardNew`, prefix-only purges) interior holes cannot occur, so the realistic trigger was a transient lookup error.

Tests added: a hole after the boundary (your case), a hole straddling it, a hole before it (under-purges, never over), and failed lookups at the midpoint and at the oldest message.

internal/api/dlq.go
```diff
@@ -1,73 +1,38 @@
...
-		if err != nil {
-			continue
+		if !errors.Is(err, mq.ErrNoDeadLetterQueue) {
+			h.Logger.ErrorContext(r.Context(), "dlq stats failed", "error", err)
+			writeJSONError(w, http.StatusInternalServerError, "stream info failed")
+			return
 		}
-		tables[decodedSubject] = count
+		// No dead-letter queue: nothing can have been parked.
+		counts = mq.DeadLetterCounts{Tables: map[string]uint64{}}
```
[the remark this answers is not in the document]
taitelee: Fixed in 1b656c12. `DeadLetterCounts` now wraps `ErrNoDeadLetterQueue` only for `jetstream.ErrStreamNotFound`; any other lookup failure propagates and the handler returns the documented `500`. This resolves the old `TODO: catch by error type` and matches how this PR already treats the sweeper's consumer lookup.

Covered by an mq test (a closed connection must not read as `ErrNoDeadLetterQueue`) and the handler's `BrokerFailureIsAnError` test. Listed as a behavior difference in the changelog entry.

internal/mq/embedded.go
```diff
@@ -108,65 +144,142 @@ func NewEmbedded(storeDir string, maxBytes int64, logger ...*slog.Logger) (*Embe
...
+	defer cancel()
+	if _, err := e.js.UpdateStream(resizeCtx, ingestStreamConfig(maxBytes)); err != nil {
+		return fmt.Errorf("resize ingest stream: %w", err)
+	}
+	if _, err := e.js.CreateOrUpdateStream(resizeCtx, dlqStreamConfig(maxBytes/dlqShare)); err != nil {
+		// The undo runs on its own budget, not the one the DLQ call has
+		// likely just exhausted.
+		rollbackCtx, cancelRollback := context.WithTimeout(ctx, rollbackTimeout)
+		defer cancelRollback()
+		if _, rollbackErr := e.js.UpdateStream(rollbackCtx, ingestStreamConfig(e.maxBytes)); rollbackErr != nil {
```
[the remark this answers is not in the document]
taitelee: Declining the behavior change; fixing the contract wording (1b656c12).

Rooting the rollback in the caller's context is deliberate and unchanged from #585: that context is the process's stop context, and the decision there was that a reload caught mid-hook by a stop gives up — undo included — rather than holding the drain past `server.shutdown_timeout`. A detached rollback would do exactly that for up to five seconds. The split state you describe lasts only for the remainder of a process that is already exiting; the next boot reconciles both streams from the adopted settings.

You are right that the `Broker.SetMaxBytes` doc overpromised ("the previous budget stays in effect"). It now says the restore is best effort, that a cancelled context abandons it, and that `MaxBytes` keeps reporting the previous budget so the next call reapplies both.

No test added: cancelling deterministically between the two JetStream calls needs an injection seam this package doesn't have, and I'd rather not add one for a path whose behavior is unchanged.

internal/mq/embedded.go
```diff
@@ -183,13 +296,200 @@ func (e *EmbeddedNATS) Subscribe(ctx context.Context, subject, consumerName stri
...
+		if err != nil {
+			continue
+		}
+		counts.Tables[name] = n
+	}
+	return counts, nil
+}
+
+// ReplaySince creates an ephemeral consumer on topic's ingest subject starting at
+// since (DeliverByStartTime) and drains it to send until caught up. The
```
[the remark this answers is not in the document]
taitelee: Deferring, with the contract corrected (1b656c12).

This is the pre-existing behavior (the handler carried the same `TODO: do we need to break out scopes here?`), and scope is inert today: `/v1/ingest` never sets it, and how scopes surface anywhere is #235's decision. Choosing an aggregation and a filter semantic here would be designing that ahead of it, in a PR meant to preserve behavior.

The interface doc did promise plain per-table counts, which was wrong; `DeadLetterCounts.Tables` and `DeadLetterStats` now state that a scoped topic counts under `table.scope` and that the table filter matches unscoped messages, until #235.

EricAndrechek:
@coderabbitai review

docs/src/content/docs/architecture.md
```diff
@@ -133,22 +133,26 @@ The SSE fan-out, factored out of `api/` so the delivery hot path ([#294](https:/
...
...go** — `EmbeddedNATS`, the one `Broker`: an in-process NATS server with JetStream. Creates stream `WAVEHOUSE` with subjects `ingest.>`, capped at the settings directory's `mq.max_bytes_gb`, and stream `WAVEHOUSE_DLQ` (`dlq.>`, `DiscardOld`) at a tenth of it — always present, since an empty stream costs nothing. `SetMaxBytes` applies a reloaded budget to both live streams as a pair: if the DLQ update fails after the ingest one succeeded, the ingest resize is undone so both stay on the previous budget and the next reload retries. Its JetStream calls are bounded to ten seconds, plus five more for the rollback (a budget of its own, not the one that just expired), since a reload holds the settings store's lock while its hooks run; `MaxBytes` reports the budget in effect. `Stats` reports connection and inbound-message counters for `observability.RegisterSystemMetrics`. Trace context rides in the message headers: `Publish` applies `observability.InjectHeaders`, and a message delivered through `Subscribe` (the hub bridge) carries `observability.ExtractHeaders` on its `Ctx`; the worker's `Consumer` path skips the extraction, since it batches across messages and reads no per-message context.
```
EricAndrechek: **[SHOULD] This overstates what the rollback guarantees, and the second half is load-bearing.**

`SetMaxBytes` resizes the ingest stream, then the DLQ. If the DLQ update fails it undoes the ingest resize — but that undo is itself best effort, and the code says so (`internal/mq/embedded.go:187`):

> The undo is best effort: if it fails too, the ingest stream stays at the new limit and the DLQ at the previous, and the error says so.

So "both stay on the previous budget" is not always true.

The `MaxBytes` half matters more. It does not report "the budget in effect" — `internal/mq/mq.go:272` defines it as "the budget last applied in full", meaning that after a failure it keeps reporting the *previous* budget. That is the retry mechanism, not a nuance: `internal/app/wire.go:259` skips the reload when `mb == broker.MaxBytes()`. If a failed resize moved `MaxBytes`, that guard would match and the resize would never be retried. Describing it as "in effect" tells a reader the opposite of why the code is written this way.

Same claim, second copy, user-facing: `docs/src/content/docs/settings-directory.mdx:192` says "If NATS rejects the update, the previous limit stays in effect and the failure is logged" — true unless the undo also fails. Worth softening to "the reload does not take effect and the failure is logged" while you are in the claim.

```suggestion
- **embedded.go** — `EmbeddedNATS`, the one `Broker`: an in-process NATS server with JetStream. Creates stream `WAVEHOUSE` with subjects `ingest.>`, capped at the settings directory's `mq.max_bytes_gb`, and stream `WAVEHOUSE_DLQ` (`dlq.>`, `DiscardOld`) at a tenth of it — always present, since an empty stream costs nothing. `SetMaxBytes` applies a reloaded budget to both live streams as a pair: if the DLQ update fails after the ingest one succeeded, the ingest resize is undone so both stay on the previous budget — best effort, since if that undo also fails the ingest stream stays at the new limit and the DLQ at the previous, and the error says so. Its JetStream calls are bounded to ten seconds, plus five more for the rollback (a budget of its own, not the one that just expired), since a reload holds the settings store's lock while its hooks run; `MaxBytes` reports the budget last applied in full, so a failed resize is retried by the next reload. `Stats` reports connection and inbound-message counters for `observability.RegisterSystemMetrics`. Trace context rides in the message headers: `Publish` applies `observability.InjectHeaders`, and a message delivered through `Subscribe` (the hub bridge) carries `observability.ExtractHeaders` on its `Ctx`; the worker's `Consumer` path skips the extraction, since it batches across messages and reads no per-message context.
```

internal/mq/mq.go
```diff
@@ -83,23 +156,122 @@ type Subscriber interface {
...
+	DeadLetter(ctx context.Context, msg *Message, opts ...PublishOpt) error
+}
+
+// DeadLetterCounts is what is parked on the dead-letter queue.
+type DeadLetterCounts struct {
+	// Tables maps table name → parked messages, for the tables asked about.
+	// Scope is not broken out yet (it is inert until #235): a message parked
+	// under a scoped topic counts under "table.scope", not under its table.
+	Tables map[string]uint64
+	// Total is every parked message, whatever the filter.
```
EricAndrechek: **[SHOULD] This branch turns "Total ignores the filter" into a contract, but the endpoint doc still implies the opposite.**

`DeadLetterCounts(ctx, "default.orders")` narrows `Tables` and not `Total`: `embedded.go` sets `Total: state.Msgs`, the whole stream's message count regardless of the subject filter. `internal/mq/embedded_test.go:452` pins it — `Tables{default.orders: 2}` with `Total: 3`.

`docs/src/content/docs/api.md:737` describes the parameter as "Filter stats to a specific table name (e.g., `?table=clicks` returns only the `clicks` count)", and the example response just below it happens to show `total` equal to the sum of `tables`. Nothing tells the reader the two fields scope differently. Someone alerting on `?table=x` and reading `total` gets the whole queue's depth and will not notice.

The behaviour is unchanged from main, so this is not a regression — it is worth fixing here because this is the PR that made it an mq-owned contract. `api.md` is outside this diff; the fix is to append to that parameter description:

> — `total` is always the whole dead-letter queue, not the filtered table.

The same sentence next to `wh.dlq.table('clicks')` in `docs/src/content/docs/sdk/admin.md` would keep the two in step.

internal/api/stream.go
```diff
@@ -200,32 +198,16 @@ func (h *StreamHandler) replayContext(r *http.Request) (context.Context, context
...
+// replay sends every message retained on topic since the given time to the
+// callback until caught up or ctx is done (the client went away, or the
+// server is shutting down — a long gap-fill must not hold the drain any more
+// than a live stream would). A replay that cannot start, or that fails before
+// catching up, is not fatal to the stream — the client still gets live events
+// from here on — so the error is logged rather than ending the connection. A
+// done ctx is the connection ending, not a failure, and is not logged.
+func (h *StreamHandler) replay(ctx context.Context, since time.Time, topic mq.Topic, send func([]byte) bool) {
+	if err := h.Replayer.ReplaySince(ctx, topic, since, send); err != nil && ctx.Err() == nil {
+		slog.Default().WarnContext(ctx, "gap-fill replay ended early; the client continues with live events only",
```
EricAndrechek: **[MAY] This is the only handler in `internal/api` reaching for the global logger, which also makes the warning untestable.**

Every sibling takes an injected one — `NewDLQHandler(stats, logger)`, `NewIngestHandler(..., logger)`, `NewPipesHandler(..., logger)` — and `internal/app/wire.go:448` already has `logger` in scope where it sets `Metrics`, `Heartbeater` and `Closing`. A `Logger *slog.Logger` field set alongside them would let a test assert this WARN fires on a replay that ends early.

Reasonable to push back: in production `slog.Default()` is the wired logger, so behaviour is identical either way. This is consistency and testability, not correctness — the lowest-priority item here.

EricAndrechek (dismissed):

EricAndrechek (dismissed):

taitelee:
@coderabbitai review

EricAndrechek:
@coderabbitai review

internal/ingest/worker_test.go
```diff
@@ -1548,9 +1522,131 @@ func TestRejectPoison_CountedByDisposition(t *testing.T) {
...
+	}
+	failed := make(chan error, 1)
+	go func() {
+		for _, m := range f.msgs {
+			handler(m.Message())
+		}
+		failed <- f.endWith
+	}()
+	return func() { f.stopped.Store(true) }, failed, nil
+}
```
EricAndrechek: **[MUST] This test is flaky — measured, not theorised.** `go test ./internal/ingest -run TestDispatchLoop_DeliveryEndedFailsLoud -race -count=300`, run 20 times: **7 runs failed, 14 bad iterations out of 6,000 (~1 in 430)**, always at `:1605` with `expected: 1, actual: 0`. Zero data races across all 6,000 — this is a test bug, not a production one.

`handler(m)` only buffers into the 200-deep `msgChan`; `failed <- f.endWith` then fires with no ordering against it. If the loop has not reached its select yet, **both** cases are ready and Go picks uniformly at random. Taking `deliveryEnded` first is correct — `worker.go:270-274` says messages still in `msgChan` are unacked and redelivered — so the test asserts a guarantee the worker deliberately does not make.

Fix by ordering, not by retrying: gate the end signal so exactly one case is ready when the loop selects.

```suggestion
type failingFakeConsumer struct {
	msgs     []*testutil.MockMessage
	endWith  error
	endAfter chan struct{} // if set, delivery ends only once the test releases it
	startErr error
	stopped  atomic.Bool
}

func (f *failingFakeConsumer) Consume(handler func(*mq.Message), _ int) (func(), <-chan error, error) {
	if f.startErr != nil {
		return nil, nil, f.startErr
	}
	failed := make(chan error, 1)
	go func() {
		for _, m := range f.msgs {
			handler(m.Message())
		}
		if f.endAfter != nil {
			<-f.endAfter
		}
		failed <- f.endWith
	}()
	return func() { f.stopped.Store(true) }, failed, nil
}
```

The caller then releases once the row is provably in hand. The catch: with `maxBatch = 100` / `maxWait = time.Hour` only the shutdown flush can write the row, and nothing observable says it has left `msgChan` — `MockMessage` exposes only `Acked`/`DoubleAcked`, both post-flush. So either:

- set `w.maxBatch = 1` so arrival flushes, `require.Eventually(..., rt.Hits() == 1, ...)`, then release — deterministic, but `:1605` becomes tautological and the drain-on-failure path loses coverage, so it wants its own test; or
- keep `maxBatch = 100` and gate on something proving `parseMsg` ran (it runs on the dispatch goroutine after the receive, so a mock-cache call count works).

docs/src/content/docs/ingest-pipeline.md
```diff
@@ -199,6 +199,12 @@ If the deadline fires first, `waitOrDeadline` returns the deadline error and the
 
 Messages still sitting in `msgChan` or the consumer's prefetch buffer at shutdown are **not** flushed; they are simply redelivered next boot. Graceful shutdown flushes the in-hand per-table batches, not the entire in-flight pipeline.
 
+### When the consumer dies
```
EricAndrechek, line 202: **[MUST] This section now contradicts line 266 of this same file.**

Line 266 still reads:

> **NATS resilience.** Remote NATS needs explicit reconnect/backoff and a `Consume` error handler — none of which the embedded path needs.

This commit adds `jetstream.ConsumeErrHandler(...)` unconditionally in `embedded.go:329-343`. That *is* a `Consume` error handler, and `embedded.go` *is* the embedded path. The claim was true before this commit; this commit is what makes it false. A reader going from here to "Scaling to multiple instances" gets opposite answers 60 lines apart.

Line 266 is outside this diff, so no suggestion block. Replacement:

```markdown
- **NATS resilience.** Remote NATS needs explicit reconnect/backoff for the connection itself — the embedded path never dials out, so there is nothing to reconnect. The `Consume` error handler that detects a dead consumer already lives in `embedded.go` and needs no change for a remote broker.
```

internal/ingest/worker.go
```diff
@@ -205,21 +215,29 @@ func waitOrDeadline(ctx context.Context, wg *sync.WaitGroup) error {
...
+	// forever. A message dropped here is unacked and simply redelivered.
+	stop, deliveryEnded, err := cons.Consume(func(msg *mq.Message) {
+		select {
+		case msgChan <- msg:
+		case <-ctx.Done():
+		}
+	}, pullMaxMessages)
 	if err != nil {
 		w.logger.Error("failed to start consumer", "error", err)
+		w.failed <- fmt.Errorf("ingest worker: start consumer: %w", err)
```
EricAndrechek: **[SHOULD] Stutters in the operator-facing log.** `app.go:207` already wraps every component error with `c.name`, which is `"ingest worker"` (`wire.go:337`). What gets logged at 3am is:

`run error="ingest worker: ingest worker: consumer delivery ended: nats: consumer deleted"`

Naming the component is the app's job, and this component's other error path (`wire.go:340`) returns bare — so one component now emits two shapes.

```suggestion
		w.failed <- fmt.Errorf("start consumer: %w", err)
```

internal/ingest/worker.go
```diff
@@ -248,6 +266,16 @@ func (w *IngestWorker) dispatchLoop(ctx context.Context, cons jetstream.Consumer
...
 			return
+		case err := <-deliveryEnded:
+			// The MQ gave up on the consumer: no message will arrive again, so
+			// waiting on msgChan would stall ingestion silently. Flush and ack
+			// what is already in hand — those rows are delivered and the
+			// publish path is not what broke — then fail loud. Messages still
+			// in msgChan are unacked and redelivered to the next consumer.
+			w.logger.Error("ingest consumer delivery ended; ingestion has stopped", "error", err)
+			shutdown()
+			w.failed <- fmt.Errorf("ingest worker: %w", err)
```
EricAndrechek: **[SHOULD]** Same stutter as `:237`; the error already arrives wrapped as `mq.ErrDeliveryEnded`.

```suggestion
			w.failed <- err
```

internal/app/app_test.go
```diff
@@ -394,6 +360,41 @@ func TestRun_ListenFailureStopsEverything(t *testing.T) {
...
+	a.mq = dyingConsumerBroker{Broker: a.mq, reason: fmt.Errorf("%w: consumer deleted", mq.ErrDeliveryEnded)}
+
+	// Bounded so a worker failure that goes unnoticed fails the assertion
+	// below (a cancelled Run returns nil) instead of serving forever — which
+	// is exactly the silent stall this guards against.
+	ctx, cancel := context.WithTimeout(t.Context(), 10*time.Second)
+	defer cancel()
+	err := a.Run(ctx)
+	require.ErrorIs(t, err, mq.ErrDeliveryEnded, "an API that accepts events nothing writes must not keep running")
+	assert.True(t, strings.HasPrefix(err.Error(), "ingest worker: "), "the failing component names itself: %v", err)
```
EricAndrechek: **[SHOULD] This assertion is too weak to catch the stutter above.** It is copy-shaped from `:356`'s `"http server: "` check, which does not stutter — so the two lines look identical and mean different things. `HasPrefix` passes happily on `"ingest worker: ingest worker: ..."`.

```suggestion
	assert.True(t, strings.HasPrefix(err.Error(), "ingest worker: "), "the failing component names itself: %v", err)
	assert.NotContains(t, err.Error(), "ingest worker: ingest worker: ", "exactly once: %v", err)
```

EricAndrechek (dismissed):

EricAndrechek (approved):

taitelee:
@coderabbitai review

PR #593, opened 2026-09-18
feat(tenant): resolve X-Tenant-ID before auth and thread the tenant

taitelee:
## Summary

Story 1 of the multi-tenant epic: every request resolves to a tenant before authentication, and the tenant is threaded through every settings read. No behavior change for a deployment that sends no tenant header — only tenant `0` exists.

- New `internal/tenant`: `ID` (a validated string — letters, digits, `_`, `-`, ≤ 64 bytes, safe as a folder name and an MQ subject token), `Default = "0"`, `Header = "X-Tenant-ID"`. Imports nothing from the repo.
- `settings.Registry` (`For(id)`) over the one store `settings.Open` adopts, keyed `0`. Open, reload and the watcher are untouched.
- `api.TenantMW` runs ahead of auth on every `/v1` route outside `/v1/ops/*`: absent or empty header is tenant `0`, malformed or repeated is `400`, unknown is `404`; the resolved `*settings.Store` rides the request context. Handlers read it once and pass it down — the ingest, structured-query and pipe getters take the store as a parameter, and a tenant route reached without one answers `500` rather than fall back to a tenant. The async paths (worker, sweeper, hub, schema registry) are constructed with a `tenant.ID` and their getters take it, wired with `tenant.Default` in `internal/app`.
- The probes, `/version`, the metrics path and `/v1/ops/*` stay tenant-exempt; the admin pipe reads name their tenant with an optional `?tenant=` (same grammar and answers). `X-Tenant-ID` joins the CORS allow-headers list so browser clients can send it.
- The slog cleanup deferred from #586: no constructor takes a `*slog.Logger` anymore (mq, the handlers, auth, discovery, sweeper, worker, plus `settings.Open`, `chconn.Open` and the `config` data-dir helpers); call sites use the context-aware calls. Tests reach log output through the new `internal/testutil/logtest` (`Silence` in `TestMain`, `Capture` for log-asserting tests, which run serially). `testutil.NopLogger` is removed.

Two commits: the tenant work, then the logger cleanup.

## Test plan

- [x] `make ci` passes locally (unit, integration, e2e, coverage gates; Go total 92.8%)
- [x] `internal/tenant`: grammar table test (default, 19-digit id, length cap, dots, slashes, wildcards, non-ASCII)
- [x] `api.TenantMW`: no header / empty / `0` / unknown `404` / malformed `400` / repeated `400`; tenant resolves before `AuthMW`; exempt routes ignore the header; tenant-route handlers `500` without a resolved store
- [x] `internal/app`: the wired registry answers `404`/`400` end to end; ops ignores the header
- [x] ops pipe reads: `?tenant=` table test (absent, empty, `0`, unknown, malformed, repeated)
- [x] logger: log-asserting tests in api, auth, discovery, config converted to `logtest.Capture`
- [ ] Manual: `curl -H 'X-Tenant-ID: acme' /v1/health` → `404`; no header → as before

## Follow-ups

- Which tenant an ops read addresses in nested mode is settled here as `?tenant=`; story 2 gives the reload route the same parameter.
- The SDK's `wh.pipes.list()` / `.get()` cannot send `?tenant=` yet (`options.headers` cannot set a query parameter); a one-option addition once a second tenant can exist.
- An unknown tenant on an async path reads as the getter's zero value (DLQ off, nil policy, gap window 0) — impossible while only tenant `0` exists; story 3 defines a removed tenant's semantics.
- Auth's policy source (story 9) and the `chconn`-backed getters (story 6) are not threaded.
- Draft #589 overlaps this diff heavily; ordering to be agreed.

## Related Issues

Part of #583 (story 1). Finishes the logger cleanup deferred from #586.


internal/api/tenant.go
```diff
@@ -0,0 +1,100 @@
...
+
+// opsTenantParam is the query parameter an ops route takes its tenant from.
+// The ops tree is tenant-exempt — it runs no TenantMW and ignores the header
+// — so an admin names the tenant explicitly, and none means tenant.Default.
+const opsTenantParam = "tenant"
+
+// opsStore resolves the tenant an ops route addresses from its
+// opsTenantParam, with the same answers as TenantMW.
+func opsStore(w http.ResponseWriter, r *http.Request, tenants *settings.Registry) (*settings.Store, bool) {
+	return resolveTenant(w, tenants, "?"+opsTenantParam, r.URL.Query()[opsTenantParam])
```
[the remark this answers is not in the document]
taitelee: `opsStore` parses the raw query with `url.ParseQuery` and answers `400 invalid ?tenant: malformed query string` when it fails, with a test for `?tenant=acme;x=1`. Deliberately the whole query rather than the tenant field only as the parser doesn't report which pair was malformed, and on an admin-only read a strict `400` is the safer default.

internal/stream/hub.go
```diff
@@ -74,14 +76,19 @@ type topicRoutes struct {
...
+// NewHub builds the event hub of tenant id. A nil policy store passes every event through
 // unfiltered (the unwired-tests case); a non-nil store whose Get returns nil is a
 // total lockout (a deleted/absent policy denies everyone). A nil registry leaves
 // every column's type unknown, so row-filter comparison degrades FAIL-CLOSED:
 // equality/set predicates admit only a byte-identical value and ordering/!= admit
 // nothing (see policy.ColumnKind); metric may be nil.
-func NewHub(policyStore policy.Source, registry *discovery.SchemaRegistry, metric *Metrics) *Hub {
-	return &Hub{topics: make(map[string]*topicRoutes), policy: policyStore, registry: registry, metric: metric}
+func NewHub(id tenant.ID, policyStore PolicySource, registry *discovery.SchemaRegistry, metric *Metrics) *Hub {
+	return &Hub{topics: make(map[string]*topicRoutes), tenant: id, policy: policyStore, registry: registry, metric: metric}
```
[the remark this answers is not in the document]
taitelee: By design for this PR. #583 story 1 threads the tenant and holds one store keyed 0; story 5 ('Stream hub subscriptions and per-role frame projection keyed by tenant live here') gives the hub its tenant from the MQ subject. Until story 2 lands there is no second tenant, so no other tenant's policy can be reached through the default hub. Tracked in the epic, not here.

taitelee:
@coderabbitai review

docs/src/content/docs/settings-directory.mdx
```diff
@@ -189,7 +189,7 @@ The `WAVEHOUSE_DLQ` stream always exists (an empty stream costs nothing) and the
...
+- `mq.max_bytes_gb` (seed default `50`) — disk budget for the embedded JetStream `WAVEHOUSE` stream that buffers ingested events until the worker writes them to ClickHouse; the `WAVEHOUSE_DLQ` stream gets a tenth of it. The stream runs `DiscardNew`, so when it's full new publishes are rejected and `POST /v1/ingest` returns `503` — [backpressure by construction](/ingest-pipeline#backpressure-and-durability-knobs). A reload updates both streams' limits in place without touching what's buffered: growing takes effect immediately; shrinking below what's currently on disk makes the stream refuse new publishes until the worker drains it back under the limit — nothing already accepted is dropped. If NATS rejects the update, the rest of the reload is still adopted; the streams stay on the previously applied budget, the failure is logged, and the next reload retries it. Size it from [Durability & Storage](/durability).
```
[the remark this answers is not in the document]
taitelee: Fixed. The sentence now states the pair semantics — a failed DLQ resize undoes the ingest one, and if the undo fails too the ingest stream keeps the new limit and the DLQ the previous until a later reload, with the log line naming which.

taitelee:
@coderabbitai review

taitelee:
@coderabbitai review

taitelee:
@coderabbitai review

taitelee:
@coderabbitai full review

docs/src/content/docs/api.md
```diff
@@ -50,6 +50,29 @@ WaveHouse extracts the role from a configurable JWT claim path (`auth.role_claim
...
+| Status | Body | When |
+| ------ | ---- | ---- |
+| `400` | `{"error": "invalid X-Tenant-ID: …"}` | The id breaks the grammar above, or the header was sent more than once |
+| `404` | `{"error": "unknown tenant: <id>"}` | The id is well formed but no such tenant exists |
+
+Both are decided before authentication, so they are returned whatever token the request carries. Every response from a tenant route, these two included, carries `Vary: X-Tenant-ID`, so a shared cache keys on the header and never replays one tenant's response to another.
+
+The probes (`/livez`, `/readyz`, `/healthz`), `/version`, the Prometheus metrics path, and `/v1/ops/*` are tenant-exempt: they ignore the header entirely. The two admin pipe reads ([`GET /v1/ops/pipes`](#get-v1opspipes--list-named-pipes) and `GET /v1/ops/pipes/{name}`) name their tenant with an optional `?tenant=` query parameter instead, with the same grammar and the same `400`/`404` answers; no other ops route takes a tenant.
+
+`X-Tenant-ID` is in the CORS `Access-Control-Allow-Headers` list, so a browser client can send it cross-origin. The SDK sends it through [`options.headers`](/sdk#custom-headers).
```
EricAndrechek: I think I'd like to move all multi-tenant stuff to either a new docs page or to the deployment and ops type docs page – it will confuse people where in 99% of cases they don't need to set or worry about multi-tenant deployments, so we should just not even mention it or its need to be set etc for most users.

internal/api/pipes.go
```diff
@@ -39,24 +43,32 @@ type PipesHandler struct {
...
-	return &PipesHandler{Source: source, PolicySource: policySource, CHConn: conn, Cache: c, queryTimeout: queryTimeout, logger: logger}
+func NewPipesHandler(source func(*settings.Store) pipes.Source, policySource PolicySource, conn driver.Conn, c cache.Cache, queryTimeout func() time.Duration) *PipesHandler {
+	return &PipesHandler{Source: source, PolicySource: policySource, CHConn: conn, Cache: c, queryTimeout: queryTimeout}
 }
 
-// List returns all named queries (admin endpoint).
-func (h *PipesHandler) List(w http.ResponseWriter, _ *http.Request) {
+// List returns all named queries of the ?tenant= (admin endpoint).
+func (h *PipesHandler) List(w http.ResponseWriter, r *http.Request) {
+	store, ok := opsStore(w, r, h.Tenants)
```
EricAndrechek: **[MUST] This guard doesn't hold when the request carries `?token=`.**

`opsStore` parses `r.URL.RawQuery` strictly on purpose — its comment at `tenant.go:103-106` says a malformed pair must be a `400` rather than silently reading as the default tenant. But `AuthMW` is registered ahead of these handlers (`router.go:169` vs the pipe routes at `:193-194`), and `bearerToken` rewrites the URL through exactly the lossy round-trip that strict parse exists to avoid:

```go
// internal/auth/auth.go:369
if params := r.URL.Query(); params.Get("token") != "" {  // discards ParseQuery's error
    params.Del("token")
    r.URL.RawQuery = params.Encode()                      // :372 — malformed pair now gone
}
```

Go's `parseQuery` skips only the offending segment and keeps going, so a surviving `token=jwt` is enough to satisfy the guard, and the rewrite then erases the malformed pair. Checked against stdlib rather than reasoned about:

```
"tenant=acme;x=1&token=jwt"  → RawQuery=""  ParseQuery err=<nil>  → tenant.Default → 200
"tenant=%zz&token=jwt"       → RawQuery=""  ParseQuery err=<nil>  → tenant.Default → 200
"tenant=acme;x=1"            → unchanged    err=semicolon separator → 400 ✓
```

So `GET /v1/ops/pipes?tenant=acme;x=1&token=<admin-jwt>` answers `200` with tenant `0`'s pipe definitions instead of the documented `400`. Same for `Get` at line 66. Harmless while only tenant `0` exists — it becomes a silent wrong-tenant admin read once story 3 lands a second tenant, which is the worst shape for this because it reports success.

The smallest fix is in `bearerToken`, not here — don't rewrite when the parse failed:

```go
params, err := url.ParseQuery(r.URL.RawQuery)
if err == nil && params.Get("token") != "" { /* … strip … */ }
```

That leaves a `?token=` unstripped in the malformed case (narrow log-hygiene regression on a request about to be refused); removing just the `token=` pair textually avoids even that.

Three follow-ons:

- `docs/src/content/docs/api.md:784` and `CHANGELOG.md:13` both currently promise this `400`, so they need correcting alongside.
- `TestPipesHandler_AdminReads_TenantParam` can't catch it: the `?tenant=acme;x=1` case at `tenant_test.go:185` calls `h.List`/`h.Get` directly, so the middleware that mangles the URL never runs. A regression test has to go through `NewRouter`.
- The CodeRabbit thread that marked this verified also recorded a persistent learning — *"In `internal/api/tenant.go`, `opsStore` must strictly parse the complete `r.URL.RawQuery` with `url.ParseQuery`… This intentional behavior applies to admin-only operations pipe reads"* — which will be replayed on future PRs. Worth correcting the thread so the learning doesn't outlive the bug.

internal/api/pipes.go
```diff
@@ -85,11 +101,11 @@ func (h *PipesHandler) Execute(w http.ResponseWriter, r *http.Request) {
 	// allowed_roles therefore authorizes nobody but admin (fails closed).
 	var p *policy.Policy
 	if h.PolicySource != nil {
-		p = h.PolicySource()
+		p = h.PolicySource(store)
 	}
 	role := policy.ResolveRole(p, auth.RoleFromContext(r.Context()))
 	if !policy.RoleAllowed(p, role, q.AllowedRoles) {
```
EricAndrechek: **[SHOULD] This gate is per-caller, but nothing about the response varies on the caller.**

`RoleAllowed` authorizes per request from the caller's resolved role, and this pipe read is a `200 GET` that emits no `Cache-Control` — the only two `Cache-Control` writers in the repo are `query.go:141` (`no-store`) and `stream.go:63` (`no-cache`), so this path sends neither. There's also no `Vary: Authorization` anywhere: the only `Vary` writers are `router.go:343` (`Origin`) and the new `tenant.go:86` (`X-Tenant-ID`).

Meanwhile `api.md:443` advertises this endpoint and `/v1/query` as "the cached read paths (dashboards, high-QPS clients, etc.)", which reads as an invitation to put a shared cache in front of it.

What actually keeps that safe today is RFC 9111 §3.5 — a shared cache may not store a response to a request carrying `Authorization` — rather than anything this handler asserts. The soft spot is the `?token=` fallback, where the credential sits in the URI and therefore in the cache's primary key, so it isolates by accident rather than by design.

Not necessarily asking for a code change: the `Vary: X-Tenant-ID` added in this PR is a good change and closes the tenant dimension properly. The ask is that the new `api.md:70` sentence not imply more than it buys. As written — "so a shared cache keys on the header and never replays one tenant's response to another" — it reads as "a CDN in front is now safe", and the credential dimension is the one that isn't keyed. Either send an explicit `Cache-Control` on this path, or state the boundary next to that sentence (and probably in `reverse-proxy.mdx`, which documents `X-Tenant-ID` as a request header and says nothing about shared caching of the pipe reads).

internal/api/pipes.go
EricAndrechek: **[SHOULD] The result-cache and singleflight keys aren't tenant-scoped.**

(File-level comment because the lines involved — `pipes.go:157` and `:168` — fall outside this PR's diff.)

`queryCacheKey(sql, params)` (`cache_key.go:27`) digests only the SQL text and the bound params, with no tenant. `pipes.go:157` derives the cache key from it and `pipes.go:168` reuses that same value as the **singleflight** key, so two tenants running a pipe with identical SQL would share not just one cached result but one in-flight execution. `structured_query.go:163`/`:185` has the same shape, and its invalidation namespace at `:172` is `{Table, Scope}` only — so one tenant's ingest would invalidate another tenant's cached reads.

Inert today: one store, so every key is already tenant `0`'s. And to be explicit, this is a *different* cache from the HTTP `Vary: X-Tenant-ID` this PR adds — that governs external shared caches keyed on the request; this is the in-process L1 plus singleflight. The new header doesn't touch it.

Flagging it on this PR rather than later for two reasons: the follow-up list covers the `chconn` getters without mentioning that the cache keyed off them also needs the tenant, and story 6 (per-tenant connections) is the point where this stops being inert and becomes a silent cross-tenant read rather than a crash.

The `TODO` immediately above at `pipes.go:155-156` already anticipates this key needing more inputs (table/scope deps so writes invalidate cached pipe results) — adding a tenant segment to the digest and to `cache.Namespace` is cheap in that same spot while it's still a no-op, and much harder to retrofit once there's real traffic on two tenants.

internal/api/pipes.go
EricAndrechek: I also wonder, do you think the tenant MW here should perhaps go BEFORE the admin MW? Like how else would the admin MW know which admin evaluation to run if it went before tenant MW?

docs/src/content/docs/api.md
EricAndrechek: I think I'd like to move all multi-tenant stuff to either a new docs page or to the deployment and ops type docs page – it will confuse people where in 99% of cases they don't need to set or worry about multi-tenant deployments, so we should just not even mention it or its need to be set etc for most users.

---

**Three findings land inside the section you'd be moving, so they're cheaper to fix during the move than to patch here and relocate later.** All three are on this page:

- **[MUST] `api.md:61` contradicts `access-control.mdx` on what "tenant" means** — see the comment on line 61. Relocating actually makes this *easier* to fix, because a dedicated page can define the word once and say plainly that it is not the row-isolation boundary.
- **[MUST] `api.md:70`'s `Vary` guarantee is false** for router-level `405`s and the CORS preflight — see the comment on line 70.
- **[MAY] `api.md:72`'s tenant-exempt list omits the deprecated `/health` and `/ready` aliases** — see the comment on line 72.

For the relocation itself: epic #583 assigns docs to **story 11** (`settings-directory.mdx` layout, "`api.md` for the header, `404`/`503`, and the reload parameter"), so a new page is a story-11 decision rather than something this PR has to settle. Two inbound links would need updating if the section moves — `reverse-proxy.mdx:197` and `sdk/index.mdx:380` both point into it.

internal/api/pipes.go
EricAndrechek: I also wonder, do you think the tenant MW here should perhaps go BEFORE the admin MW? Like how else would the admin MW know which admin evaluation to run if it went before tenant MW?

---

**Answer, from reading the wiring: the admin gate never consults the request's tenant, which is why it works today — and that's the real problem.**

There are two policy-source types, and this PR tenantised only one:

- `internal/api/tenant.go:19` — `type PolicySource func(*settings.Store) *policy.Policy` (the new tenanted one; handlers use it)
- `internal/policy/source.go:9` — `type Source func() *Policy` (the old untenanted one)

`Dependencies.PolicySource` is the **untenanted** one (`router.go:42`), `RequireAdmin` takes `policy.Source` (`router.go:259`), and `wire.go:63` binds it as `a.policies = policy.Source(store.Policy)` — a method value on the single adopted store. So `RequireAdmin` always evaluates **tenant `0`'s `policies.json`**, whatever `?tenant=` says.

On the literal ordering ask: epic #583's Decisions section already settles it the other way — *"Ops routes stay tenant-exempt (`/livez`, `/readyz`, metrics, version, `/v1/ops/*`)"* — and `TenantMW` reads `X-Tenant-ID` while the ops reads take `?tenant=`. Putting `TenantMW` in front of the ops tree would contradict that decision *and* `404` any admin request carrying a stray header.

But the thing the question detects is real, and it is **assigned to no story in the epic**: which tenant's `policies.json` decides who is admin when an ops call names a different tenant? Full detail in the comment on `internal/app/wire.go:63`.

EricAndrechek: **[MUST] This guard doesn't hold when the request carries `?token=`.**

`opsStore` parses `r.URL.RawQuery` strictly on purpose — its comment at `tenant.go:103-106` says a malformed pair must be a `400` rather than silently reading as the default tenant. But `AuthMW` is registered ahead of these handlers (`router.go:169` vs the pipe routes at `:193-194`), and `bearerToken` rewrites the URL through exactly the lossy round-trip that strict parse exists to avoid:

```go
// internal/auth/auth.go:369
if params := r.URL.Query(); params.Get("token") != "" {  // discards ParseQuery's error
    params.Del("token")
    r.URL.RawQuery = params.Encode()                      // :372 — malformed pair now gone
}
```

Go's `parseQuery` skips only the offending segment and keeps going, so a surviving `token=jwt` is enough to satisfy the guard, and the rewrite then erases the malformed pair. Checked against stdlib rather than reasoned about:

```
"tenant=acme;x=1&token=jwt"  → RawQuery=""  ParseQuery err=<nil>  → tenant.Default → 200
"tenant=%zz&token=jwt"       → RawQuery=""  ParseQuery err=<nil>  → tenant.Default → 200
"tenant=acme;x=1"            → unchanged    err=semicolon separator → 400 ✓
```

So `GET /v1/ops/pipes?tenant=acme;x=1&token=<admin-jwt>` answers `200` with tenant `0`'s pipe definitions instead of the documented `400`. Same for `Get` at line 66. Harmless while only tenant `0` exists — a silent wrong-tenant admin read once story 3 lands a second tenant, which is the worst shape for this because it reports success.

The smallest code fix is in `bearerToken`, not here — don't rewrite when the parse failed:

```go
params, err := url.ParseQuery(r.URL.RawQuery)
if err == nil && params.Get("token") != "" { /* … strip … */ }
```

Three follow-ons:

- `docs/src/content/docs/api.md:784` and `CHANGELOG.md:13` both currently promise this `400`, so they need correcting alongside.
- `TestPipesHandler_AdminReads_TenantParam` can't catch it: the `?tenant=acme;x=1` case at `tenant_test.go:185` calls `h.List`/`h.Get` directly, so the middleware that mangles the URL never runs. A regression test has to go through `NewRouter`.
- The CodeRabbit thread that marked this verified also recorded a persistent learning — *"In `internal/api/tenant.go`, `opsStore` must strictly parse the complete `r.URL.RawQuery` with `url.ParseQuery`… This intentional behavior applies to admin-only operations pipe reads"* — which will be replayed on future PRs. Worth correcting the thread so the learning doesn't outlive the bug.

**Scope note, and it may be the cheaper resolution.** This bug lives entirely in scope that was pulled forward. Epic #583 puts the optional tenant parameter in **story 2** ("Reload route gains the optional tenant parameter"), and this PR's own follow-up list concedes it: *"Which tenant an ops read addresses in nested mode is settled here as `?tenant=`; story 2 gives the reload route the same parameter."* Story 1's checklist doesn't ask for `?tenant=` on the pipe reads at all. So rather than fixing `bearerToken` here, consider dropping `?tenant=` from the ops reads and letting story 2 land it together with the reload route, the `bearerToken` fix, and a router-level regression test.

internal/api/pipes.go
EricAndrechek, line 99: **[SHOULD] This gate is per-caller, but nothing about the response varies on the caller.**

In one sentence: the authorization decision depends on *who* is asking, while the response carries nothing telling a cache the answer is caller-dependent — so a cache that stores one caller's `200` can replay it to a caller this gate would have `403`'d.

`RoleAllowed` authorizes per request from the caller's resolved role, and this pipe read is a `200 GET` that emits no `Cache-Control` — the only two `Cache-Control` writers in the repo are `query.go:141` (`no-store`) and `stream.go:63` (`no-cache`), so this path sends neither. There's also no `Vary: Authorization` anywhere: the only `Vary` writers are `router.go:343` (`Origin`) and the new `tenant.go:86` (`X-Tenant-ID`).

Meanwhile `api.md:443` advertises this endpoint and `/v1/query` as "the cached read paths (dashboards, high-QPS clients, etc.)", which reads as an invitation to put a shared cache in front of it.

What actually keeps it safe today is RFC 9111 §3.5 — a shared cache may not store a response to a request carrying `Authorization` — rather than anything this handler asserts. The soft spot is the `?token=` fallback, where the credential sits in the URI and therefore in the cache's primary key, so it isolates by accident rather than by design.

Not necessarily a code change: the `Vary: X-Tenant-ID` added in this PR is a good change and closes the tenant dimension properly. The ask is that `api.md:70` not imply more than it buys — as written, *"so a shared cache keys on the header and never replays one tenant's response to another"* reads as "a CDN in front is now safe", and the credential dimension is the one that isn't keyed. Either send an explicit `Cache-Control` on this path, or state the boundary next to that sentence and in `reverse-proxy.mdx`.

**Scope note:** the credential axis isn't assigned anywhere in epic #583 — story 8 is "Cache keyed by tenant", i.e. the *tenant* axis, and nothing covers keying on the caller. So this isn't waiting on a later story; it's either one header here or one paragraph of docs.

internal/api/pipes.go
EricAndrechek, line 1: **[SHOULD] The result-cache and singleflight keys aren't tenant-scoped — correctly deferred, but the epic's ordering is what makes that safe and it isn't written down as a dependency.**

(File-level because the lines involved — `pipes.go:157` and `:168` — fall outside this PR's diff.)

`queryCacheKey(sql, params)` (`cache_key.go:27`) digests only the SQL text and the bound params, with no tenant. `pipes.go:157` derives the cache key from it and `pipes.go:168` reuses that same value as the **singleflight** key, so two tenants running a pipe with identical SQL would share not just one cached result but one in-flight execution. `structured_query.go:163`/`:185` has the same shape, and its invalidation namespace at `:172` is `{Table, Scope}` only.

**Out of scope for story 1, and already assigned:** epic #583 gives this to **story 8, "Cache keyed by tenant"** — *"Tenant prefix on pipe/query cache keys and version namespaces; single global ristretto pool unchanged"* — and the Decisions section even fixes the key shape (`tenant.table.version.scope`). Nothing needs to change in this PR.

What's worth adding to the epic is the **ordering dependency**. The Order section puts story 8 in the parallel batch after story 2, and story 6's *implementation* (per-tenant ClickHouse connections) strictly later — so cache-keying is already sequenced ahead of per-tenant connections. That ordering is load-bearing rather than incidental: if story 6 ever landed first, every tenant-blind cache key becomes a silent cross-tenant **read** rather than a crash. Worth one line on story 8 (or in #583's Order section) stating that 8 must precede 6's implementation, so a future re-plan can't reorder them by accident.

Small bonus: the `TODO` at `pipes.go:155-156` already anticipates this key needing more inputs (table/scope deps so writes invalidate cached pipe results). When story 8 lands, that TODO and the tenant prefix are the same edit.

internal/app/wire.go
```diff
@@ -52,18 +54,36 @@ func withoutContext(release func() error) func(context.Context) error {
...
 // next request with no hook.
 func (a *App) wireSettings() error {
-	store, _ := settings.Open(a.cfg.Settings.Dir, slog.Default())
+	store, _ := settings.Open(a.cfg.Settings.Dir)
 	if store == nil {
 		return fmt.Errorf("settings directory %s invalid, refusing to start — findings above; `wavehouse validate` reproduces them, `wavehouse bootstrap` writes a starter directory", a.cfg.Settings.Dir)
 	}
 	a.store = store
+	a.tenants = settings.NewRegistry(store)
 	a.policies = policy.Source(store.Policy)
```
EricAndrechek, line 63: **[MUST — epic scope, not a defect in this PR] No story in #583 owns the admin-authorization axis, and this line is where that becomes visible.**

`a.policies = policy.Source(store.Policy)` binds the admin gate's policy source to the **single adopted store** — a method value on tenant `0`'s store, captured at wiring time. It flows to `Dependencies.PolicySource` (`wire.go:520`) and into `RequireAdmin` (`router.go:170`), which takes the *untenanted* `policy.Source` (`router.go:259`, `internal/policy/source.go:9`) rather than the new tenanted `api.PolicySource` (`internal/api/tenant.go:19`).

Consequence once a second tenant exists: **tenant `0`'s `policies.json` becomes a fleet-wide authorization authority.** An admin under tenant 0's policy can read any other tenant's pipe definitions via `GET /v1/ops/pipes?tenant=acme`, and tenant `acme`'s own `policies.json` has no say in who may do that. Same root on the operator-key path, where the authenticator reads this same source purely to learn the admin role's *name* (`auth.go:257-262`, `policy.AdminRole(p)`), so `admin_role`'s spelling is global too.

**Why this is an epic gap rather than a story-1 defect.** Story 1's checklist is about threading the tenant through *settings* getters, and this PR does that. The follow-up list parks the rest as *"Auth's policy source (story 9) … not threaded"* — but story 9 as written in #583 is only the per-tenant JWKS verifier (*"One verifier per tenant, rebuilt on that tenant's reload… HMAC secret and operator key stay boot-level"*). It says nothing about the policy source backing `RequireAdmin`, nor about role resolution. Meanwhile the Decisions section fixes ops routes as tenant-exempt, so `?tenant=` — not the header — is how an ops call names a tenant, and nothing anywhere states which tenant's policy then authorizes it.

So there's a real question with no owner: **is admin a fleet-operator role, or per-tenant?**

- *Fleet-operator* (one team runs the estate) — current behaviour is correct as built, and it should be recorded in #583's Decisions plus `access-control.mdx`, because it means a tenant's own `policies.json` cannot restrict who inspects its pipes.
- *Per-tenant* — `RequireAdmin` has to resolve the tenant named by `?tenant=` before evaluating, which is a story of its own. And it needs settling **before story 6**, because per-tenant ClickHouse connections are the point where an over-broad admin read stops being a settings read and becomes a data read.

Suggested amendment: widen story 9 to "per-tenant auth verifier **and policy source**", or add a story for it, and record the fleet-vs-per-tenant decision in the Decisions section either way. Nothing needs to change in this PR — but it shouldn't merge with the question unasked, because story 1 is the PR that makes the untenanted source load-bearing.

docs/src/content/docs/api.md
```diff
@@ -50,6 +50,29 @@ WaveHouse extracts the role from a configurable JWT claim path (`auth.role_claim
...
+A request without the header, or with an empty one, resolves to tenant `0`, the default tenant, whose settings are the [settings directory](/settings-directory). A settings directory defines that one tenant, so any other id is unknown. Setting the header on every request is the client's or the fronting proxy's job; WaveHouse never derives it from the token.
+
+A tenant id is 1–64 characters of ASCII letters, digits, `_`, and `-`. It is a string, not a number, so a long numeric id keeps every digit.
+
+| Status | Body | When |
+| ------ | ---- | ---- |
+| `400` | `{"error": "invalid X-Tenant-ID: …"}` | The id breaks the grammar above, or the header was sent more than once |
+| `404` | `{"error": "unknown tenant: <id>"}` | The id is well formed but no such tenant exists |
+
+Both are decided before authentication, so they are returned whatever token the request carries. Every response from a tenant route, these two included, carries `Vary: X-Tenant-ID`, so a shared cache keys on the header and never replays one tenant's response to another.
```
EricAndrechek: **[MUST] This guarantee is false for two response classes — and the CHANGELOG's wording for the same change is the accurate one.**

> Every response from a tenant route, these two included, carries `Vary: X-Tenant-ID`…

The header is added in `TenantMW` (`tenant.go:86`), which is `r.Use`d inside the inline group at `router.go:133`. chi composes an inline group's middleware **per registered endpoint handler**, so two response classes from a `/v1` route never run it:

- **Router-level `405`** — `r.MethodNotAllowed` is registered at the root (`router.go:75`), outside the group.
- **CORS preflight `204`** — `corsMiddleware` short-circuits every `OPTIONS` at the root (`router.go:359-365`) before `/v1` is entered.

Measured against the pinned chi v5.3.2 rather than read off the source:

```
GET     /v1/health            → Vary=[Origin X-Tenant-ID]   ✓
GET     /v1/ingest            → 405, Vary=[]
DELETE  /v1/stream            → 405, Vary=[]
OPTIONS /v1/query + Origin    → 204, Vary=[Origin]
```

No practical harm — all three are tenant-invariant responses — but `405` *is* one of RFC 9111 §4.2.2's heuristically-cacheable statuses, so it isn't a status a cache refuses to store, and this sentence exists precisely to state a guarantee.

`CHANGELOG.md:13` scopes the identical change correctly: *"Every answer **from the middleware**, the `400` and `404` included…"*. This page generalized past its own changelog. One-clause fix: "Every answer from the tenant middleware, the `400` and `404` included, carries…".

To be clear about what *is* right: the `400`/`404` half of the claim holds and is pinned by `tenant_test.go:53`, and the `Vary` code itself is correct — `Add` rather than `Set`, so the CORS `Vary: Origin` survives, and the new test in `router_test.go` genuinely pins the composition order rather than asserting its own setup.

See also the comment on `pipes.go:107` — the second half of this sentence ("never replays one tenant's response to another") is true on the tenant axis but reads as a broader safety claim than `Vary` alone delivers.

docs/src/content/docs/api.md
```diff
@@ -50,6 +50,29 @@ WaveHouse extracts the role from a configurable JWT claim path (`auth.role_claim
...
 
+## Tenant Selection
+
+Every `/v1` route outside `/v1/ops/*` resolves a tenant before it authenticates the request. The tenant comes from the `X-Tenant-ID` request header:
+
+```text
+X-Tenant-ID: 0
+```
+
+A request without the header, or with an empty one, resolves to tenant `0`, the default tenant, whose settings are the [settings directory](/settings-directory). A settings directory defines that one tenant, so any other id is unknown. Setting the header on every request is the client's or the fronting proxy's job; WaveHouse never derives it from the token.
```
EricAndrechek: **[MUST] "tenant" means two different things ten lines apart on this page.**

`api.md:51`, just above: *"Policies support Hasura-style row-level and column-level permissions with JWT claim templating (e.g. `{{ jwt.app_metadata.tenant_id }}`)"*.

This line: *"WaveHouse **never derives it from the token**"*.

And across to the access-control page, `access-control.mdx:213` says of *its* tenant: *"the value comes from the signed token, **not from anything the client sends**"* — with `:225` calling claim-based scoping *"the multi-tenant case"* and `:401` titling a worked example *"A multi-tenant analytics deployment"*. That's roughly 23 uses of the word in the older sense: a row-scoping value carried in the JWT.

**Why this isn't pedantry.** A reader designing an isolated multi-tenant deployment meets `X-Tenant-ID` first, concludes it's the supported isolation mechanism, and drops policy-based row scoping. Today that is a straight security regression: the header is unauthenticated, resolved pre-auth, and selects a *settings directory* only. Per epic #583, real data isolation arrives with story 6 (per-tenant ClickHouse connections) and story 5 (per-tenant JetStream streams) — neither in this PR. So for the whole of story 1 the word promises an isolation it does not provide.

Fix: one sentence here saying `X-Tenant-ID` selects which settings directory (and therefore which `policies.json`) serves the request, is client-supplied and resolved before authentication, and is **not** a row-isolation boundary — row scoping is `policies.json`'s job — linking `/access-control#row-level-security`. Plus the mirror note on the access-control side (comment there on `:363`).

The epic's Decisions section already has the right framing to borrow: *"`scope` is a within-table partition… tenant selects the settings and the tables. Different axes."*

docs/src/content/docs/api.md
```diff
@@ -50,6 +50,29 @@ WaveHouse extracts the role from a configurable JWT claim path (`auth.role_claim
...
+A tenant id is 1–64 characters of ASCII letters, digits, `_`, and `-`. It is a string, not a number, so a long numeric id keeps every digit.
+
+| Status | Body | When |
+| ------ | ---- | ---- |
+| `400` | `{"error": "invalid X-Tenant-ID: …"}` | The id breaks the grammar above, or the header was sent more than once |
+| `404` | `{"error": "unknown tenant: <id>"}` | The id is well formed but no such tenant exists |
+
+Both are decided before authentication, so they are returned whatever token the request carries. Every response from a tenant route, these two included, carries `Vary: X-Tenant-ID`, so a shared cache keys on the header and never replays one tenant's response to another.
+
+The probes (`/livez`, `/readyz`, `/healthz`), `/version`, the Prometheus metrics path, and `/v1/ops/*` are tenant-exempt: they ignore the header entirely. The two admin pipe reads ([`GET /v1/ops/pipes`](#get-v1opspipes--list-named-pipes) and `GET /v1/ops/pipes/{name}`) name their tenant with an optional `?tenant=` query parameter instead, with the same grammar and the same `400`/`404` answers; no other ops route takes a tenant.
```
EricAndrechek: **[MAY] This list reads as exhaustive but omits the deprecated `/health` and `/ready` aliases**, which are also tenant-exempt (`router.go:117-118`).

Ordinarily a shrug. It earns the two words here because the omission lands on exactly the pair a reader conflates — bare `/health` ignores the header while `/v1/health` resolves a tenant — and this same page documents `/health` as a live alias at `:114` and `/v1/health` as tenant-scoped at `:176`.

Fix: "(`/livez`, `/readyz`, `/healthz`, and the deprecated `/health`, `/ready`)".

(`architecture.md` is fine — it says "the probes" generically, so no second copy to sweep.)

Separately, the `?tenant=` correction in the second half of this line is now exactly right, not merely reworded — I verified `opsStore` has precisely two callers repo-wide, `pipes.go:52` (`List`) and `:66` (`Get`), while `Execute` uses `requestStore`, and `router.go:173-197` confirms `schema`, `schema/refresh`, `dlq/stats`, `ops/query` and `settings/reload` take no tenant. So "no other ops route takes a tenant" holds and `api.md:784` agrees with it.

docs/src/content/docs/access-control.mdx
```diff
@@ -360,7 +360,7 @@ A few more edges worth knowing when you write a policy — the stream evaluates
...
+One more boundary is temporal: a subscriber's claims (and role) are captured when the SSE connection is established and are never re-read, while the policy itself is re-read on every event, live and replayed alike — so a policy adopted mid-gap-fill applies to the next replayed row. Tightening a policy therefore applies from the next event either way, but a token that expires — or claims revoked at the identity provider — keeps its open stream until the client disconnects, so treat connection lifetime as the revocation window for stream row-scoping.
```
EricAndrechek, line 367: **[MUST — mirror of the `api.md:61` comment] This is the only line the branch touches in this file, and the file's meaning of "tenant" now collides with the new one.**

`access-control.mdx` uses "tenant" ~23 times to mean *a row-scoping value carried in the signed token*:

- `:213` — "the value comes from the signed token, not from anything the client sends"
- `:225` — "This is the multi-tenant case — the allowed set rides on the signed token"
- `:401` — "A multi-tenant analytics deployment"

This PR makes the same word mean *the settings directory selected by a client-supplied, pre-auth header*. Nothing on this page is wrong on its own terms — the collision is only visible across the two pages, which is exactly why it survived review on both sides.

Fix on this side is one note near `:225`: the policy document itself is selected by `X-Tenant-ID` before authentication, today exactly one tenant (`0`) exists, and claim-templated row scoping is a different axis from tenant selection. Epic #583's Decisions section already phrases it usefully — *"`scope` is a within-table partition… tenant selects the settings and the tables. Different axes."*

Worth coordinating with the relocation proposed on `api.md:74`: if the multi-tenant material moves to its own page, this page should link to it once and keep its own "tenant" firmly meaning the row-scoping claim.

(For the record, the change actually made on this line is correct — I verified all four copies of the gap-fill policy-snapshot claim moved together: here, `api.md:642`, `sdk/streaming.md:136` and `architecture.md:99`, matching `hub.go:478` now calling `snapshotPolicy()` inside the replay closure. No stale fifth copy.)

docs/src/content/docs/reverse-proxy.mdx
```diff
@@ -194,6 +194,7 @@ These limit the *whole* request regardless of traffic, so no keepalive extends t
...
 - **`X-Operator-Key`** (optional) — only relevant if you present the [operator key](/access-control#operator-key) via this alias header instead of `Authorization: Operator <key>`. Custom request headers are forwarded by default, but confirm your proxy doesn't strip it — and note nginx silently drops header names containing **underscores** (this one uses hyphens, so it's fine as named). The `Authorization` form needs none of this.
+- **`X-Tenant-ID`** (optional) — selects the [tenant](/api#tenant-selection); absent means tenant `0`. It is resolved before authentication and never derived from the token, so if the proxy owns tenant selection it must **replace or strip** a client-supplied value, not add to it: two `X-Tenant-ID` lines are refused with `400`, the same append-vs-replace hazard as `Content-Type` below. The name is hyphenated, so nginx's underscore rule does not apply.
```
EricAndrechek: **Two findings on this bullet.**

---

**[SHOULD] The strip advice is conditional, but the failure mode isn't.**

As written — *"so **if** the proxy owns tenant selection it must replace or strip"* — an operator who isn't using tenants reads this as inapplicable and skips it. But `X-Tenant-ID` is a generic header name, so **any** deployment whose gateway or service mesh already stamps one gets `404 unknown tenant: <id>` on *every* `/v1` route after upgrade — including `/v1/health` (`router.go:143`, inside the `TenantMW` group), which is the SDK's public reachability ping (`api.md:176`). So `wh.sys.health()` breaks too, and the thing an operator would reach for to diagnose it is the thing that's broken.

There's also no upgrade note anywhere: `deployment.md` contains **zero** occurrences of "tenant" (`grep -ci tenant` → 0), despite §"Upgrading across the v2 ingest envelope" (`deployment.md:351`) being the established precedent for exactly this kind of note.

Suggest making the advice unconditional — "strip inbound `X-Tenant-ID` at the edge unless you are deliberately using it; WaveHouse now answers `404` on every `/v1` route, the SDK's `/v1/health` ping included" — plus a short upgrade bullet naming the 404-everything symptom.

---

**[MAY] The `Content-Type` analogy runs the wrong way.**

This bullet calls a repeated `X-Tenant-ID` *"the same append-vs-replace hazard as `Content-Type` below"*, but `:204` teaches that repeated `Content-Type` **lines** "are resolved together and accepted when they agree" — only a comma-joined merge is refused. `resolveTenant` refuses on `len(values) > 1` unconditionally (`tenant.go:58-61`), so two *agreeing* `X-Tenant-ID: acme` lines are a `400`. A reader who carries the `Content-Type` rule across concludes an agreeing duplicate is fine and meets a fleet-wide `400`.

The strictness is deliberate and I'd keep it — the code's own comment gives the reason ("a value a proxy sets can never be shadowed by one the client sent"). This is only the cross-reference: "a **stricter** version of the append-vs-replace hazard as `Content-Type` below — here even two agreeing lines are refused."

---

If the multi-tenant docs move (see `api.md:74`), this bullet's `/api#tenant-selection` link needs repointing, and the `Vary`/shared-cache boundary from the `pipes.go:107` comment would naturally live here too — this page currently documents `X-Tenant-ID` as a request header and says nothing about the new `Vary` response header or about caching the pipe reads.

internal/discovery/discovery.go
```diff
@@ -429,7 +432,7 @@ func (sr *SchemaRegistry) RetryRefresh(ctx context.Context, initialBackoff, maxB
 // — an in-flight wait finishes at the old cadence rather than resetting,
 // which keeps a reload from ever deferring an imminent refresh.
 func (sr *SchemaRegistry) StartAutoRefresh(ctx context.Context) {
-	interval := sr.refreshInterval()
+	interval := sr.refreshInterval(sr.tenant)
 	ticker := time.NewTicker(interval)
```
EricAndrechek, line 449: **[SHOULD] `NewTicker` panics on the zero duration a registry miss produces, and it takes the process down rather than degrading.**

`perTenant` (`wire.go:75-85`) logs and returns `T`'s zero value on a registry miss, and `wire.go:185` passes it straight in as the schema-refresh interval:

```go
registry := discovery.NewSchemaRegistry(a.ch, a.ch.Database, tenant.Default,
    perTenant(a.tenants, (*settings.Store).SchemaRefreshInterval))
```

`time.NewTicker(interval)` on this line and `ticker.Reset(interval)` at `:448` both panic on a non-positive duration. `StartAutoRefresh` runs under `errgroup.WithContext` (`app.go:195`, `g.Go` at `:204`) and there is no `recover()` anywhere in `internal/app`, so the panic is fatal.

Unreachable today: `settings/registry.go:14` is the only constructor and there's no mutator, so the registry always holds `tenant.Default`. Worth fixing now anyway, because **epic #583 story 5 makes the tenant id data-derived** — *"the ingest worker, sweeper, DLQ, and stream replay read it back from `Message.Topic()`"* — at which point a miss becomes reachable from message content rather than from a wiring bug.

Also a small correction to the PR body: the follow-up list enumerates the zero-value consequences as "DLQ off, nil policy, gap window 0" and omits this fourth one. It's different in kind from the other three — those degrade, this one crashes — so it's worth naming explicitly rather than folding into the same bullet.

Cheapest fix: clamp here (non-positive → keep the current interval), or have `perTenant` fall back to the default store's value, so the documented "reads as the zero value" contract is honest for every getter rather than for three of four.

internal/ingest/worker.go
```diff
@@ -719,7 +723,7 @@ func (w *IngestWorker) rejectPoison(ctx context.Context, m *mq.Message, tableNam
 		})
 		return
 	}
-	w.logger.ErrorContext(ctx, "unreadable envelope dropped — the DLQ is disabled for this table, and a message that can never insert must not redeliver forever",
+	slog.ErrorContext(ctx, "unreadable envelope dropped — the DLQ is disabled for this table, and a message that can never insert must not redeliver forever",
```
EricAndrechek, line 726: **[SHOULD] An unknown tenant reads as "DLQ off", and on this path that destroys the message.**

`wire.go:358`'s `dlqEnabled` returns `false` on a registry miss (logged "treating its DLQ as off"). At `worker.go:711` that `false` selects this branch, which logs and then `DoubleAck`s — the envelope is gone. On the isolation path (`:577`) the same `false` is safe, because the row is left unacked and redelivers.

The log line here says *"the DLQ is disabled for this table"*, and the doc comment at `:705-709` describes dropping as the fallback for exactly that case. But "the DLQ is disabled" and "we can't tell which tenant this is" are different statements, and only the first justifies discarding.

Failure scenario once **story 5** derives the tenant from the MQ subject: a tenant is removed from the registry while its envelopes are still in the stream, and every unreadable envelope for it is silently destroyed — precisely the case the DLQ exists for, and it erases the evidence of the wiring bug that caused it.

Worth noting this would also contradict an explicit epic decision: #583 story 3 says removal *"never touches data: the tenant's JetStream stream, DLQ, and dedupe store stay on disk, and restoring the folder restores the tenant."* A drop-on-miss makes that false for anything in flight at the moment of removal.

Fail-safe for an unknown tenant is to park on the DLQ, or leave unacked — not to drop. Currently unreachable (one tenant), so this is a "fix while it's free" rather than a live bug.

internal/api/tenant_helpers_test.go
```diff
@@ -0,0 +1,34 @@
...
+// call a tenant-route handler without the router.
+func withTenant(r *http.Request) *http.Request {
+	return r.WithContext(WithStore(r.Context(), testStore))
+}
+
+// testTenants is a registry whose default tenant is testStore.
+func testTenants() *settings.Registry { return settings.NewRegistry(testStore) }
+
+// staticPolicy is a PolicySource fixed to p, whatever the tenant.
+func staticPolicy(p *policy.Policy) PolicySource {
```
EricAndrechek, line 26: **[SHOULD] Every double discards the store it's handed, so no test in this PR can catch a wrong-store derivation — and the blocker is the registry, not the doubles.**

`staticPolicy` is `func(*settings.Store) *policy.Policy { return p }`, `staticPipes` is the same shape (`:31`), and `internal/stream/tenant_test.go:9` does it with `tenant.ID`. Each discards the argument the handler just derived. So `IngestHandler.Handle`, `StructuredQueryHandler.Handle` and `PipesHandler.Execute` could pass `nil`, a stale global, or any other store to `PolicySource` / `Source` / `DedupeSettings` / `bucketSecs` / `defaultMaxRows` and the suite would stay green. Threading the tenant is what this PR *is*, and it's the one property untested — the trap AGENTS.md names directly: a harness that supplies the value the code is supposed to derive tests your belief about the computation, not the computation.

One mitigant I checked and want to credit: `testStore = &settings.Store{}` (`:14`) has a nil doc and `settings/store.go` dereferences through it, so a handler reading the store *directly* would panic. That catches "didn't read from the context" but not "read the wrong store".

**Why I don't think it's fixable in this PR, and where it belongs.** `settings/registry.go:14` exposes only `NewRegistry(store)`, which hard-codes `map{tenant.Default: store}` with no mutator — so a two-tenant registry cannot be constructed and the catching test cannot be written today. That registry is itself story-2 scope pulled forward out of necessity (story 1's middleware must resolve against *something*, while epic #583 story 2 is where `settings.Open` "returns a registry keyed by tenant with `For(id) (*Store, bool)`").

So the honest resolution is a line on story 2 rather than a test seam bolted on here: when the real registry lands, add one case that drives `NewRouter` + `TenantMW` with two tenants and asserts `assert.Same` on the store a recording `PolicySource` was handed. Same shape as the existing `assert.Same` in `TestTenantMW`, one layer further down. Flagging it now so it lands as planned story-2 scope instead of being re-raised as a doubles nit every round.

internal/api/router.go
```diff
@@ -120,45 +125,49 @@ func NewRouter(deps Dependencies) http.Handler {
...
-		// the hood (200 past boot, 503 while degraded), no body.
-		r.Get("/health", deps.Health.Online)
+			// Public content-free liveness ping. Lives under /v1 deliberately:
+			// it's documented API surface the SDK relies on to check "is this
+			// server reachable" before sending data, so it must stay public even
+			// in deployments that filter the bare /livez|/readyz|/healthz probe
+			// paths at the reverse proxy. AuthMW runs but never rejects, so no
+			// token is required and there's no authz gate. Mirrors /livez under
+			// the hood (200 past boot, 503 while degraded), no body.
+			r.Get("/health", deps.Health.Online)
```
EricAndrechek, line 143: **[MAY] `/v1/health` is now an unauthenticated tenant-existence oracle.**

`TenantMW` (`:133`) runs before `AuthMW` (`:134`), and this route is deliberately public — the comment above it notes AuthMW "runs but never rejects", and AGENTS.md calls it the SDK's content-free public ping. So anyone, with no credential and no rate limiting on this route, can send `GET /v1/health` with a guessed `X-Tenant-ID` and distinguish `200` (tenant exists) from `404` (it doesn't).

Discloses nothing today — only the reserved `0` exists. It becomes free customer enumeration once ids are customer-derived, which is exactly what epic #583 story 3 enables (*"a new folder plus a reload call adds the tenant"*), and tenant ids are folder names chosen by the control plane.

Not asking for a redesign — resolve-before-auth is story 1's premise and the epic states it outright (*"Requests resolve to a tenant before auth"*). But it's a consequence #583 doesn't record anywhere, and the answer gets much harder to change once clients depend on the `404`. Cheapest options if you want it closed: exempt `/v1/health` from `TenantMW` (it's content-free, so it has nothing tenant-specific to serve), or have the probe answer tenant-independently.

Worth one line in the epic so story 3 picks deliberately rather than inheriting it. Related: the `404` for an unknown tenant is also what makes the `reverse-proxy.mdx:197` upgrade hazard bite on this very route — see that comment.

internal/auth/auth.go
EricAndrechek, line 1: **[MUST — pointer] The fix for the ops `?tenant=` bug lives in this file, at `bearerToken` (`auth.go:369-372`).**

File-level because those lines fall outside this PR's diff — this file is touched by the slog cleanup, but not there.

`bearerToken` strips a `?token=` by round-tripping the query through `r.URL.Query()` / `params.Encode()`, which discards `ParseQuery`'s error and rewrites `RawQuery` wholesale. Because `AuthMW` runs ahead of the ops handlers (`router.go:169` vs `:193-194`), that erases a malformed query *before* `opsStore`'s deliberately-strict `url.ParseQuery` can reject it. Net effect: `GET /v1/ops/pipes?tenant=acme;x=1&token=<admin-jwt>` silently reads tenant `0` instead of returning the documented `400`.

Suggested change:

```go
params, err := url.ParseQuery(r.URL.RawQuery)
if err == nil && params.Get("token") != "" { /* … strip … */ }
```

That leaves a `?token=` unstripped in the malformed case — a narrow log-hygiene regression on a request about to be refused, and the comment above `bearerToken` is explicit that the strip only protects our own logs anyway. Removing just the `token=` pair textually avoids even that.

Full detail, the stdlib repro, the affected docs, and a scope argument for deferring the whole thing to story 2 are in the comment on `internal/api/pipes.go:52`.

EricAndrechek (changes requested):
Reviewed the full diff at `f1bbc279` against epic #583's story 1, plus both CodeRabbit rounds and CI.

**Scope first, because that's where the defects cluster.** Story 1's checklist is delivered essentially in full: `internal/tenant`, the middleware (validate → default `0` → resolve → `404` → context), running before auth, handlers taking the store as an explicit argument, the getters threaded through `api`/`ingest`/`discovery`/async, wired to the default tenant in `wire.go`, and the slog cleanup owed from #586. Nothing material is missing.

Three things went beyond it, and every merge-relevant defect is in one of them:

1. **`?tenant=` on the ops pipe reads** — story 2 scope per the epic, and the PR body says as much. The one MUST-level bug (`bearerToken` defeating `opsStore`'s strict parse) lives entirely here. Worth considering dropping it so story 2 lands it together with the reload route, the `bearerToken` fix, and a router-level regression test.
2. **`Vary: X-Tenant-ID`** — unlisted scope from round two. The code is right — `Add` not `Set`, and the new test genuinely pins the composition — but the sentence documenting it overstates the guarantee.
3. **`settings.Registry`** — story 2 scope, but unavoidable, since story 1's middleware must resolve against something. Shipping it as a stub is why no test can catch a wrong-store derivation.

**One epic gap I'd like settled before merge:** no story in #583 owns the admin-authorization axis. `wire.go:63` binds `RequireAdmin`'s policy source to tenant `0`'s store, so tenant 0's `policies.json` becomes a fleet-wide authorization authority the moment a second tenant exists. Story 9 covers the JWKS verifier only, not the policy source. Detail on that line.

Everything else is latent until story 3/5/6, or a docs fix. Full context on each comment, including which story owns it.

internal/app/wire.go
taitelee: Settled, and recorded in #583. The new Decisions bullet says the operator key is the fleet credential and a token's admin role is per tenant, so an ops call that names a tenant is authorized against that tenant's `policies.json`. Tenant `0`'s policy is never a fleet-wide authority and nested mode needs no `0` folder.

Where it lands: story 2 carries the `RequireAdmin` half in the same PR as `?tenant=`, so the parameter never ships ahead of the gate. Story 9, now "verifier and policy source", keeps the tenant-route half (the operator key's admin role reads the request's resolved store).

Nothing changes in this PR. With `?tenant=` dropped in d0913406 no ops call names a tenant, so the gate behaves exactly as it did before the PR.

internal/api/tenant.go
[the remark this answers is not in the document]
taitelee: @coderabbitai the learning recorded on this thread is obsolete. `opsStore` and the `?tenant=` parameter were removed from this PR in d0913406, and the strict `url.ParseQuery` was bypassed anyway: `auth.bearerToken` rewrites `RawQuery` ahead of the handler whenever the request carries `?token=`, which erases the malformed pair before it can be refused. The parameter returns in #583 story 2 together with that fix. Please remove the learning about `opsStore` strictly parsing `r.URL.RawQuery`.

internal/app/wire.go
taitelee: Update: decided differently with Eric after this. In nested mode the ops routes take the operator key only and tenant admins never touch them. Flat mode is unchanged. #583's Decisions section has the current wording.

taitelee:
@coderabbitai full review

